PR #163, opened 2014-11-26
[WIP] add capability to extract isosurfaces within proteus and generate povray scenes

cekees:
This has a prototype implementation of triangulating an isosurface of a scalar field, and it generates a povray scene. Needs a lot of refactoring into modules and script with real interface.


cekees:
# TODO
- [ ] Factor our scene info to .inc file so .pov frames only have geometry
- [ ] Properly add domain boundary and solid obstacles to  scene
- [ ] Factor out isosurface extraction so it can be used from script and AV
- [ ] Cythonize and optimized parallel i/o

PR #171, opened 2014-12-12
[WIP] cekees/add nonlinear pod

cekees:
@MajinSaha and @mfarthin this is my branch with the general nonlinear POD implemented as a special solver for the full system. I need  to shelve this work for a couple of days at least so I'm going to leave this PR here as a work  in progress unless Matthew wants me to merge it with his deim experiments. Here are the things I see that need thought/work on this branch.

- [ ] Decide on notation: our U is Sorenson's V, Sorenson's U is the reduced based for F, our u is the unknown
- [ ] Optimize loops for calculating U^t J U (this is just to speed  up our planned sensitivity studies), current implementation is just a pure python prototype.
- [ ] Add DEIM to calculation P and modify F and J calculations (still using full F  and J at first)
- [ ] Modify F and J API to allow efficient computation of U (P^t U)^-1 P^t F and U (P^t U)^-1 P^t J U


cekees:
@mfarthin I went ahead  and merged your branch into this PR since this branch hadn't changed since you merged it with your branch. If I'm following what you did correctly, I should be able to build the DEIM approximations to F and J in the POD_Newton algorithm by generating rho and PF from  snapshots of the full residual (by archiving snaps of F with the right simFlags and then feeding those to your get_deim function).   

It's hard for me to tell the difference between the two NonlinHeat notebooks. Are there supposed to be two?

test/POD/burgers.py
```diff
@@ -0,0 +1,71 @@
...
+
+.. math: \frac{\partial u}{\partial t} + \nabla \cdot (\frac{1}{2} u^2 - \epsilon \nabla u) = 0
+"""
+from burgers_init import *
+
+physics.name = "burgers_3d"
+so.name = physics.name
+Profiling.logLevel=5
+Profiling.verbose=True
+ns = NumericalSolution.NS_base(so,[physics],[numerics],so.sList,opts)
```
cekees: @mfarthin, here I should be able to add the proper simFlags to archive the full nonlinear residual right?

test/POD/svd_burgers.py
```diff
@@ -0,0 +1,27 @@
...
+
+T = 1.0
+nDTout = 100
+DT = T/float(nDTout)
+
+archive = Archiver.XdmfArchive(".","burgers_3d",readOnly=True)
+
+import numpy as np
+
+u = read_from_hdf5(archive.hdfFile,'/u0')
```
cekees: This bit I should be able to refactor using deim_utils to generate the SVD for  the u and F snapshots at the same time.

proteus/NonlinearSolvers.py
```diff
@@ -617,6 +618,102 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
...
+                norm,
+                convergenceTest,
+                computeRates,
+                printInfo,
+                fullNewton,
+                directSolver,
+                EWtol,
+                maxLSits)
+        self.DB = 11
+        U = np.loadtxt('SVD_basis')
```
cekees: @mfarthin, if  I generate the SVD_basis  for  both u and F in svd_burgers, then here I just read them both in and call `deim_utils.get_deim` to construct `rho` and  your `PF`.

proteus/NonlinearSolvers.py
```diff
@@ -617,6 +618,102 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
...
+        self.linearSolverFailed = False
+        while (not self.converged(pod_r) and
+               not self.failed()):
+            log("   Newton it %d norm(r) = %12.5e  \t\t norm(r)/(rtol*norm(r0)+atol) = %g test=%s"
+                % (self.its-1,self.norm_r,(self.norm_r/(self.rtol_r*self.norm_r0+self.atol_r)),self.convergenceTest),level=1)
+            if self.updateJacobian or self.fullNewton:
+                self.updateJacobian = False
+                self.F.getJacobian(self.J)
+                self.pod_J[:] = 0.0
+                for i in range(self.DB):
```
cekees: This block gets and alternative implementation for a `pod_deim_J`

proteus/NonlinearSolvers.py
```diff
@@ -617,6 +618,102 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
...
+            if not self.linearSolverFailed:
+                #self.linearSolver.solve(u=self.du,b=r,par_u=self.par_du,par_b=par_r)
+                #self.linearSolverFailed = self.linearSolver.failed()
+                self.pod_linearSolver.solve(u=self.pod_du,b=pod_r)
+                self.linearSolverFailed = self.pod_linearSolver.failed() 
+            #pod_u-=np.dot(self.U_transpose,self.du)
+            pod_u-=self.pod_du
+            u[:] = np.dot(self.U,pod_u)
+            self.computeResidual(u,r,b)
+            pod_r[:] = np.dot(self.U_transpose,r)
```
cekees: Initially I just replace this line with the `PF` projection applied to `r[rho]`.

cekees:
One thing I noticed is that we're basically moving forward with Sorenson's notation where U and Um are the nonlinear function snapshots.

proteus/NonlinearSolvers.py
mfarthin: @cekees sorry I missed these comments. I thought I had my gmail open on my laptop and just pushed some more additions. Yes, you should be able to build the rho and PF here. I added archiving for the 'mass' residual in Transport and NumericalSolution. That way, we can build the deim interpolants for both the full residual. Presumably, the rho indices built for the spatial term would do fine for the temporal terms here because they're linear but I think we need to have separate bases in general.

proteus/NonlinearSolvers.py
mfarthin: @cekees Yes, that's right.

proteus/NonlinearSolvers.py
mfarthin: @cekees sorry I forgot to add that the PF in the deim_utils still needs to be left multiplied by the snapshot basis matrix (transposed) to get the final matrix for the coarse space

proteus/NonlinearSolvers.py
mfarthin: @cekees, I think that will work ok for burgers, but I also starting adding the pieces for doing

```
pod_r[:] = np.dot(self.U_t_PFs,r_s[rho_s]) + np.dot(self.U_t_PFm,r_m[rho_m])
```
where the `_s` and `_m` suffixes mean the deim approximations for the `spatial` and `mass` residuals

mfarthin:
Yeah, I'm ok if we switch the notation but I had my notes already scribbled in the Sorensen notation

cekees:
@mfarthin I think we should just stick with Sorensen's notation.

mfarthin:
@cekees I went ahead and merged my `wip_add_deim` branch with this pull request. The deim implementation needs more work, but seems to be running for 1d burgers depending on the number of modes being used but isn't giving quadratic convergence.

cekees:
Thanks. Are you thinking we should go ahead and merge this into master and then prioritize a set of smaller projects to improve what is here? Or should we keep working on the branch?

cekees:
@mfarthin this is ready to merge as far as I'm concerned

mfarthin:
Thanks

PR #181, opened 2014-12-31
cekees/enable shippable

cekees:
After AGU I messed around a little bit with [docker](https://www.docker.com/) and found this relatively new continuous integration/deployment service  called  [shippable](https://www.shippable.com/) that integrates with docker and github. This PR does the following:

- builds  proteus using a hashdist cache built in the same docker container*
- runs `nose` with `coverage` enabled to generate coverage reports (https://app.shippable.com/projects/545531c244927f89db3e7d70)
- pushes the resulting docker container to [docker hub](https://registry.hub.docker.com/u/erdccm/proteus)

*right now I have to download  the hashdist cache because I'm using shippable's "minions".  There is a way to use your own docker container, which would just have the cache already, but you have to set up your own minion on rackspace or something.


test/test_meshParitionFromTetgenFiles.py
```diff
@@ -1,98 +0,0 @@
-#!/usr/bin/env python
```
cekees: This is going away because of the spelling error in the file name.  I'll update make check  to point to the correct one.

Issue #182, opened 2015-01-01
Gauge is grabbing the wrong value in buildQuantityRow


ahmadia:
In: https://github.com/erdc-cm/proteus/blob/master/proteus/Gauges.py#L110

```
location, node = quantity
# get the FiniteElementFunction object for this quantity
femFun = self.model.levelModelList[-1].u[quantity_id]
# get the mesh for this quantity
mesh = femFun.femSpace.mesh
```

`quantity_id` here is a numbering used internally by the Gauge to simplify indexing into the linear operator used for computing point values, and is the wrong index for `u`.

We need to pass in the `field` object available in the calling function, `buildGaugeOperators`, to fix this.

PR #183, opened 2015-01-01
[Ready for Review] Better Gauge support

ahmadia:
I am now verifying a very simple gauge in parallel.  I'm going to run this at scale on Garnet before we merge this.


ahmadia:
This fixes: https://github.com/erdc-cm/proteus/issues/182 as well.

ahmadia:
This fixes two bugs in the current implementation of Gauges, which are currently completely broken in master for parallel runs.

proteus/test/mock_model.py
```diff
@@ -0,0 +1,19 @@
+from collections import namedtuple
+
+Level = namedtuple('Level', ['variableNames', 'nodeArray', 'u', 't', 'mesh', 'coefficients'])
+
+
+class MockModel:
```
cekees: Are you using this  anywhere?

cekees:
@ahmadia did you run this at scale on garnet yet? Just out of curiosity, what was the main bug fix besides #182?  I think the PR looks good and is ready  to go (after removing the MockModel if  it's not needed), but we might want to come up with some issues for future work. For  example, the gauge implementation uses FemTools in such a way that it should be valid for all meshes and elements, but we only test on 3D simplices.

proteus/test/mock_model.py
ahmadia: No, let's toss it.

proteus/test/test_gauges.py
```diff
@@ -0,0 +1,131 @@
...
+    p.nd = 3
+
+    class LinearSolution:
+        def uOfXT(self,x,t):
+            return (x[0]+10*x[1]+100*x[2])*(t+1.0)
+
+    p.initialConditions = {0:LinearSolution()}
+    p.dirichletConditions = {0: lambda x,flag: None}
+    p.domain = Domain.RectangularDomain(name="test_gauges_domain")
+    p.coefficients = TransportCoefficients.PoissonEquationCoefficients(aOfX = [lambda x: np.eye(p.nd, p.nd)],
```
cekees: Since PoissonEquationCoefficients can generate a multi-component problem we could easily test the multi-component support of the gauges and set `p.coefficients.variableNames`  to something besides the default names `[`u0,u1,...]`.

ahmadia: Agreed.  I'd also like to test more than two locations, and perhaps generate a grid of 50 or so points to measure and test, clustering some of them so that there are many on one processor, and fewer on others.

proteus/test/test_gauges.py
```diff
@@ -0,0 +1,131 @@
...
+        gauge_names = [gauge_name.strip() for gauge_name in header[1:]]
+        f.seek(0)
+        data = np.genfromtxt(f, delimiter=",", skip_header=1)
+    return gauge_names, data
+
+
+def test_gauge_output():
+    filename = 'test_gauge_output.csv'
+
+    p = PointGauges(gauges=((('u0',), ((0, 0, 0), (1, 1, 1))),),
```
cekees: Since these are precisely at corner nodes of the mesh we may want to add explicit locations outside the mesh and within an interior element, like (0.5,0.5,0.5)+(eps,eps,eps).

ahmadia: Agreed.

proteus/test/test_gauges.py
cekees: Good idea.

ahmadia:
I ran this on Garnet, but it failed because it couldn't partition the mesh in under an hour.  @cekees - I would definitely benefit from your guidance here.  The goal, of course, is to test the gauges, not the mesh partitioning, and ideally, the test shouldn't take more than a minute to run (since it runs in less than a second on a single process).

cekees:
We also might want to add ConstantSolution, QuadraticSolution, etc. and check accuracy. I'm not suggesting we do all  this before merging the PR.  Just mentioning things as they occur to me while reviewing. Some if of this is really testing the underlying finite elements, but it's sometimes useful to test in an arbitrary setting, which is what the arbitrary point locations provide an opportunity to do.

cekees:
That's strange.  What was your input (how many processors, etc.)?

ahmadia:
I'm just reusing the proteus test mesh partitioning setup, so once on a single process and another with 512 processes.  I am also still getting the spurious errors, I don't know if my preload configuration is correct.

ahmadia:
I think the Garnet admins/Cray have subtly renamed a few of the installation files.  For example, I think `/opt/cray/mpt/6.0.0/gni/mpich2-gnu/48/lib/libmpichcxx.so.2.0.0` may now need to be `/opt/cray/mpt/6.0.0/gni/mpich2-gnu/48/lib/libmpichcxx_gnu_48.so.2.0.0` based on an inspection of the installed petsc4py file.  I think at install time it might be handy to generate a `.py` file that automatically does the preloading based on the linked MPICH library.

ahmadia:
I am belatedly pointing out that `util` is not currently installed, so this branch is not ready to merge until I push a fix to the installer.

ahmadia:
`proteus.test` is now installed.  The test appears to be working correctly up to 32 processes, but then segfaults on 64.  I'm going to try and add a few more nodes to the mesh.

ahmadia:
If I double the number of nodes in the mesh, it runs correctly on 64 and 128 processes.

ahmadia:
And if I understand what's happening correctly, every process tries to write to the three `tetgen` files as a result of the call here:

https://github.com/erdc-cm/proteus/blob/master/proteus/MeshTools.py#L1936

ahmadia:
I'm surprised this isn't crashing, but I guess the filesystem is just deadlocking to allow every process a chance to write to the tetgen file before it quits, which is where the "mesh partitioning" slowdown is occurring.  This looks like a serial operation that assumes the entire mesh is available on the first processor, but I really don't know my way around this part of the code.

ahmadia:
After playing around with this a bit, the job appears to still be crashing on 512 processes, even with twice as many nodes.  I'm increasing the number of nodes once more to a total factor of 8 per process, and I've commented out the call to output the tetgen files in my local install, which seems to have sped things up.

ahmadia:
Okay this is working, I needed to double the number of nodes from what we'd originally calculated to make things work on 512 processes.

ahmadia:
I've also renamed the directory to `proteus/tests` so it gets correctly picked up by nosetests.

ahmadia:
Looks like [tests are passing](https://app.shippable.com/builds/54ab0ad942798d1100543f77).  I'm going to add Gauges to coverage and merge this.

proteus/tests/test_gauges.py
```diff
@@ -0,0 +1,131 @@
...
+    p.domain = Domain.RectangularDomain(name="test_gauges_domain")
+    p.coefficients = TransportCoefficients.PoissonEquationCoefficients(aOfX = [lambda x: np.eye(p.nd, p.nd)],
+                                                                       fOfX = [lambda x: 0], nc=1, nd=p.nd)
+    #Simplified and Incomplete Numerics
+    n.femSpaces = {0:FemTools.C0_AffineLinearOnSimplexWithNodalBasis}
+    n.elementQuadrature = Quadrature.SimplexGaussQuadrature(p.nd,3)
+    n.elementBoundaryQuadrature = Quadrature.SimplexGaussQuadrature(p.nd-1,3)
+    n.numericalFluxType = NumericalFlux.NoFlux
+
+    total_nodes = 2*comm.size()
```
ahmadia: We've verified that this works on up to 512 processes.  We may hit more problems at bigger scale.

PR #187, opened 2015-01-04
Comm calls init for you if that's already happened

ahmadia:
@cekees - Any reason not to merge this?

### DIFF
--- a/proteus/Comm.py
+++ b/proteus/Comm.py
@@ -29,7 +29,8 @@ def init():
     return new_comm
 
 def get():
-    assert(comm, "Call Comm.init() once before calling Comm.get()")
+    if comm is None:
+        return init()
     return comm
 
 class Comm():


cekees:
Seems OK to me. Should we warn that init hadn't been called?

ahmadia:
No, because this object is self-instantiating.  I don't see an explicit reason right now to force the user to know whether init has been called or not.

ahmadia:
Sorry, I misunderstood.  Yes, let's log this if it occurs.

ahmadia:
I'm a little concerned about the order of events with the `Profiling.logEvent` function, and that if you try to call it before the profiler has been set up, I believe it hard crashes.  For now I'm not going to add the log functionality here, but I'll raise an issue related to this.

ahmadia:
Ooph, I need another cup of coffee.  That function looks safe to me.  I'll open a new PR.

PR #195, opened 2015-01-14
Refactor Gauges interface to allow lines

ahmadia:
This is work-in-progress on the line gauges interface.  The idea here is to be able to scalably monitor integrals along lines and for fields specified by the user.  I've refactored the PointGauges constructor into a factory method, and created a LineGauges factory method as well, naming the generic class beneath "Gauges".  

My inclination is to use the trapezoid rule along each element for now, but it shouldn't be hard to choose the sampling to do something else. 

On the to-do:

* [x] Properly identify line intercepts with the underlying mesh.
* [x] Correctly calculate the values of the function at the intercepts
* [x] Use a numerical integration technique to approximate the integral along the line.
* [x] Ensure that gauge points added for "tracking" are not output
* [x] Add line gauge output
* [x] Verify correctness of the line gauges
* [x] Identify and resolve problems at 256 / 512 processes on Garnet
* [x] Add non-integral line gauges
* [x] Raise issues for deferred tasks (correct/selectable integration order on elements, kdtree for finding nearest nodes/elements, implement and compare a particle-tracing strategy, perform vectorized operations everywhere)
* [x] merge master in so shippable builds work.
* [x] Discuss multiple femspace issue with Chris/Matthew
* [x] Move everything from Geom into MeshTools/EGeometry


ahmadia:
I think that I've got segmentation working correctly (at least on the first trial mesh).  Still to do:

* [] Ensure that gauge points added for "tracking" are not output
* [] Add line gauge output
* [] Verify correctness of the line gauges

ahmadia:
@cekees - I could use some help reviewing/validating the line gauges for correctness / scalability bugs.  A couple of shortcuts I took to get this working:

* There's a pruning step after all intersections have been computed to verify that no processes have detected the same intersection.  This occurs on the boundaries of elements because I couldn't figure out a safe local uniqueness criterion (it should not be too hard).  This is a small barrier to scalability (it only happens when the Gauge is set up, but requires a global scatter/reduce that is O(p) in time and O(global_intersections) in space.

* I refactored the Gauge class to operate as either as PointGauge or a LineGauge.  This gets a little bit ham-fisted in a few places.  The line gauge needs most of the point gauge's operability, but it needs to behave slightly differently in that it doesn't track global points, which are needed for correctly ordering a point gauge, but are unneeded for a line gauge.  If others are going to be modifying this class in the future, it could use a refactor pass through to make the logic between the two gauge types a bit easier to comprehend and distinguish. 

* I'm only supporting simplices since I need to be able to convert element vertices to faces and don't know how to in the general case.  

* The code is untested for the 2D case.  My plan is to validate the 2D code before launching jobs on Garnet.

ahmadia:
This is very close to working.  I'm seeing some sort of subtle bug cropping up on Garnet for larger numbers of processes where the value of the integral is slightly incorrect, so this still needs some debugging.

ahmadia:
The bug is at 64-way parallelism.  I'm digging into it.

ahmadia:
Interesting, I think my deduplication algorithm isn't catching subtly different line segments.  I will loosen the tolerances for detecting duplicates as well as add a verification step to guarantee that the line segments are more or less non-overlapping.

ahmadia:
Some duplicate examples (this is from a post-run highly verbose log enabled by calling the test file directly):

```
[       5] Proc 12 : Processing segment [ 3.333333e-01 5.000000e-01 0.000000e+00 ] to [ 5.000000e-01 5.000000e-01 0.000000e+00 ]
[       5] Proc 12 : Processing segment [ 3.333333e-01 5.000000e-01 0.000000e+00 ] to [ 5.000000e-01 5.000000e-01 0.000000e+00 ]
[       5] Proc 2 : Processing segment [ 5.000000e-01 3.333333e-01 0.000000e+00 ] to [ 5.000000e-01 5.000000e-01 0.000000e+00 ]
[       5] Proc 2 : Processing segment [ 5.000000e-01 3.333333e-01 0.000000e+00 ] to [ 5.000000e-01 5.000000e-01 0.000000e+00 ]
```

ahmadia:
Testing again on Garnet.

ahmadia:
Okay, it looks like the deduplication fix is working.  I think this code is ready for review and merge.  I'm going to kick off several of the paper experiments with the line gauge code enabled.

cekees:
Great. I will take a look tonight or first thing tomorrow.

ahmadia:
Annoyingly enough, I'm now seeing failures up at 512 processes.  I'm guessing it's related to the same issue I was seeing where the mesh had an insufficient number of points per process.  I've cranked it up just a little bit and am re-running tests at 256 and 512.

ahmadia:
I'm not going to worry about it too much since I suspect the problem is coming from the design of the test and not the code that it's verifying.

ahmadia:
I'm verifying against one of the spilling problems.  There is still a little bit of trickery here in ensuring that the mesh is being correctly segmented.  Right now I'm encountering a situation where one of the segments to the boundary of the mesh isn't being identified (the intersection with the boundary of the mesh is likely not being detected).

ahmadia:
Confirmed that the problem is in the intersection code.  Intersecting over a mesh in the rectangular box:
[[0, 29], [0, 0.975], [0, 0]], The line segment defined by ((8.1199999999999992, 0, 0), (8.1199999999999992, 0.9750000000000001, 0)) is not intersecting with any elements between ~ 8.12, 0 and ~8.12, 0.0033149540517960971.

ahmadia:
Here's the nearest node in the mesh.
[  8.15896875e+00,   4.42857143e-03,   0.00000000e+00])

ahmadia:
Here are the elements it belongs to:

```
ipdb> mesh.elementNodesArray[np.nonzero(m)[0]]
array([[395,   1, 393],
       [391, 394, 393],
       [393, 394, 395]], dtype=int32)
ipdb> mesh.elementNodesArray[np.nonzero(m)[1]]
array([[477, 481,  16],
       [477, 481,  16],
       [400, 404, 395]], dtype=int32)
```

ahmadia:
Looks like the problem is that this mesh is not fully covering the rectangular domain.  The line segment is intersecting all elements on the mesh, but there's a gap here.

ahmadia:
```
ipdb> m = mesh.elementNodesArray == 393
ipdb> mesh.nodeArray[mesh.elementNodesArray[np.nonzero(m)[0]],:]
array([[[  8.01421985e+00,   2.29541614e-01,   0.00000000e+00],
        [  8.00400000e+00,   0.00000000e+00,   0.00000000e+00],
        [  8.15896875e+00,   4.42857143e-03,   0.00000000e+00]],

       [[  8.31393750e+00,   8.85714286e-03,   0.00000000e+00],
        [  8.23216272e+00,   1.56776878e-01,   0.00000000e+00],
        [  8.15896875e+00,   4.42857143e-03,   0.00000000e+00]],

       [[  8.15896875e+00,   4.42857143e-03,   0.00000000e+00],
        [  8.23216272e+00,   1.56776878e-01,   0.00000000e+00],
        [  8.01421985e+00,   2.29541614e-01,   0.00000000e+00]]])
```

ahmadia:
You can see the mesh following the boundary up until 8.004, at which it departs along the boundary ~8.15, 0.004

```
ipdb> mesh.nodeArray[mesh.elementNodesArray[np.nonzero(m)[0]],:]
array([[[  8.01421985e+00,   2.29541614e-01,   0.00000000e+00],
        [  8.00400000e+00,   0.00000000e+00,   0.00000000e+00],
        [  8.15896875e+00,   4.42857143e-03,   0.00000000e+00]],

       [[  8.01421985e+00,   2.29541614e-01,   0.00000000e+00],
        [  7.75387500e+00,   0.00000000e+00,   0.00000000e+00],
        [  8.00400000e+00,   0.00000000e+00,   0.00000000e+00]]])
```

ahmadia:
@cekees - Is this expected behavior?  Working around this is a bit tricky, as I'm not sure how to make the code smart enough to realize that the line segment is being drawn outside of the mesh itself, and then correctly crop the line segment to within the mesh, though I see how this could be handy.  For now, I'm going to reduce the tolerances on the line segmentation algorithm and warn when the difference in the segmented line length is greater than 0.1% of the length of the original line, and outright fail if the difference is more than 5%.  I just made both of these numbers up, so please let me know if you've got a better idea.

cekees:
I'm not sure I'm following. You have a rectangular domain and the generated mesh doesn't cover it? Can you plot the mesh?  On the other hand, it is expected that the line segment end points might not lie within an element just as the gauge points might not lie within the mesh.  I remember discussing that at one point with @adimako . I think we talked about just doing something sensible and warning. In the line case it makes sense to truncate the generated gauge points and segments to the intersections with the mesh. In the point gauge setting it seems like choosing the nearest node would be  OK, though in the long run we'd probably want to do something better  like find the nearest point on the boundary or extrapolate from the nearest element.

ahmadia:
> You have a rectangular domain and the generated mesh doesn't cover it?

Correct.

> Can you plot the mesh?

I have it, yes.  Do you have a quick one-liner for doing this using triangle or matplotlib?

> I think we talked about just doing something sensible and warning. In the line case it makes sense to truncate the generated gauge points and segments to the intersections with the mesh.

I agree, for the case I'm running (a low-resolution version of the plunging problem), one of the lines specified over the original domain is a full 5% shorter.  Since we're integrating over this area, I assume this leads in up to a 5% error in the vof line integral at this coordinate.  Is that a correct assessment, or am I being too cautious?

cekees:
for C0P1 FEM on triangles:
```Python
tricontourf(m.levelModelList[-1].mesh.nodeArray[:,0],#x
                 m.levelModelList[-1].mesh.nodeArray[:,1],#y
                 m.levelModelList[-1].mesh.elementNodesArray,#triangles
                 m.levelModelList[-1].u[0].dof)#solution
```

cekees:
The plunging problem isn't rectangular though. Is that the problem?

ahmadia:
Yeah, that's the problem.  I'm an idiot :)

ahmadia:
mesh

ahmadia:
I guess I can use these?

        vertices=[[0.0,0.0],#0
                  [L[0]*0.276,0.0],#1
                  [L[0]*0.96,(L[0]*0.96-8.0)/35.0],#2
                  [L[0],(L[0]-8.0)/35.0],#3
                  [L[0],L[1]],#4
                  [L[0]*0.96,L[1]],#5
                  [0.0,L[1]]]#6

cekees:
Yes, the vertices list that the domain object uses usually give you an idea of where the boundary is, though sometimes they include points inside the domain in cases where you want "hard points" in the mesh. That's what my old gauge approach did--inserted vertices that the mesh generator was forced to use.

cekees:
Anyway, no problem on the domain shape.  I should have put a plot of the domain in the README file for the test problem. @adimako and @ezve have done a great job with their README files in  air-water-vv, but I'm also looking forward to having a better  "gallery" view that makes it easy to  know what  to expect from a given test.

ahmadia:
I've generated some first images/visualizations at https://github.com/erdc-cm/adwr-nearshore-hydro/pull/42

ezve:
@cekees @ahmadia do you think it is a good moment to start using the gauge functions in the tests I run? As we run the cases in ORS, it would be helpful to have quickly transferable outputs. I could use them, also, to make the appropriate comparisons to validate them. @adimako

cekees:
@ezve I'll leave it  up to you about whether to start using the new  gauges. It would be safer to wait until  @ahmadia is able to reproduce some existing results extracted by post-processing the hdf5 in some real test problems. We don't know of any issues with this version of the gauges, but they aren't fully tested  yet so the  likelihood  is that there  will be a  few more issues.

adimako:
@ezve I think it is worth, for one case, to include the probe function and maintain for a high sampling frequency of the field data. If they compare well, you can continue using the functions without the high sampling frequency, if they don't, continue as you do now. Even if we discover bugs later, this will be eventually to our benefit

ahmadia:
There are bugs in the current version in master, so the PR here is definitely more trustworthy.  I'm confident we'll be able to merge this into master by the end of the week.

ahmadia:
The current approach to deduplication of code segments is too fragile.  It's not going to be sufficient to just look for intersections, I need to actually trace a line through.  The trick is going to be doing this in something less than O(n^2) time on the mesh.

proteus/Gauges.py
```diff
@@ -7,14 +9,84 @@
...
+            if point not in points:
+                l_d = {'fields': set()}
+                points[point] = l_d
+            # add any currently unmonitored fields
+            points[point]['fields'].update(gauge_fields)
+    return Gauges(fields, activeTime, sampleRate, fileName, points=points)
+
+
+def LineGauges(gauges, activeTime=None, sampleRate=0, fileName='line_gauges.csv'):
+    """ Create a set of line gauges that will automatically be serialized as CSV data to the requested file.
```
cekees: If I understand the code below, all line gauges are now line integral gauges. We're going to need both line integral gauges and line gauges that store the profile of the field over a line.

ahmadia: Can you explain what you mean by "profile of the field over a line"?

cekees: The gauges are extracting values of a field over a subset of the domain. The point gauges extract the value of the field at a point. The line gauges in general should extract the field over a line, which is what I meant by profile.  In some cases like the vof  we want to go ahead an integrate along the line, but that's just a special case. Often we want to plot the field along the line, which is just a 1D plot (field variable vs. arc length along line for example). Take a look at figures 10, 16, 20,.. in the draft paper.

proteus/Gauges.py
```diff
@@ -25,67 +97,55 @@ class PointGauges(AV_base):
...
+
+    def getLocalNearestNode(self, location):
+        # determine local nearest node distance
+        node_distances = np.linalg.norm(self.vertices - location, axis=1)
+        nearest_node = np.argmin(node_distances)
+        nearest_node_distance = node_distances[nearest_node]
+        comm = Comm.get().comm.tompi4py()
+        return comm.rank, nearest_node, nearest_node_distance
 
     def findNearestNode(self, location):
```
cekees: I think this needs to be moved  into MeshTools and replaced with a kd-tree implementation of the search for the nearest node. What do you think @mfarthin.

mfarthin: @cekees yeah, that would be nice to have in the mesh interface

cekees: An initial kd-tree implementation is already here https://github.com/erdc-cm/proteus/blob/master/proteus/MeshTools.py#L4509 so it would be good to refactor and reduce code duplication. Basically, any proteus mesh should be able to locate the nearest node in the mesh for a set of points in something reasonably close to  optimal. Since finding the containing element of the point is then just a small constant time addition (because the number  of elements "supporting" a node is effectively bounded by a small constant), each mesh should also be able to find the containing elements of a set of points in something close  to reasonably optimal.

proteus/Gauges.py
```diff
@@ -267,18 +351,139 @@ def buildGaugeOperators(self):
...
+            m.setUp()
+            self.lineGaugeMats.append(m)
+
+        # Assemble contributions from each point in each line segment
+
+        for lineIndex, (line, segments) in enumerate(zip(self.lines, linesSegments)):
+            field, endpoints = line
+            fieldIndex = self.fields.index(field)
+
+            # Trapezoid Rule to calculate coefficients here
```
cekees: For general integrated line gauges I think we'll have to use quadrature based on the order of the finite element space.  We're effectively creating a 1D mesh and integrating the finite element function over it, so most of the functionality is already implemented in Quadrature and FemTools. We just haven't ever applied to this case. It's not a high priority.

proteus/Geom.py
```diff
@@ -0,0 +1,228 @@
+"""
```
cekees: All this functionality should be moved  into MeshTools since these are all functions for operating on mesh elements and meshes. We also have an Euclidean Geometry module (EGeometry) for basic geometric  calculations, but the  functions in here are specific  to meshes.

ahmadia: Sounds good.  Will just be minor renaming.

proteus/Geom.py
```diff
@@ -0,0 +1,228 @@
...
+            if t_l < t_e:
+                # segment exits polyhedron before entering, no intersection
+                return None
+
+    assert(t_e <= t_l)
+
+    return [a + t_e*v_l, a + t_l*v_l]
+
+
+def getMeshIntersections(mesh, toPolyhedron, endpoints):
```
cekees: I agree that we can probably do better. @ahmadia, I think it would be really be helpful if you read the Haselbacher/JCP paper  that @mfarthin sent us in October.  The basic  idea is that if you know the element containing one endpoint then you can quickly march to the other endpoint locating the element boundary intersections as you go. This is needed in general for particle tracing with a velocity field  that  varies from element to element, but the line gauge is just a special case equivalent to having a uniform velocity field (the direction vector for the line segment). The paper basically presents the approach in two 20 line algorithms. We already compute the element-to-faces and faces-to-elements maps that it requires (elementBoundariesArray and elementBoundaryElementsArray)

proteus/tests/test_gauges.py
```diff
@@ -14,9 +14,9 @@
 from proteus import default_p as p
```
cekees: Since you've already accumulated quite a few tests, you should probably merge in master so shippable will start working for you again.

proteus/Geom.py
ahmadia: My primary concern with following the suggested approach was that we don't have the ability to quickly determine where the line segment intersects the local mesh (in general, the mesh will intersects the segment at some point on the interior of the segment in parallel, and the current algorithm for locating nearest the nearest point of intersection searches the entire local mesh space).

Since I don't have a search algorithm implemented better than O(m*n), there's no penalty to this approach (calculate all intersections), except that this approach is not robust against duplicate segments.  

I was also concerned about situations where the segment exited and re-entered the local mesh, since that didn't seem like something that could be efficiently handled in parallel without some work.

ahmadia: I've modified my approach slightly to deduplicate using a tracing approach, instead of the hashing approach I was using earlier.  I agree that implementing something akin to Hasselbacher will be of general use, but I'd need deeper knowledge in how the Proteus data structures work (and some time to refactor/test the code based on this approach, which would be a relatively big overhaul to the current code).  I don't think the scalability of this code is a real issue (10s of thousands of processors), but if it does become one, we should consider a functional approach to specifying the line gauges.  Instead of declaring the line gauges ahead of time, instead write a function that describes the desired spacing of the line gauges, and let the processes use the function to assemble local line gauges through the elements they own.  We'd also need to switch the I/O away from CSV to something we could write in parallel as well.

proteus/Gauges.py
ahmadia: I thought a left a comment to this effect earlier, but I understand this requirement better now and its on the todo.

proteus/Gauges.py
ahmadia: @cekees - I agree that this is something we want, but I'm going to defer this unless we need it, since it looks it's going to take me a little time to refactor the code in MeshTools if you need to preserve the functionality of the existing code.

ahmadia:
I've updated the task list at the top.

ahmadia:
With regards to the error showing up on 256 processors, I've decided that it's serious enough to figure out what's going on.  All tests pass up to 256, but at 256, a single point gauge pointed at the location: (0.5, 0.5, 0.5) on our test mesh no longer returns the correct value for our function, which is linear, and should be exact anywhere on an element approximating that value.  My initial suspicion was that we had encountered a situation where the process owning the nearest node didn't also own an element containing the point, but I think https://github.com/erdc-cm/proteus/commit/9d196813a7a220b4dded230ff323a3bf6532e945 should take care of that.

ahmadia:
Ooph, I made the mistake of not cleaning up the test files between runs, and silent failures are silent.  The bad commit is here: 9d196813a7a220b4dded230ff323a3bf6532e945

ahmadia:
This is verified at 512 processes on Garnet.

ahmadia:
Line (and line integral) gauges appear to be working correctly, untested at scale on Garnet.  Ping if more documentation is needed anywhere.

ahmadia:
See #204, #205, and #206 for carry-forward issues.

@cekees / @mfarthin - Lets get this merged.

proteus/Comm.py
```diff
@@ -2,10 +2,24 @@
...
 
+import ctypes
+import imp
+import os
 import sys
+
 import petsc4py
 from .Profiling import logEvent as log
 
+try:
```
cekees: Nice. That warning has been a source of confusion.

ahmadia: I just copied/pasted over from parun, I think the warning will still show up for any config files where the preload_libs environment variable is not specified.  But at least it's in a more reasonable place now.

proteus/FemTools.py
```diff
@@ -7740,3 +7740,228 @@ def __init__(self,
 #  #end checkFunctionTrace
     #check if contiuous solution has no jumps?
     #checkFunctionTrace(u,mesh,globalDGspace,traceOp)
+
```
cekees: OK, to clarify: my previous comment on these  was  to move them into MeshTools, not FemTools. There is a comment bit of abused  terminology in the FEM world that we continue to abuse. I "Finite Element" is the mathematical abstraction defined by the Ciarlet triple. We also use "Element" to to mean the mesh cells and ReferenceElement to mean the reference cell used  in the parametric finite element approach, where things are defined on a reference geometry and mapped  to the physical cells. So the stuff you're doing here is really operating on mesh cells and geometry, so I was thinking it's more natural to keep it  in MeshTools.py

ahmadia:
I'm just going to fix this now.

ahmadia:
@cekees: Should be fixed.

proteus/Comm.py
cekees: Oh, right.  I'm going to raise in issue if there's not one already.

cekees:
`test_gauges.py` is running for me on ubuntu linux

PR #197, opened 2015-01-17
changed url of hashdist.tgz  to dropbox

cekees:


ahmadia:
wow, that was fast.

ahmadia:
25 minutes from PR to build.  Not bad.

PR #202, opened 2015-02-05
stable/hydra installation

adimako:


adimako:
Renamed to config.py.centos, added mpi include and libs variables which solved the problem with mpi dependencies

PR #207, opened 2015-02-06
Add provenance tracking

ahmadia:
This should be safe to merge.  I've cherry-picked the commit off onto its own branch.


cekees:
Just to make sure I understand: is this just pushing the version info into text files? Or is it enabling a way to insert provenance into xdfm or gauge csv files?

ahmadia:
Those text files are now available from `proteus.version`.  It's going into the proteus `parun` logs now but there's no reason we couldn't also add it to xdmf and csv logs.

Issue #214, opened 2015-02-10
Proc=None causes profiling to go into infinite loop


robertsawko:
I am not sure at what point that happened but suddenly my proteus run that use `import proteus.iproteus *` freeze. It turns out that they go into an infinite loop whilst handling `preInitBuffer`. After closer inspection I realized that this is the problem is in `Profiling.py`:

```python
def openLog(filename,level,logLocation=None):
    ....
    for string,level,data in preInitBuffer:
        logEvent(string,level,data)

...
def logEvent(stringIn, level=1, data=None):
...
    elif procID==None:
        preInitBuffer.append((stringIn,level,data))
```
So assuming `preInitBuffer` was not empty it gets bigger and bigger. In my case it starts with `NO PROTEUS_PRELOAD_LIBS LOADED; CHECK config.py`. Also before my `procID` was equal to 0 and now it's 'None' which causes the other if to activate. I was looking around for a while, but couldn't understand the difference so I temporarily commented out the append line. I am afraid of side effects though. Could you please look at it, too?

ahmadia:
@robertsawko - Thanks for reporting this.  This looks like a bug in our implementation in the profiling implementation.  The situation you're describing is this:

* A user imports Proteus and starts calling Proteus functions that call logEvent (really this is a logging functionality, not profiling, but they're in the same module for historical reasons).

* Because the Profiling module hasn't been initialized, it starts off in "pre-initialization" mode, that is, log events are saved to a buffer to later be written out after MPI has been initialized.

* The buffers fill up and overflow

@cekees probably has more to say about this, but the solution I'd use is to set up the Profiling module when you launch your code.  We do this automatically for you if you use parun, but there's also an example in: https://github.com/erdc-cm/proteus/blob/master/proteus/tests/util.py#L5

You probably want to set loglevel to something much lower than 10, I think 3 or 5 is usually the right amount of verbosity.  Let me know if that helps!

In the meanwhile, let's leave this issue open.  I need to discuss with the other developers how they want to fix this issue, since the ideal solution is to call MPI initialization immediately on job launch.

cekees:
I'm not sure what's going on here. @robertsawko can you point me to your script? 

[...]

PR #217, opened 2015-02-10
assert Profiling.procID not None before calling Profling.openLog

cekees:
...fix  iproteus


cekees:
closes #214

cekees:
This is a short-term fix. Raised #218 for the long-term fix.

cekees:
@robertsawko could you check to see  if this fixes your problem?

robertsawko:
Checked it out and it works. Thanks.

PR #222, opened 2015-02-19
Parametrize tolerance

ahmadia:
It's also set to a very loose 1e-3.  Right now the algorithm tries to
pick "longest" segments when deduplicating intersections, but with this
loose tolerance it might make more sense to pick segments with the
"least" violation in some sense.


ahmadia:
I'm going to close this in favor of a slightly different approach.

ahmadia:
See #223

ahmadia:
This isn't great, but it's better than #223, which is failing for reasons I haven't had time to investigate.

PR #223, opened 2015-02-19
Prefer closer segments to longer ones

ahmadia:
Please don't merge this yet.  I'd like to test it out on some of the larger meshes that have been problematic in the past.


ahmadia:
See https://github.com/hashdist/hashstack/issues/650 for the shippable build error.

ahmadia:
This isn't a good solution either.

PR #229, opened 2015-02-27
Mfarthin/add read xml

mfarthin:
This isn't ready to merge yet.
Adds initial implementation of functions for reading mesh from xdmf in serial. Not intended to be a general solution yet. First goal right is to be able to generate hex mesh files from OpenFoam by first translating them to xdmf via foamMesh2vtk and then paraview. 

Functions are in `MeshTools.py` right now
 
TODO 
- [x] test reading simple hex mesh from vtk 
- [x] fix c0q1 poisson example with `RegularDomain`
- [x] fix c0q1 poisson example when read a hex mesh
- [ ] test reading simplicial meshes in xdmf


cekees:
@mfarthin , you mentioned that this is about ready to go in, didn't you?

mfarthin:
@cekees I added the poisson based tests for the hex mesh stuff to the test/ci. They run on my machine ...
Assuming Shippable says it's ok then I think it's good to merge

cekees:
OK, thanks. I've got to update the shippable image and insulated us from
changes in hashstack before the shippable badges are reliable again.

ahmadia:
@cekees - It sounds like you've got this in hand, but let me know if I can help on the HashDist isolation angle.  One option would be to let HashDist control the install of Proteus itself, and to tell HashDist to look locally for the Proteus revision to build/test.  You could then hardcode in a HashDist "stack" profile that just requires proteus and nosetests.  Like the ones we have in air-water-vv (see the pelican branch) and roams.

cekees:
@mfarthin The test is failing, but it runs fine on my machine when I just run that test. This is not the first time we've had  issues with strange behavior while running all the tests. I'm trying to debug it.

mfarthin:
Ok, thanks. Let me know if there’s something I can look at to fix it. Matthew

cekees:
I think that should do it for now. We may want to add some file handling tools  to proteus/tests/util.py to make our test file i/o cleaner. Right now  if you run nosetests proteus/tests you fill up your proteus/ dir with xdmf and log  files. We could probably write something simple that would make it simpler to reference test input data and also store output data in per test output directories.

mfarthin:
@cekees thanks for fixing the path.

PR #230, opened 2015-03-02
Refactor setup

ahmadia:
This removes the top-level `config.py` from version control.  From now on, all configuration changes are version-controlled in `proteus/proteus/config/`.  If the configuration changes are general, then they should go into the `default.py` module.  Otherwise, specialized "one-offs" can be activated in `__init__.py` by environment variable or hostname.  Most configurations of interest are handled by environment variable at run-time.  Exceptions to this include the preload libraries on Garnet/Copper.

@cekees - I'll test this on Garnet or Copper this week if you don't have any objections.


proteus/config/__init__.py
```diff
@@ -0,0 +1,6 @@
+import platform
+
+if platform.node().startswith('garnet') or platform.node().startswith('copper'):
```
ahmadia: Unfortunately, this doesn't appear to trigger on the Garnet compute nodes.  I may need to do a quick interactive debug session to see what's happening on the back-end here.

proteus/config/__init__.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+if 'HOSTNAME' in os.environ:
```
ahmadia: Now there is a `config` subpackage within `proteus`.  It has a default configuration that works on OS X, Linux, and Cygwin (maybe :) when HashDist is the build agent.  You can look through the `default.py` file and see that for the most part, so long as the system as a sane `prefix`, whether it was built by HashDist or no, you should be able to build Proteus.  However, there are going to be systems where things are different or weird enough that the logic shouldn't go into default.py, and that's the purpose of this configuration file.  For now, I suggest checking the system hostname, or some other flag available to the Python run-time for accessing this configuration file.  In this example, the `HOSTNAME` environment variable is set on both the front-end nodes (build-time) and the compute nodes (run-time), so it's a straightforward way to detect that Proteus is running on Garnet/Copper.

ahmadia:
@cekees - This is tested and working on Garnet, and this should work on Copper as well (I tested that the `HOSTNAME` environment variable was present/available).

This is ready for review/merge.

cekees:
Looks like the proteus-mprans and friends will have to be updated as well:
```bash
$ pip install .
Unpacking /home/cekees/proteus-mprans
  Running setup.py (path:/tmp/cekees/pip-08Ferk-build/setup.py) egg_info for package from file:///home/cekees/proteus-mprans
    Traceback (most recent call last):
      File "<string>", line 17, in <module>
      File "/tmp/cekees/pip-08Ferk-build/setup.py", line 21, in <module>
        raise RuntimeError("Missing or invalid config.py file. See proteusConfig for examples")
    RuntimeError: Missing or invalid config.py file. See proteusConfig for examples
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 17, in <module>

  File "/tmp/cekees/pip-08Ferk-build/setup.py", line 21, in <module>

    raise RuntimeError("Missing or invalid config.py file. See proteusConfig for examples")

RuntimeError: Missing or invalid config.py file. See proteusConfig for examples

----------------------------------------
Cleaning up...
Command python setup.py egg_info failed with error code 1 in /tmp/cekees/pip-08Ferk-build
Storing debug log for failure in /home/cekees/.pip/pip.log
```

ahmadia:
Have you pulled master from `proteus-mprans`?  I merged the two PRs at the same time.

ahmadia:
Oops, I'm confusing this with another PR...

ahmadia:
Merge this concurrently with https://github.com/erdc-cm/proteus-mprans/pull/54

ahmadia:
Thanks for testing and catching that @cekees

PR #231, opened 2015-03-05
Refactor offsets into top-level Transport function

ahmadia:
Also, interleaved DOFs are now used consistently in serial and parallel code.

### DIFF
--- a/proteus/ADR.pyx
+++ b/proteus/ADR.pyx
@@ -798,21 +798,14 @@ class LevelModel(proteus.Transport.OneLevelTransport):
         log(memory("TimeIntegration","OneLevelTransport"),level=4)
         log("Calculating numerical quadrature formulas",2)
         self.calculateQuadrature()
-        #lay out components/equations contiguously for now
-        self.offset = [0]
-        for ci in range(1,self.nc):
-            self.offset += [self.offset[ci-1]+self.nFreeDOF_global[ci-1]]
-        self.stride = [1 for ci in range(self.nc)]
-        #use contiguous layout of components for parallel, requires weak DBC's
+
         comm = Comm.get()
         self.comm=comm
         if comm.size() > 1:
             assert numericalFluxType != None and numericalFluxType.useWeakDirichletConditions,"You must use a numerical flux to apply weak boundary conditions for parallel runs"
-            self.offset = [0]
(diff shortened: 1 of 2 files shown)


proteus/Transport.py
```diff
@@ -1493,21 +1493,14 @@ def findReferenceComponent_phi(diffusion):
...
         if comm.size() > 1:
             assert numericalFluxType != None and numericalFluxType.useWeakDirichletConditions,"You must use a numerical flux to apply weak boundary conditions for parallel runs"
-            self.offset = [0]
-            for ci in range(1,self.nc):
-                self.offset += [ci]
-            self.stride = [self.nc for ci in range(self.nc)]
-        #
+
+        self.setupInterleavedFieldStrides()
+
```
cekees: I don't mind pushing this into a function, but we do  need to preserve the ability to do both layouts, particularly in Transport.py.  The contiguous approach you deleted will always work, while the interleaved approach will break for  strong Dirichlet conditions and mixed trial spaces for multi-component PDEs.

ahmadia: Sorry, I misunderstood your comment yesterday.  I'll fix this later tonight to allow the application code to switch to a contiguous layout.

ahmadia: I redid the PR.

proteus/ADR.pyx
```diff
@@ -798,21 +798,14 @@ class LevelModel(proteus.Transport.OneLevelTransport):
...
         if comm.size() > 1:
             assert numericalFluxType != None and numericalFluxType.useWeakDirichletConditions,"You must use a numerical flux to apply weak boundary conditions for parallel runs"
-            self.offset = [0]
-            for ci in range(1,self.nc):
-                self.offset += [ci]
-            self.stride = [self.nc for ci in range(self.nc)]
-        #
+
+        self.setupInterleavedFieldStrides()
+
```
cekees: just needs the name updated

cekees:
This should work for now. At some point we should add both some logic to test whether the dof can be interleaved  and fall back to the contiguous layout if  not as well as allow the user to manually chose the layout.

ahmadia:
Redone once more.  I think everything is fixed.

cekees:
Looks good.

PR #235, opened 2015-03-19
fixed the field split strides for interleaving components

cekees:
@ahmadia  here's a start on fixing the strides. They were set up for end-to-end  stacking so when you switched  proteus to  consistently interleave when possible, the field split preconditioners broke.  I believe we can actually get a reference  to  the  model via L so we could do this more generally and  switch automatically based  on the strides/offsets of the underlying discrete PDE from which L is  generated.


cekees:
@ahmadia I tested this in parallel on various levels of refinement up to 114073 triangles in 2D, and it seems to be working well. There is a difference in what I've tested and what you were reporting, namely additive, multiplicative, and schur with a11  all work. The a11 preconditioner is significantly less effective than selfp, though, which fits with what we discussed with @jedbrown. I also  noticed that the petsc docs say selfp is actually lumped, which I think should make it more robust than just the raw diagonal, given that we don't  lump the mass matrix itself. I'm going to go ahead an merge this in. We can re-open if you were  able to do any tests this week and found any issues.

jedbrown:
Lumping is optional (see MatSchurComplementSetAinvType()) and off by
default.  @karpeev, is this the default you intended?  I wouldn't say
one is better than the other, but the docs should be consistent.

karpeev:
I don't think the docs indicated one way or the other, which is an omission.  
In any event, I [pushed] (https://bitbucket.org/petsc/petsc/commits/0a4569348d06e683c17ae77db1dd4b4985071d19) what I hope is a clarification on when lumping is used and how to turn it on/off.

PR #236, opened 2015-03-24
Cekees/explicit hashdist dep

cekees:


ahmadia:
This looks fine to me.  We've had this conversation before but I wanted to note that this is 90% morally equivalent to using gitmodules to track a specific commit of hashdist/hashstack.

cekees:
Yes, I considered that but decided it wasn't worth creating tracking branches. With this approach we just upgrade proteus to a new hashdist or hashstack commit via a proteus  PR's and it's explicit what we're  using.

PR #237, opened 2015-03-25
removed nested if  with no else to load default

cekees:
@ahmadia make sure this is what you wanted. Not  sure why it wasn't getting triggered on other systems


ahmadia:
Yup, nice catch.

ahmadia:
@cekees - The reason it wasn't causing a problem on other systems is because `HOSTNAME` isn't a commonly defined environment variable :), that's `hostname`'s job.

PR #239, opened 2015-03-25
Adimako/wave tools random

adimako:
I am trying to finalise the random waves. I noticed that using the Monochromatic waves class from the random waves makes things dead slow. So time series is reconstructed and calculated directly in the RandomWave class. Anyway, @ezve will test it in a 2D flume and we will let you know


cekees:
Thanks @adimako . There's s relatively easy way to optimize the python version by making the module a Cython (.pyx) module and adding some type information. I was thinking we'd wait  until the implementation is reasonably settled. See http://cython.org/ if you're interested. It's now a fairly standard approach.

cekees:
Also, I reconfigured the the continuous integration so that it should be more stable from now on. The failed build above  is real, and you can view it via the Details link. Something about VTIME.

adimako:
It is a good thing that we will optimise it with cython. However, I prefer rewrote the velocity part to be more compact. I fixed the VTIME, let's see

PR #242, opened 2015-05-02
load lapack,blas headers from Accelerate.h

minrk:
vecLib/clapack.h doesn't work on OS X 10.10, and I think has been deprecated for some time.

I'm not sure for how many versions Accelerate.h works here, but I think it's at least a few.

This was the only patch I needed for proteus to build on 10.10.

### DIFF
--- a/proteus/config/default.py
+++ b/proteus/config/default.py
@@ -20,8 +20,8 @@ platform_lapack_integer = None
 if sys.platform == 'darwin':
     platform_extra_link_args = ['-framework', 'Accelerate']
     platform_lapack_integer = '__CLPK_integer'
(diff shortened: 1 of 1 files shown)


ahmadia:
let me check quickly on 10.9

ahmadia:
yup, looks good.

minrk:
I *think* Accelerate.h should be good back to 10.6, but I'm not certain.

ahmadia:
It's up to @cekees on what he wants to continue to support, but I tend to merge if it works on at least 10.9, and at this point, if I'm pretty sure it won't break 10.10.

ahmadia:
What are you using for your Fortran compiler, homebrew gfortran?

minrk:
yup

PR #252, opened 2015-05-26
re-enabled non-interleaved DOF so mixed spaces work again

cekees:


cekees:
This is  still going to break if we have different numbers of DOF globall (e.g. c0p1 and dgp1).

cekees:
I'm going to  go ahead an merge this but #253 flags the issue for fixing this permanently

PR #255, opened 2015-05-27
README: https link to landing page

ahmadia:


cekees:
Thanks. I'm not  sure what's going on with the proxy server. http was supposed to remain accessible when we moved behind it.

PR #257, opened 2015-06-11
fixing some corner cases that come up in intialization of fields and …

cekees:
…bc's


cekees:
@mfarthin and @srobertp This PR should fix the issue where the velocity is not initialized  the intial conditions on the first pass through a sequenntially split model sequence. We found that some of the velocities were getting set to zero.  I also fixed  an issue where a problem with no advection (but diffusion and hamiltonian operators) would generate an error. Some of  the options I added  to @srobertp's  test in proteus-ps won't run without this.

PR #261, opened 2015-06-16
tries to fix analyticalSolutionVelocity logic

mfarthin:
should allow user to only specify a single velocity component that has an analytical solution

### DIFF
--- a/proteus/SimTools.py
+++ b/proteus/SimTools.py
@@ -671,7 +671,9 @@ class SimulationProcessor:
             hasAnalyticalSolution[ci] = (self.analyticalSolution.has_key(ci)  and
                                          self.analyticalSolution[ci] != None)
             hasAnalyticalSolutionVelocity[ci] = ('analyticalSolutionVelocity' in dir(p) and
-                                                 p.analyticalSolutionVelocity != None)
+                                                 p.analyticalSolutionVelocity is not None and 
+                                                 p.analyticalSolutionVelocity.has_key(ci) and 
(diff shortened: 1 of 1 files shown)


cekees:
Thanks. I'm merging this, but one thing I should point out for @srobertp is that `has_key` is not in python 3. Instead you test for dictionary keys using a test like

```
dictionary = {'a':1,'b':2}
if 'a' in dictionary: # instead of dictionary.has_key('a')
```

srobertp:
@cekees and @mfarthin Thanks for poinying that out,  i have seen both approaches but hadn't realized they were doing the same thing.  Why don't we switch it to the more "pythonic" way using `in` ?  Is there ever any plan to support python 3? with proteus?  is the has_key approach O(1) like 'in'?

cekees:
Sounds good. Yes, in and has_key are equivalent. We'll provide python 3 support...next year:)                                                                                                                                                                                                                                                                          Sent from my BlackBerry 10 smartphone on the Verizon Wireless 4G LTE network.                                                                                                                                                                                                                                         From: srobertpSent: Wednesday, June 17, 2015 1:22 PMTo: erdc-cm/proteusReply To: erdc-cm/proteusCc: Chris KeesSubject: Re: [proteus] tries to fix analyticalSolutionVelocity logic (#261)@cekees and @mfarthin Thanks for poinying that out,  i have seen both approaches but hadn't realized they were doing the same thing.  Why don't we switch it to the more "pythonic" way using in ?  Is there ever any plan to support python 3? with proteus?  is the has_key approach O(1) like 'in'?

Issue #104, opened 2014-06-19
study and  refactor XDMF i/o


cekees:
Just following up on a discussion with @ahmadia. We suspect that the current approach of writing one xml and one h5 file per mpi  task will limit our scaling pretty soon as we go to 1000s of mpi tasks. I don't think it would be that hard to refactor the current Archiver to write only a single xml and h5 file per  simulation. Here are a few approaches

- communicate and compute all the metadata needed for the global  xml metadata on the root process
- figure out how to do parallel h5 writes from tables or switch to h5py for  the heavy data  writes
- instead of having the  root process  write  the xml metadata we could turn off writing the xml data during the simulation completely and just reconstruct the xml metadata file from the hdf5 as a post-processing step. This might require storing more data in the hdf5 file together with a more explicit set of conventions and assumptions for how proteus  lays out the mesh and fields in h5.
- another idea  I've toyed with is  switching the logging to YAML or  something like that and storing more of the metadata in a yaml log. That could then be used to help generate the xml at the end of the simulation or in a post-processing step.

References:

- http://xdmf.org/index.php/XDMF_Model_and_Format
- http://www.hdfgroup.org/HDF5/Tutor/poverview.html
- https://github.com/PyTables/PyTables/issues/216
- http://docs.h5py.org/en/latest/mpi.html

PR #264, opened 2015-06-25
added capability use single xdmf file

cekees:
This combines 1) mpi gathers of the new metadata at each time step so it can be
written to the global xmf file 2) use of parallel hdf5 via h5py rather than
pytables to handle writing a single h5 file.

### DIFF
--- a/proteus/Archiver.py
+++ b/proteus/Archiver.py
@@ -318,6 +318,7 @@ class AR_base:
         log("Done Gathering Archive Time Step")
     def sync(self):
         log("Syncing Archive",level=3)
+        memory()
         self.allGatherIncremental()
         self.clear_xml()
         if not self.useGlobalXMF:
@@ -347,6 +348,8 @@ class AR_base:
                 self.comm.barrier()
             else:
                 self.hdfFile.flush()
+        log("Done Syncing Archive",level=3)
+        log(memory("Syncing Archive"),level=4)
     def create_dataset_async(self,name,data):
         comm_world = self.comm.comm.tompi4py()
         metadata = comm_world.allgather((name,data.shape,data.dtype))
--- a/proteus/Archiver.py
+++ b/proteus/Archiver.py
@@ -31,8 +31,14 @@ class ArchiveFlags:
     UNDEFINED            =-1
 #
 class AR_base:
-    def __init__(self,dataDir,filename,useTextArchive=False,gatherAtClose=True,useGlobalXMF=True, hotStart=False,readOnly=False):
+    def __init__(self,dataDir,filename,
+                 useTextArchive=False,
+                 gatherAtClose=True,
+                 useGlobalXMF=True,
+                 hotStart=False,
+                 readOnly=False):
         import os.path
+        import copy
         self.useGlobalXMF=useGlobalXMF
         comm=Comm.get()
         self.comm=comm
@@ -48,6 +54,7 @@ class AR_base:
         try:
             import h5py
             self.has_h5py=True
+            comm_world = self.comm.comm.tompi4py()
         except:
             self.has_h5py=False
         try:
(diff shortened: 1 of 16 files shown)


cekees:
This PR is to address #104.  For now I've implemented it so that the single xmf file is created by default. Writing all the submdomain xmf can be turned back on with `parun -s`.  Whether or not a single hdf5 file is written depends on whether you have h5py installed. It can be added  to  your stack by adding `h5py:` to `stack/examples/proteus.linux2.yaml` 

There's a lot of additional work that  can be done such as cleaning up `Archiver.py`, writing unit test, and naming conventions in the  h5.

Bigger issues are whether we transition to writing the global meshes  and data  to hdf5 instead of the subdomain meshes and whether to go ahead and add enough attributes to the datasets to make the xmf redundant, in which case we could generate  it efficiently at the end of the simulation or as a post-processing step.

cekees:
@mfarthin could you take a look this PR? I think this  could go  in as is. I can update hashstack so the proteus stack builds with h5py by default but it  should work with the old  stack with pytables. One more think I'd like to do is stop rewriting the entire xml tree everytime we sync, but I'm not sure that's  really possible given the way the temporal collections are organized.  In any case I think this  PR will give a big improvement in our i/o.

mfarthin:
Sure, I’ll take a look at it this weekend. Matthew

cekees:
After some testing on copper, I'm discovering that we're now taking an even bigger memory hit for keeping the entire element tree in memory because now it is the full global xml tree (has all subdomains). This hurts on runs with many time steps and many cores.  xml is just not nice for long time series and incremental updating. I'm considering adding the xml elements to the hdf5 and then deleting prior time steps in the tree. The idea would be for the xml tree to only contain the last time step written. We'd still carry the tree in memory and rewrite a complete xml file at every write step, but it would only contain the latest information. To construct the full time series we'd have to open the hdf5 and write a new xdmf, much like we do now with gatherArchives, except it should be a lot faster and less memory intensive.

srobertp:
@cekees: A note on your automatic gathering times commit 71cb75d1d2950aca6deb539fffdb011608b72e81

I am running into the following problems when I try to run the current proteus_ps system on command line using the parun function with proteus on the nsps_numerical_fluxes branch which is an offshoot of this parallel_xdmf  branch:  After all the steps have been run and it is time to archive the solutions, the new function 

gatherAndWriteTimes() in Archiver.py is called.  There are a number of issues that pop up there in.

1. (line 215) the .__get_items__ call [] must be replaced with a .getNode() function call
             # grid_array = self.hdfFile["/"+dataset_name]
                grid_array = self.hdfFile.getNode("/"+dataset_name)

2.  (line 214)  The following lines
                dataset_name = TemporalGridCollection.attrib['Name']+"_"+`i`
                dataset_name = dataset_name.replace(" ","_")

end up with dataset_name = "Mesh_Spatial_Domain_*" and then "Mesh_c0p2_Lagrange_*"  neither of which exist in the hdf File table.  However the following names are available in the table:  "nodesSpatial_Domain*" and "nodes_c0p2_Lagrange*"   (where * = 0 or 1 based on which mesh we are looking at)  Did we mean these without the spaces? 

3.  (line 218) the attribute of 'Time' in  grid_array.attrs['Time'] doesn't currently exist for any entry of the hdfFile table, so either it needs to be added or for now this line needs to be commented out.  Also the variable 'time' set here is not used again

I would have changed all these things myself, but I wasn't sure about what the desired result was for issue 2 and 3.

-Spencer

cekees:
Thanks for reporting this. Did you add  `h5py` to your `stack/examples/proteus.Darwin.yaml` file? I'm guessing I  need to go back and test things without h5py. I tried to make the branch work with both h5py and pytables, but I wouldn't be surprised if I missed something.

srobertp:
the above errors were with the pytables without the h5py in my stack.  I will try it with the h5py in my stack tomorrow.

srobertp:
@cekees  It appears to run just fine when I added the h5py package to my stack.  So I guess it was the fact I was using pytables instead of the h5py extension.  Do we need compatibility with pytables there?  Or is h5py going to make it into the permanent stack?  Which one is used for sequential runnings?  Will the mpi running be set differently than sequential?

srobertp:
@cekees  I should qualify that.  On my mac it runs no problem but the same code on the linux machine triggers the following error

Traceback (most recent call last):
  File "/home/srobertp/proteus_files/proteus/linux2/bin/parun", line 452, in <module>
    runName + '_run_prof')
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/proteus/Profiling.py", line 186, in __call__
    return func(*func_args, **func_kwargs)
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/proteus/NumericalSolution.py", line 819, in calculateSolution
    self.closeArchive(model,index)
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/proteus/NumericalSolution.py", line 1002, in closeArchive
    self.ar[index].close()
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/proteus/Archiver.py", line 239, in close
    self.gatherAndWriteTimes()
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/proteus/Archiver.py", line 216, in gatherAndWriteTimes
    grid_array = self.hdfFile["/"+dataset_name]
  File "/home/srobertp/proteus_files/proteus/linux2/lib/python2.7/site-packages/h5py/_hl/group.py", line 153, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5o.pyx", line 183, in h5py.h5o.open (h5py/h5o.c:3602)
KeyError: "Unable to open object (Object 'meshspatial_domain_0' doesn't exist)"

srobertp:
The code works for linux as well as mac with the h5py package.  I had left a snippet of code in Archiver.py when I was debugging for the pytables version that removed the spaces " " in the name instead of changing them to underscores "_".  After fixing that back, it does run with the h5py as expected on the mac and linux machines.  There are still issues when we are using pytables without h5py.

PR #282, opened 2015-08-18
add new flux HamiltonJacobi_Advection_pN_DiagonalLesaintRaviart_Diffu…

srobertp:
…sion_SIPG_exterior for conservative pressure terms in the projection navier stokes methods in the velocity component

@cekees: @mfarthin :
I wish there were a way to include a switch to turn on and off this two line adjustment based on whether we want the advective flux or not.  Right now we have two of essentially the same Fluxes,  the standard one and the one that includes the two line advective adjustment of adding the term  'f' dot 'n'  to the existing HamiltonJacobiFlux, 

Is there a way to combine these into one functional Flux that can determine whether or not to add the term 'f'dot 'n' or not?

### DIFF
--- a/proteus/NumericalFlux.py
+++ b/proteus/NumericalFlux.py
@@ -3372,6 +3372,7 @@ class HamiltonJacobi_DiagonalLesaintRaviart_Diffusion_IIPG(NF_base):
                         #                                                                    self.penalty_floor)
 
 class HamiltonJacobi_DiagonalLesaintRaviart_Diffusion_SIPG_exterior(Diffusion_SIPG_exterior):
+    useStrongDirichletConstraints=False
     def __init__(self,vt,getPointwiseBoundaryConditions,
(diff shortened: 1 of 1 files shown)


srobertp:
I merged the two numericalFluxes into one with a check to see if ('f',0) is a key in ebqe and has the right shape, meaning that was initialized,  in which case we add the extra "advection" term  ebqe[('f',0)]*ebqe['n']).sum(-1)  representing (p * \n ) to the existing HamiltonJacobiFlux term.

srobertp:
I changed it around and added the above fixes into pull request #283. the numericalFlux HamiltonJacobi_Pressure_DiagonalLesaintRaviart_Diffusion_SIPG_exterior for velocity that includes the conservative term of pressure inherits from the numericalFlux HamiltonJacobi_DiagonalLesaintRaviart_Diffusion_SIPG_exterior and two terms are added in to account for the grad pressure integration by parts.

PR #292, opened 2015-09-16
[WIP] Refactor Domain

tridelat:
This PR is for cleaning and refactoring the module Domain.py so it can accept any number of predefined shapes (Shape.py) with associated boundary conditions (BC.py). This will make it easier to set up a simple domain with an arbitrary number of shapes in a more automatic fashion and should not affect the original way of defining a domain.

WIP on the following 4 files:
- [ ] **Domain.py**
  - [ ] cleaning of module
  - [x] refactoring for accepting predefined shapes (only done for PSL and PLC domains so far)
  - [x] list of all boundary conditions accessed by flags as index in p and n files
- [ ] **SpatialTools.py** - predefined shape classes
  - [x] predefined shapes
  - [x] custom shapes (by passing vertices, etc)
  - [x] add functions for spatial operations (rotation, translation, etc.)
  - [x] automatic adding of vertices, segments, facets, flags, etc
  - [x] rigid/floating body calculations (python) as auxiliary variable (in mprans.SpatialTools)
  - [ ]  rigid/floating body calculations using external package (e.g. ChronoEngine) with collision detection
- [ ] **BC.py** - predefined boundary conditions associated to shapes
  - [ ] modify/improve current BCs
  - [ ] ~~make initial condition class~~ (might be added in other module)
- [ ] **test_domain_shape.py** - testing of the new functionality, and checking if nothing is broken
  - [ ] fully verify that the new way of defining domain is working
  - [ ] fully verify that the original way of defining domain is still working

A working example of a full case showing the alternative way to define a domain will also be pushed in the  future.


proteus/Domain.py
```diff
@@ -17,6 +19,17 @@ def __init__(self,nd,name="defaultDomain",units="m"):
         self.units=units
         self.x=[]#minx,miny,minz
         self.L=[]#bounding box when self.x is origin
+        self.g=g
```
cekees: @tridelat can you explain why you're attaching gravity to the domain? Generally we're trying to separate the geometry from the other  physical parameter inputs, so I might want to find a way to pass gravity where you need  it without adding to the domain classes.

proteus/Domain.py
```diff
@@ -17,6 +19,17 @@ def __init__(self,nd,name="defaultDomain",units="m"):
...
+        # list of shape instances automatically attached to domain
+        self.shape_list = []
+        # list of auxiliaryVariables automatically attached to domaim
```
cekees: I think we want the parameters below to  go somewhere else as well, if possible.

proteus/Domain.py
```diff
@@ -227,36 +265,40 @@ class PlanarStraightLineGraphDomain(D_base):
...
-            self.regions=regions
-            self.vertexFlags=vertexFlags
-            self.segmentFlags=segmentFlags
-            self.regionFlags=regionFlags
-            self.regionConstraints=regionConstraints
-            if self.regionConstraints != None:
-                assert(self.regionFlags != None)
-            if self.vertices != None:
+            self.polyfile = None
+            self.vertices = vertices or []
```
cekees: nice!

proteus/Domain.py
```diff
@@ -692,67 +723,56 @@ def __init__(self,fileprefix):
...
 
+
 class PiecewiseLinearComplexDomain(D_base):
     """
     3D domains desribed by closed surfaces made up of general polygonal facets.
     """
-    def __init__(self,fileprefix=None,vertices=None,facets=None,
-                 facetHoles=None,holes=None,regions=None,vertexFlags=None,facetFlags=None,regionFlags=None,regionConstraints=None,
-                 name="PLCDomain",units="m"):
+    def __init__(self, fileprefix=None, vertices=None, segments=None, facets=None,
```
cekees: Using segments in 3D is adding a new concept to the PiecewiseLinearComplexDomain type, since these domains originally only define segments implicitly through the definition of facets (usually loops defining polygons but facets can also be degenerate points and segments).  How are the new explicit segments being used?

proteus/SpatialTools.py
```diff
@@ -0,0 +1,1469 @@
+"""
```
cekees: I think we should probably rename this mprans/SpatialTools.py or something like that. There is a lot of functionality in here, but much of it is specific to the two-phase modeling and rigid  body modeling.

proteus/BC.py
```diff
@@ -0,0 +1,340 @@
+"""
```
cekees: These should probably be moved to mprans since they're specific to two-phase air/water modeling.

cekees:
I still don't think I fully understand the new abstractions here. Domain classes represent domains of PDE's, generally bounded sets on R^n, but which require specific approaches to define concretely, such as intervals defining "boxes", planar straight line graphs, piecewise linear complexex, etc.  Is a Shape a Domain, a sub-component of a domain, or rigid body that forms part of a domain?  Boundary conditions are constraints on the solution values and fluxes along the domain boundary, so they need a domain and a PDE to be applied, but they can be defined abstractly based on flags and coordinates.  Can you explain the BoundaryConditions class you added  in terms of it's interface (member data and functions that operate on the data)? I'm not saying the functionality here isn't needed, I just want to make sure it's as simple as possible and maps clearly onto existing concepts.

proteus/Domain.py
tridelat: @cekees I attached it there but actually don't actually use it from there yet. 
As gravity is only defined once, I thought that it would be convenient if we need to access it somewhere else, such as predefined boundary conditions. This could be moved somewhere else ultimately.

tridelat:
@cekees A Shape is a sub-component of the domain, it could be a tank and other bodies within the tank, all having their set of vertices, segments, facets, etc. When the domain is built, it iterates through the list of shapes and adds their vertices, BCs, and auxiliaryVariables to itself, if any. 
The boundary conditions are accessed through flags that are used directly as indice in the `*_p.py` files. 
Every time a user defines a new shape, boundary conditions are associated with it and can be accessed directly from the Shape.
For example:
```
boundaryTags = { 'bottom': 1, 'front':2, 'right':3, 'back': 4, 'left':5, 'top':6, 'obstacle':7}
vertices=[[0.,0.,0.],#0
          ....
          [0.,1.,1.]]#7
vertexFlags=[boundaryTags['left'],
             boundaryTags['right'],
             ...
             boundaryTags['left']]
facets=[[[0,1,2,3]],
        ...
        [[4,5,6,7]]]
facetFlags=[boundaryTags['bottom'],
            ...
            boundaryTags['top']]
custom = st.CustomShape(domain=domain,
                        vertices=vertices,
                        vertexFlags=vertexFlags,
                        facets=facets,
                        facetFlags=facetFlags,
                        boundaryTags=boundaryTags)
# after defining the shape, you can access the boundary conditions like this
# (the keys in the boundaryTags dictionary passed to CustomShape must have acceptable variable names for it to work)
custom.BC.left.setNoSlip()
custom.BC.right.u_advective = lambda x, t: 0.0
# or also:
custom.BC_dict['top'].setFreeSlip()
# and even (if the index is known)
custom.BC_list[2].v_advective = lambda x, t: 0.0
```
Then, after building the domain, the BC is simply called like this in `*_p.py` files:
```
dirichletConditions = {0: lambda x, flag: domain.bc[flag].p_dirichlet,
                       1: lambda x, flag: domain.bc[flag].u_dirichlet,
                       2: lambda x, flag: domain.bc[flag].v_dirichlet}
```
The BC indice are automatically adjusted to global indice (from their local indice within shapes) when the domain is built.
I think it makes things much simpler to set up, as there would be no need to edit all the `*_p` files individually for BC, and we can keep useful predefined boundary conditions in BC.py (accessed through functions such as `setNoSlip()`)

proteus/Domain.py
tridelat: If preferable, we could create a new class, such as `World()`, where we can attach variables like `g` and the `shape_list`, and call the domain functions from there instead. For now, the `shape_list` is necessary there for me because there could be several rigid bodies or obstacles within a tank, and it is easier to set them up separately rather than pass the full set of vertices, segments, etc.

proteus/Domain.py
tridelat: It is actually not used in any functions of `PiecewiseLinearComplexDomain`, so it is not necessary to pass them, it does not break anything to have them there but should it be removed?

cekees: There are a couple of different options. I believe tetgen allows "singular" facets, that is facets that are just edges or points, so you could actually use the segments that are provided by turning them into facets. I think you'd need to do the  same with the segment flags. The quickest  solution is probably just to remove them or raise  a loud  warning that the segments are not  currently being used. I wouldn't want somebody to go to a lot of trouble defining segments and then go through a long debugging process to figure out whey their segments don't show up in the mesh.

proteus/Domain.py
```diff
@@ -1,7 +1,55 @@
 """
 A class hierarchy and tools for building domains of PDE's.
 """
+import numpy as np
+import BC
+from MeshTools import MeshParallelPartitioningTypes as mpt
 
+class Mesh:
```
cekees: There is already a Mesh class in MeshTools. I would name this class MeshOptions.

proteus/Domain.py
```diff
@@ -17,6 +65,19 @@ def __init__(self,nd,name="defaultDomain",units="m"):
...
+        # list of auxiliaryVariables automatically attached to domaim
+        self.auxiliaryVariables = []
+        # attach a Mesh class
+        self.Mesh = Mesh(self)
+        # For absorption/generation zones:
+        self.porosityTypes = None
+        self.dragAlphaTypes = None
+        self.dragBetaTypes = None
+        self.epsFact_solid = None
+
```
cekees: We don't want to mix physics-specific  quantities like g, porosity, etc. with the domain geometry.

proteus/Domain.py
```diff
@@ -28,6 +89,31 @@ def writeXdmf(self,ar):
...
+        """
+        self.x = []
+        self.L = []
+        for d in range(self.nd):
+            xMax = max(row[d] for row in self.vertices)
+            xMin = min(row[d] for row in self.vertices)
+            self.x += [xMin]
+            self.L += [xMax-xMin]
+
+    def setGravity(self, g=None):
```
cekees: This would also go somewhere else.

cekees:
I think I see where you're headed now. The Shape class is a good idea.  To get this ready to merge I think the following are still  needed. I'll discuss with @adimako today and maybe we can have a phone call tomorrow morning. 

- [ ] convert tests to either nose-based or standard  python tests  see test_meshtools.py or test_wavetool.py for examples
- [ ] move classes that are mprans specific  into the mprans  sub-package (I think of the new ones still  are at the moment. We can eventually make Shape and BC more general but for now we might want to move them)
- [ ] remove specific  physical information like porosity and gravity from the Domain module to keep it purely geometric

proteus/Domain.py
tridelat: I went for the quickest solution for now and removed segments for the PLC domain
(commit bb5e6143958cb8d59542f22e0d11a3db90583b1f)

proteus/Domain.py
tridelat: Do you mean setGravity or getBoundingBox?
Domains had their own getBoundingBox previously:
https://github.com/erdc-cm/proteus/blob/master/proteus/Domain.py#L267-L283
https://github.com/erdc-cm/proteus/blob/master/proteus/Domain.py#L726-L747

proteus/SpatialTools.py
tridelat: Yes I agree, that would be better. We could also keep classes such as Rectangle, Cuboid, CustomShape in here and some general functions that could be useful (e.g. rotate/translate), but make a module in mprans that would import everything from here, and add functionality to it related to two-phase flow and floating bodies to the Shapes (e.g. setAbsorptionZones, setRigidBody, etc). Let me know what you think.

tridelat:
@cekees SpatialTools and Domain have been cleaned up on commit 8a6f03e204bf28d61853596ca2395905b218c65f 
Functions and classes specific to mprans are now in a file also called SpatialTools. `super()` is now used in the classes so everything can be inherited (including default attributes) from the base proteus.SpatialTools.Shape while overriding or adding specific attributes/functions. I ran a simulation, and it works as expected when doing `from proteus.mprans import SpatialTools`. Let me know if this could be a solution.

cekees:
I think you may have already fixed it, but on the Domain functionality, it was just the setGravity function I wanted to remove. The getBoundingBox belongs there, as you said.

cekees:
@tridelat I  think your use of inheritance is great for SpatialTools and mprans/SpatialTools and should work  for BC as well if there's enough generic functionality in BC to make it useful to  have a proteus.BC module.

geomak89:
I observe that in BC.py file and specifically under the setGenerationFunctions method, I should change the previous definition of he as follows: he=self.domain.MeshOptions.he in order to be in accordance with the latest version of the new class MeshOptions. 
I changed the definition of he in my file but I don't have the permission to push this change.

tridelat:
@geomak89 that's right, it should be `MeshOptions`.  It seems that I cannot give permissions for someone to push on proteus so I pushed the change on the latest commit, and also moved `reset()` at the beginning in `setParallelFlag0()`.

proteus/mprans/Kappa.py
```diff
@@ -136,7 +136,7 @@ def __init__(self,LS_model=None,V_model=0,RD_model=None,dissipation_model=None,M
         self.variableNames=['kappa']
         nc=1
         self.nd = nd
-        assert self.nd == 3, "Kappa only implements 3d for now" #assume 3d for now
+        #assert self.nd == 3, "Kappa only implements 3d for now" #assume 3d for now
```
cekees: You can go ahead and remove that line if you've tested in 2D

cekees:
This looks great!  I think you should go ahead and merge.

proteus/mprans/Kappa.py
tridelat: It did not work very well during 2D simulations, with the time step becoming increasingly small, until the simulation crashes. I will uncomment the line before merging.

tridelat:
This branch depended on adimakoWaveTools/cython branch which has therefore also been merged

tridelat:
An example of SpatialTools, BoundaryConditions and WaveTools being used with proteus for two-phase flow can be found here:
https://github.com/erdc-cm/air-water-vv/tree/tridelat/caisson_oscillation/2d/floating_caisson

PR #293, opened 2015-09-16
fixes bug in saving of spatial and mass residuals for POD

mfarthin:
@cekees @MajinSaha There was a cut and paste typo in the archiving of the spatial residuals. The burgers example works a lot better now (DEIM looks great with 31 spatial modes and 25 solution modes)


cekees:
Great. I'll try it out.

PR #295, opened 2015-09-25
Missing file

adimako:
Not sure why I had a problem with this. I got an error message during openssl building


adimako:
@cekees I will go on and merge this
@tridelat This should work on guinode home folder, with some minor fixes and workarounds

PR #309, opened 2015-10-27
some updates for POD DEIM and Hyper reduction cleanup

mfarthin:
This is not ready to merge yet.

@MajinSaha I believe `cleanup_deim` is about ready for a pull request into master. This has some modifications that I'd like us to finish before doing a pull request though.

- [x] Implement 'general' nonlinear and linear splitting for residuals and jacobians
- [ ] get rid of spatial and temporal splitting between residuals and jacobins
- [ ] Combine `POD_Newton` and `POD_HyperNewton` into one class that can do full nonlinear POD or hyper reduction based on user input
- [x] Add tests for new functionality in `test_deim.py` and make sure all of the tests pass
- [ ] Put back general user-specified tolerance in POD Newton solvers 
- [ ] Switch initial condition to use projection from POD instead of fine grid solution

### DIFF
--- a/proteus/NonlinearSolvers.py
+++ b/proteus/NonlinearSolvers.py
@@ -1363,9 +1363,30 @@ class POD_HyperReduced_Newton2(Newton):
 
 	self.pod_Jtmp= np.zeros((self.nhyper_sample,self.DB),'d')
         self.rl = None; self.rn = None
-        #self.skip_mass_jacobian_eval = True
-        #self.linear_reduced_mass_matrix=None
         self.pod_initialized=True
+    def setup_POD_initial_conditions(self):
+        """
+        Set the Transport model to use the projection of the initial 
+        condition in the POD solution bases rather than the initial 
+        condition in the fine space
+
+        model degrees of freedom should come in initialized from fine space initial conditions
+
+        Work In Progress
+        """
+        
+        assert 'setFreeDOF' in dir(self.F)
+        assert  'setUnknowns' in dir(self.F)
+        import types
+        
+        #duplicate storage so that this can be called at the time the solver is constructed
+        assert os.path.isfile(self.SVD_basis_file), "SVD basis file {0} not found".format(self.SVD_basis_file)
+        U = np.loadtxt(self.SVD_basis_file)
+
+        F.project_initial_conditions = types.MethodType(deim_utils.project_initial_conditions,F,OneLevelTransport)
+        F.ic_projection_matrix = U.conj().T #fine to pod
+        F.ic_global_vector = self.du.copy()
+        
     def norm(self,u):
         return self.norm_function(u)
     def computeHyperResiduals(self,u,rl,rn):
@@ -1559,6 +1580,8 @@ class POD_HyperReduced_Newton2(Newton):
         for parm in optional_params:
             if parm in dir(nOptions):
                 setattr(self,parm,getattr(nOptions,parm))
+        #Go ahead and modify Transport behavior now that we have the correct input options
+        self.setup_POD_initial_conditions()#does not initialize all of POD machinery
         
 class NewtonNS(NonlinearSolver):
     """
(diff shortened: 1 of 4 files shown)


MajinSaha:
@mfarthin I saw you returned to previous methodology of subtraction of the lienar jacobian from the total jacobian in order to get nonlinear jacobian. I didn't use that on purpose, since this is less efficient than independent assembly of the nonlinear jacobian. But that is ok if our goal is to test method accuracy only.

mfarthin:
@majinsaha you are right. I was thinking we should get all the other pieces working well and then switch the nonlinear evaluation routine. For now, it just leaves us with fewer things to change and less ambiguity while we are working on the overall design.                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone on the Verizon Wireless 4G LTE network.                                                                                                                                                                                                                From: MajinSahaSent: Tuesday, October 27, 2015 11:33 PMTo: erdc-cm/proteusReply To: erdc-cm/proteusCc: mfarthinSubject: Re: [proteus] some updates for POD DEIM and Hyper reduction cleanup (#309)@mfarthin I saw you returned to previous methodology of subtraction of the lienar jacobian from the total jacobian in order to get nonlinear jacobian. I didn't use that on purpose, since this is less efficient than independent assembly of the nonlinear jacobian. But that is ok if our goal is to test method accuracy only.

mfarthin:
@MajinSaha I updated some of the tests in `proteus/tests/POD` I also started adding the functionality to perform the projection of the initial conditions. I think the easiest place to do this would be at the end of the setInitialConditions function in OneLevelTransport. I added some indirection to try to hide the fact that POD was being used from the Transport classes so that we don't have to propagate a lot of the information

For now, I'm trying to setup the initial condition projection from the `self.setup_POD_initial_conditions` function in the POD_HyperReduced_Newton2 class if you want to see what it's doing. I still haven't tested it yet

MajinSaha:
@mfarthin Just to make sure we are talking about the same thing. 
I checked your committed code. The projection of the initial condition is already done in function initialize_POD of POD_Newton2, at lines 831-832 of NonlinearSolvers.py of commit 38aeaa. Analogously for POD_HyperReduced_Newton2.
So there is no problem here. The only problem I meant was that the residual for Newton uses the mass term corresponding to the previous time level (or levels, if we use multi-step method). It is in that term that we need to have projection done. Hope you went in the right direction.

MajinSaha:
@mfarthin Let me then do removal of the spatial-temporal splitting and merge of POD_Newton2 and POD_HyperReduced_Newton2. One test would be to make sure POD_HyperReduced_Newton2, if use_hyper is set to false, to reproduce exactly the solution of POD_Newton2. If test passes, we'll remove POD_Newton2.

mfarthin:
@MajinSaha Yes, they are doing the same thing, but not at the same time. When initialize_POD gets called, it only updates the guess for the current solution, since that’s what’s coming into the solve function. In proteus, the time derivative approximations are handled by the a TimeIntegration class. These work by caching values of the mass of the solution from previous time steps. So, Backward uses in effect

q[(‘mt’,0)] = (q[(‘m’,0)] - self.m_last[0])/dt 

I believe the issue is just that the POD projection is not being done until after the time integration is initialized and has cached the m_last values. We could do the projection in initialize_POD and then repeat the timeIntegration and Transport initialization steps to reset the history, but this requires the POD Newton class to know how to repeat all of the Transport did initialization which could get out of date if we change things. It also would be easier to get something out of sync if not everything gets reset. 

If we can get the projection done right after the initial conditions are calculated analytically, then I think/hope everything should take care of itself.

mfarthin:
@MajinSaha That sounds good. Can you check that the code still reproduces your matlab code? I think that’s important. I have just been running the test_deim.py tests so far. Matthew

MajinSaha:
@mfarthin Sure, I was going to compare with matlab anyway.

mfarthin:
@MajinSaha I just tweaked the `test_deim` function `get_burgers_ns` to use more POD options. All of the tests should pass now (or at least they are passing on my machine)

PR #312, opened 2015-11-04
tweaks get_burgers_ns so that multiple imports don't affect test results

mfarthin:
This is not ready to merge yet. It's a reopening of pull request #309 

@MajinSaha I believe `cleanup_deim` is about ready for a pull request into master. This has some modifications that I'd like us to finish before doing a pull request though.

- [x] Implement 'general' nonlinear and linear splitting for residuals and jacobians
- [x] get rid of spatial and temporal splitting between residuals and jacobins
- [x] Combine `POD_Newton` and `POD_HyperNewton` into one class that can do full nonlinear POD or hyper reduction based on user input
- [x] Add tests for new functionality in `test_deim.py` and make sure all of the tests pass
- [x] Put back general user-specified tolerance in POD Newton solvers 
- [x] Switch initial condition to use projection from POD instead of fine grid solution


MajinSaha:
@mfarthin I think I completed two checkboxes. You're welcome to check.

mfarthin:
Great. I will run them today.                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone on the Verizon Wireless 4G LTE network.                                                                                                                                                                                                                From: MajinSahaSent: Thursday, November 5, 2015 1:42 AMTo: erdc-cm/proteusReply To: erdc-cm/proteusCc: mfarthinSubject: Re: [proteus] tweaks get_burgers_ns so that multiple imports don't affect test results (#312)@mfarthin I think I completed two checkboxes. You're welcome to check.

MajinSaha:
@mfarthin I restored the previous tolerance criteria in Newton solver for POD. Note that I also modified the last test in test_deim.py, since the proteus and matlab solutions now differ by error greater than machine precision because of the new criteria.

Once the projection of the initial data is implemented for the zero's time level term in the Backward Euler, we'll modify the matlab solution file to make another comparison.

mfarthin:
Hello @majinsaha,
I’ve been working some on the initial condition projection. What changes do we need to make to the matlab to do a comparison? Matthew

MajinSaha:
Hi @mfarthin !
I checked erdc-cm/pod-swe repository and the latest commit should be producing the solution with the projected term in the residual at the initial time step. So you can try to rerun matlab function npod and generate new solution file (just rename it from S_hyper to S_matlab). The parameters for the run are 100, 0.01, 1, 10, 10, 1.
I'll try to see what the problem is too.

mfarthin:
@MajinSaha I just added the option to project the initial condition. It seems to be passing the test for comparing with the matlab deim solution. However, I didn't go back and rerun the proteus example with the matlab nonlinear convergence criterion. If you want to try that to double check, I think we can go ahead and merge this pull request and then do a pull request for proteus master.

MajinSaha:
@mfarthin That's good. I'll pull your recent changes and look right now. Yes, I'd like to double check to be safe.

MajinSaha:
@mfarthin the test with my tolerance passed. The difference between two solutions fell below 10^-10. So I'll just merge.

PR #313, opened 2015-11-12
reduced number of calls to flcbdfWrappers.globalSum/.globalMax

cekees:
@smattis I'm not ready to merge this into master, but it should work for your  purposes. It would be  interesting to  see  a  comparison with whatever the most  efficient profiles you have on your problems.


cekees:
@smattis the changes to the pstats analysis still  don't really give us what we want because they sum times across  all processes. There are probably several views worth doing but maybe max totime for each func across  processes  would be better.

cekees:
@tridelat I think commit commit a423050 should fix the problem you so with the '-p'  causing runs to hang. The master process reads all the other processor's statistics from files so I needed to ensure that the other processors were actually finished writing before allowing the master  process to proceed.

PR #315, opened 2015-11-12
Test global hdf5 pwl only

cekees:
@smattis this branch contains  some parallel i/o  optimizations that I've been testing  on HPC systems for a while. It  can't be merged into master because it  only works for  linear nodal elements, but it  should work for your vegetation cases.  The basic  idea here is  to write  the global  (parallel) mesh directly to hdf5 as a single mesh using slicing rather than writing the subdomains to hdf5. You may a difference  in how paraview behaves and performs but you will still be getting a single hdf5 and xmf like we're already doing.  The reason this approach is  more effiicent is that  hdf5 requires that all processors write exactly the same metadata for each item stored. Every processor  has the metadata for the global mesh but not the metadata for all the subdomain meshes, so the master branch is  currently communicating that metadata. Writing the global  mesh directly avoids  communication of the metadata.


cekees:
Hot starting is not working on this branch yet.

PR #323, opened 2015-11-20
Cekees/wave tools/cython

cekees:
@adimako  The  only changes I made to WaveTools.pyx were in strings and comments, but don't worry about merging this into your if you've made subsequent changes to WaveTools that introduce conflicts. These are easy changes to re-apply later. I just wanted to see how the documentation was looking, which is  roughly like this:
[proteus.WaveTools module — Proteus 1.0.pdf](https://github.com/erdc-cm/proteus/files/40399/proteus.WaveTools.module.Proteus.1.0.pdf)
/cc @tridelat


adimako:
@cekees, I have edited comments in WaveTools my self, this is why the conflicts. I guess I can check for conflicts and then merge

PR #324, opened 2015-11-24
working placeholders for air/water/sed  submodels

cekees:


proteus/tests/test_sedclosure.py
```diff
@@ -0,0 +1,102 @@
...
+class TestHsu(unittest.TestCase):
+    def testGradularDrag1(self):
+        from proteus.mprans.SedClosure import HsuSedStress
+        # Constant params
+        aDarcy = 1.
+        bForch = 1.
+        grain = 0.1
+        packFraction = 0.2
+        packMargin = 0.01
+        uf = np.array([5,4])
```
cekees: @adimako I think the issue may be that uf and us are arrays of integers here but are  getting passed  to C++ as double arrays. You can either explicitly give  the dtype  as 'd' in the array() function or make it 5.,4. etc.  There's probably a way to be more specific in Cython to enforce the dtype of the ndarray's,

cekees:
https://github.com/erdc-cm/proteus/commit/4a048528f459448e2ffe39638e62a478f48aaeb0 runs and has (I think) all the correct terms in the equations but no constitutive relations beyond the linear and quadratic drag terms. The dambreak example  in proteus-sed runs. It still needs a lot  of cleaning up and possibly also  some tweaking of  the formulation. I haven't started  the projection scheme version yet.

adimako:
@cekees, the function now calculates \beta and not \beta (uf- us). Should we calculate the second term or leave it as it is, in case we want to solve this equation in a psedo-implicit manner?

cekees:
- [ ] Pass const. rel. parameters into coefficients
- [ ] ensure fluid and sediment inter-phase momentum transfer terms have equal magnitude and opposite sign
- [ ] provide lagged  velocity so inter-phase momentum transfer  can be made linear even with Forchheimer term
- [ ] enable sediment stress term
- [ ] enable k-eps model

cekees:
@adimako I'm still cleaning up and checking this branch. An example running just with air/water active is in https://github.com/erdc-cm/proteus-sed/tree/master/tests/dambreak . I still plan to come up with a simple "adapter" that will run exisiting input files with the projection scheme.  I'll go over some of the changes on this branch tomorrow in the weekly meeting.

cekees:
@adimako I noticed some places in the testGranularDrag* functions where it looks to  me like the drag2 value can only end up equal to 1.0 because right before the test we do `drag2 /= drag2`.  Is that what you were intending? The travis testing won't run on this branch until I update the testing stack to the new version of petsc...

adimako:
@cekees, I do not remember why I had written like this, but in the pervious line, I also divide  "drag /= drag2". "drag2" is tested against "drag" so if they are both 1, the test will pass. I think at the time I did it with the view to round it up easier, if needed, but it turned out that the test passed any way

cekees:
@mfarthin @adimako @alistairbntl @tridelat @manuel-quezada @Giovanni-Cozzuto-1989  this is not perfect, but I think I'm about ready to merge it in so some of you can try to merging with your branches this week while we're together and can more easily debug. It contains

- the Guermond/Salgado projection scheme extended to two and three phase flow
- The sediment closure relations implemented by @adimako 
- Some rearrangements of the testing tools but not the full cleanup and re-organization on our sprint todo
- And optimized nodal partitioning scheme used by the parun '-F' option (direct parallel partitioning of the tetgen input files and some out-of-core partitioning), which is required for very large meshes (>20M tets)
- The simplified fixed time stepping modes implemented by controllers in SplitOperator
- Fixes required to update superlu, petsc, and petsc4py to their latest versions (have to build superlu with petsc for now--stack/pkg/superlu.yaml hasn't been updated yet)
- Updates to the stack and a related Makefile target `make default_stack` that will bring your stack up-to-date with what is stored in .hashstack_default
- Updates to travis.yaml to use a docker container of Ubuntu 14.04, which starts up faster and is closer to what most of us run

PR #325, opened 2015-12-11
Proposed changes to BC.py

adimako:
@tridelat @cekees, these are proposed changes for the boundary conditions (or at least to start a discussion because I believe that some of them are over specified).

a) for setParallelFlag0, I believe everything should be "None"? Could it be that the issues we are having are related to this?

b) for kappa and epsilon (dissipation). In walls in general kappa = 0 and dissipation is d/dn = 0. But since we are not so far using wall boundary conditions nor the right resolution, this will not give a correct representation anyway. So, we can assume d(diss)/dn = 0 for every wall and k_dirichl = 0, which means a weak imposition. In any case, we expect any turbulence to propagate from the far field, rather than generated from the boundary condition. For outlets I used  (d/dn = 0) for both as we need to let them go with the flow

c) For walls, I think we do not need to specify any of the diffusive conditions, as the combination of advective and dirichlet will do the job, and vice versa as for outlets we define an appropriate combination of dirichlet and diffusive

d) I introduced a two phase boundary condition for unstready flow and waves. I am worried that eta(x,y) is calculated too many times. @cekees  is there a clever way to go around that?

### DIFF
--- a/proteus/BC.py
+++ b/proteus/BC.py
@@ -93,10 +93,7 @@ class BoundaryConditions:
         This function should be used only on BC flag 0 (index 0 in the list
         domain.bc)
         """
-        self.vof_advective = constantBC(0.)
-        self.u_diffusive = constantBC(0.)
-        self.v_diffusive = constantBC(0.)
-        self.w_diffusive = constantBC(0.)
+        self.reset()
 
     def setTank(self):
         b_or = self._b_or[self._b_i].tolist()
@@ -118,19 +115,8 @@ class BoundaryConditions:
         self.u_dirichlet = constantBC(0.)
         self.v_dirichlet = constantBC(0.)
         self.w_dirichlet = constantBC(0.)
-        self.k_dirichlet = constantBC(0.)
-        self.dissipation_dirichlet = constantBC(0.)
         self.p_advective = constantBC(0.)
-        self.u_advective = constantBC(0.)
-        self.v_advective = constantBC(0.)
-        self.w_advective = constantBC(0.)
-        self.vof_advective = constantBC(0.)
-        self.k_advective = constantBC(0.)
-        self.dissipation_advective = constantBC(0.)
-        self.u_diffusive = constantBC(0.)
-        self.v_diffusive = constantBC(0.)
-        self.w_diffusive = constantBC(0.)
-        self.k_diffusive = constantBC(0.)
+        self.k_dirichlet = constantBC(0.)
         self.dissipation_diffusive = constantBC(0.)
 
     def setFreeSlip(self):
@@ -142,13 +128,7 @@ class BoundaryConditions:
         self.u_advective = constantBC(0.)
         self.v_advective = constantBC(0.)
         self.w_advective = constantBC(0.)
-        self.k_advective = constantBC(0.)
-        self.dissipation_advective = constantBC(0.)
-        self.vof_advective = constantBC(0.)
-        self.u_diffusive = constantBC(0.)
-        self.v_diffusive = constantBC(0.)
-        self.w_diffusive = constantBC(0.)
-        self.k_diffusive = constantBC(0.)
+        self.k_dirichlet = constantBC(0.)
         self.dissipation_diffusive = constantBC(0.)
 
     def setClosed(self):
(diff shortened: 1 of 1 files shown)


tridelat:
@adimako 
a) For the BC on flag 0, proteus will throw warnings when running in parallel if the diffusive u, v, and w and advective vof are not set to zero on flag 0. (it will also automatically set them to zero anyway)

b) and c) noted and I will try

d) I think that might be the same problem as for rigid body calculations, where each processor recalculates the body motion.

cekees:
On d), I'm not sure I see a way around the multiple eta evaluates right now.  Maybe something like this

```
from math import sin

class A:
    def __init__(self):
        self.ETA = 0.5
        self.ETA_t = 0.0
    def eta(self,t):
        return sin(t)
    def update_eta(self,t):
        if t > self.ETA_t:
            self.ETA = self.eta(t)
            self.ETA_t = t
    def getFunc(self):
        def f(x,t):
            self.update_eta(t);
            return t > self.ETA
        return f
    def getFunc2(self):
        def f2(x,t):
            self.update_eta(t);
            return t < self.ETA
        return f2

a = A()
f = a.getFunc()
f2 = a.getFunc()
print f(0,0), f(1,0), f(2,0)
print f2(0,0), f2(1,0), f2(2,0)
a.ETA = 1.0
print f(0,3.14/2), f(1,3.14/2), f(2,3.14/2)
print f2(0,3.14/2), f2(1,3.14/2), f2(2,3.14/2)
```

tridelat:
d) We could attach the values of `t`, and the resulting `eta(x, t)` for each `x` in a dictionary (or a numpy structured array if it is faster) directly in the `BoundaryConditions` class and use `if` statements in all the functions that use `eta` such as:

```python
def ux_dirichlet(x, t):
    if t != self._t:  # only once per time step
        self._t = t
        self._eta_dict.clear()  # in case the mesh moved, better to clear the dict
    if x not in self._eta_dict:  # only once per point
        self._eta_dict[x] += eta(x, t)
    if x[vert_axis] < self._eta_dict[x]:
        return 'something'
    else:
        return 'something else'
```

So eta is called only once per time step for a given point and can be fed directly to all the remaining boundary conditions using the same `x`. 
If `eta` was only dependent on a `x[2]` value, then it can be even further reduced by calling it only once per time step (no need to store the whole `_eta_dict` anymore)

cekees:
I think that  would work to eliminate the extra eta function calls.  I'm not sure it's with the trouble at this point since the extra  logic is expensive.  I would vote for continuing with the current approach unless this shows up in profiling as a major computational expense.

adimako:
@cekees, @tridelat, I guess the question is how slow are the look-up algorithms against the cos and sin calculations. Also, @tridelat, @ceekees, I think I can reduce the time, by getting rid of the cosh calcs (they are not time dependent). So at the beginning of the simulation, I would like to a create a cosh(k(d+z)) mesh field in the relaxation zone, so I can pull the values from there and not having to calculate then every timestep. That could be done with @tridelat solution

cekees:
First we need to establish that the calls to eta() are a significant computational expense. I prefer the simplicity of the extra calls to eta() over any kind of caching unless they actually show up high in the profile on problems we care about.  If we were down at the C level, the re-computing eta would still probably be faster than caching it.  Since a lot  of this is  in Python, the actually computation times may not be coming from what you think they are (i.e they may be coming from various kinds of Python overhead, not memory access and floating point calculations).

cekees:
Just to be clear, @tridelat this is a PR into your branch so you would decide when to merge it if you want the changes from @adimako . I was just trying help with the eta() calls, though I'm happy to just leave it as is for now.

proteus/BC.py
```diff
@@ -266,12 +239,64 @@ def inlet_p_advective(x, t, u=U):
...
+        (!) This condition is best used for boundaries and gravity aligned with
+            one of the main axes.
+        """
+        self.reset()
+        U = np.array(U)
+        
+        def get_inlet_ux_dirichlet(ux):
+            def ux_dirichlet(x, t):
+                if x[vert_axis] < eta(x,t):
+                    return ux(x,t)
```
tridelat: Shouldn't it be returning just `ux`?

proteus/BC.py
```diff
@@ -93,10 +93,7 @@ def setParallelFlag0(self):
         This function should be used only on BC flag 0 (index 0 in the list
         domain.bc)
         """
-        self.vof_advective = constantBC(0.)
-        self.u_diffusive = constantBC(0.)
-        self.v_diffusive = constantBC(0.)
-        self.w_diffusive = constantBC(0.)
+        self.reset()
```
tridelat: We need to set these vof advective and u diffusive to 0 for parallel runs, otherwise the log gets spammed with errors and sets these values to 0 by itself

tridelat:
Merged but will revert changes to `setParallelFlag0` on next commit.
@adimako: should `ux(x,t)` also be changed to `ux`? 
https://github.com/erdc-cm/proteus/pull/325/files#diff-b8ca655491c60eb061b1b24933516dacR267

proteus/BC.py
adimako: Yep, you are right

proteus/BC.py
adimako: ok

adimako:
Both suggestions are correct. I can do it today

PR #326, opened 2016-01-13
WaveTools.py: x passed and vel returned as arrays

tridelat:
This is just a slight modification in WaveTools.py for x to be passed as an array (as provided by Proteus for boundary conditions) when using functions and velocities to be returned as arrays.
All tests in test_wavetools.py have passed.


adimako:
@tridelat since tests are successful, I think I will go on and merge it.

PR #331, opened 2016-01-21
added back support for offset rectangular domains

cekees:
specify x0 in the p-file to set left, front, lower corner


cekees:
@alistairbntl , this should allow you to solve over [-1,1]^2 by setting x0=(-1.,-1.) and L=(2.,2.) instead of defining a domain object. You should also be able to do
```
domain = Domain.RectangularDomain(x=(-1.,-1.), L=(2.,2.), name="box)
```

PR #336, opened 2016-02-16
WIP - Initial commit for iterative Taylor Hood implementation

alistairbntl:
- Added a description of the issue in the function which needs to be fixed.


cekees:
TODO:
- [ ] update  SimpleNavierStokes3D like 2D
- [ ] extend to parallel case (fix  ghost nodes  and subdomain2global)
- [ ] clean up logic  in MultilevelTransport.py so we don't have so many cases

proteus/Archiver.py
```diff
@@ -1345,7 +1357,7 @@ def writeMeshXdmf_C0Q2Lagrange(self,ar,name,mesh,spaceDim,dofMap,t=0.0,init=Fals
                         if init or meshChanged:
                             if ar.has_h5py:
                                 ar.create_dataset_sync('elements'+spaceSuffix+`tCount`,
-                                                       offsets = mesh.globalMesh.elementOffsets_subdomain_owned*nsubelements,
+                                                       offsets = [x*nsubelements for x in mesh.globalMesh.elementOffsets_subdomain_owned],
```
cekees: BTW,  you can write this more concisely as  `mesh.globalMesh.elementOffsets_subdomain_owned*nsubelements`

proteus/FemTools.py
```diff
@@ -529,18 +521,26 @@ def __init__(self,nd=3, order=2):
         elif nd == 2:
             for j in range(order+1):
                 for i in range(order+1):
-                    basis.append(lambda xi,i=1,j=j:self.fun[i](xi[0])*self.fun[j](xi[1]))
+#                    basis.append(
+#                        lambda xi,i=1,j=j: self.fun[i](xi[0])*self.fun[j](xi[1]))
+#                    basisGradients.append(
+#                        lambda xi,i=i,j=j: numpy.array([self.dfun[i](xi[0])*self. fun[j](xi[1]),
+#                                                        self. fun[i](xi[0])*self.dfun[j](xi[1])]))
+#            funMap=[0,4,1,  7,8,5,   3,6,2]
```
cekees: In the past we've left in a lot of commented out code like  this, but I'm trying to do a lot  less  of this sort of thing. I would only leave  it in if it's explanatory, as in, "here is one way to do it, but for now we're going with the following approach, which seems better because..."

proteus/FemTools.py
```diff
@@ -3998,6 +4083,16 @@ def writeVectorFunctionXdmf(self,ar,uList,components,vectorName,tCount=0,init=Tr
...
 
 
+# Lagrange Factory On Cube
+def LagrangeCubeFactory(OrderIn):
+    class LagrangeCubeOrderN(C0_AffineLagrangeOnCubeWithNodalBasis):
+        def __init__(self,mesh,nd):
+            C0_AffineLagrangeOnCubeWithNodalBasis.__init__(self,mesh,nd,order=OrderIn)
+    return LagrangeCubeOrderN
+
+# Q1 = LagrangeCubeFactory(1)
```
cekees: We could just to `Q1 = C0_AffineLagrangeOnCubeWithNodalBasis` to make example code cleaner.

proteus/LinearAlgebraTools.py
```diff
@@ -50,11 +53,19 @@ def scatter_reverse_add(self):
...
     """
     Parallel vector using petsc4py's wrappers for PETSc
-    """
-    def __init__(self,array=None,bs=None,n=None,N=None,nghosts=None,subdomain2global=None,blockVecType="simple"):
-        if array == None:
-            return#cek hack, don't know why init gets called by PETSc.Vec duplicate function
+    WIP -- This function builds the local to global mapping for the PETSc parallel vectors.  At this
+    point it only works when the variables can be interwoven (eg. stablized elements where velocity and
+    pressure come from the same space).  We would like to extend this functionality to include finite
+    element spaces that cannot be interwoven such as Taylor Hood.
```
cekees: I need to remove this comment, since it's not true anymore. We can to interleave and non-interleaved now.

proteus/LinearSolvers.py
```diff
@@ -514,6 +517,7 @@ def prepare(self,b=None):
             self.petsc_L.setValuesLocalCSR(self.csr_rep[0],self.csr_rep[1],self.csr_rep[2],p4pyPETSc.InsertMode.ADD_VALUES)
         self.petsc_L.assemblyBegin()
         self.petsc_L.assemblyEnd()
+        self.petsc_L.save("L")
```
cekees: We should probably comment this out for now or add a `--print-matrices` option that defaults  to `False`

cekees:
@alistairbntl This looks great. I'm going to go ahead and merge. I put in some line comments that are mostly trivial. We can do some small branches for cleaning up and documenting at some point later. Thanks for your hard work  on getting this PR ready to go in.

PR #337, opened 2016-02-17
Stable/copper

adimako:
@cekees just creating this pr to compare files, no need to merge. We (me and @tridelat) have some issues in copper with messages regarding -fPIC option and relocation errors for scipy and grib_api


adimako:
@cekees no changes there so I am going to close it

PR #339, opened 2016-02-20
Raises and possibly addresses a bug in the multilevelprojection class.

alistairbntl:
- Since the interpolation functionals can return positive or
	 negative values it is possible rbcSum can be zero after
	 completion of the loop even if the interpolationfunctional
	 return non-zero numbers.
	- This can cause issues in line 6617 when rbcSum is used
	 to scale the rbc values.
	- I believe the best solution is to make rbcSum the sum
	 of positive numbers and avoiding the scaling in the corner
	 case where all interpolation values are zero.
	- As I'm not familiar with the multilevelprojection algorithm,
	 this solution may not be appropriate.

### DIFF
--- a/proteus/FemTools.py
+++ b/proteus/FemTools.py
@@ -6606,7 +6606,7 @@ class MultilevelProjectionOperators:
                 #end coarse_eN
                 for I in range(coarseSpace.dim):
                     for J in rbcColumnIndeces[I]:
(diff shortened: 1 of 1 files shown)


cekees:
Thanks, I think this should be fine for now. I'll revisit when I get around to finishing #162

PR #353, opened 2016-04-03
Diminishing porous dissipation from the air phase

adimako:


adimako:
@Giovanni-Cozzuto-1989, try to compile this branch on your proteus installation. I did a minor correction on the source terms, I scaled it with the density ratio and so that it is 1 on the water and 0 on the air phase. Since you observed that reducing the height of the porous domain makes things better, I am suspecting that the porous dissipation in the air phase is over-estimated

PR #355, opened 2016-04-04
Fixes a bug in the xmf archiver for post-processed solutions.

alistairbntl:
- Also adds a new test for testing postprocessing output.


proteus/tests/ci/test_zpostprocessing.py
```diff
@@ -0,0 +1,71 @@
...
+    opts.profile=True
+    opts.gatherArchive=True
+    # create conservative velocity field
+    simFlagsList = [{}]
+    simFlagsList[0]['storeQuantities'] = ['u',"q:('velocity',0)"]
+    # solve problem
+    ns = NumericalSolution.NS_base(so,pList,nList,so.sList,opts,simFlagsList)
+    ns.calculateSolution('test1')
+    #
+    test_path = os.path.dirname(os.path.abspath(__file__))
```
cekees: @alistairbntl this is the slightly tricky bit to get the test working when run from an arbitrary directory. The generated h5 will be in the directory where the test  command is run but the expected.h5 will be in the same directory as the test  file.

PR #359, opened 2016-04-07
Adimako/wave tools/cython

adimako:
Giovani digged an error in the equations. When we derive the higher armonics, we should multiply the phases as well. Does not affect any previous results as so far the phase has been 0.


adimako:
It was tested successfully, so I am merging it

PR #371, opened 2016-04-25
copy of coords for Cuboid if no barycenter

tridelat:
`self.barycenter` and `self.coords` were the same object on this class when `barycenter` was not passed as an argument.


cekees:
Thanks!

PR #372, opened 2016-04-27
[WIP] initial commit  of updated gmsh scripts

cekees:
@tridelat Here are the most recent versions that  I've been using.  There are 4 gmsh2tetgen scripts and 2 generate_gmsh_mesh scripts.  I'm leaving them in for now, but we should probably reduce  these to  one of each for  now. 

This works for the air-water-vv repository commit fa54a9562424ef604a3776a8be440220f79cd472

`cd air-water-vv/3d/dtmb/mesh`
`export PATH=${HOME}/proteus/scripts:${PATH}`
`generate_gmsh assembly.geo 10.0`


cekees:
The tests are failing because of something related to a new writemtr function. @tridelat are you working on this branch or should we just close it? I had just pushed updated  generate_gmsh_msh scripts here, so I was thinking that we should merge it in.

PR #373, opened 2016-04-27
Schur qp

alistairbntl:
This PR adds Qp and PCD functionality to the Navier Stokes solver class.


proteus/FemTools.py
```diff
@@ -3095,15 +3095,21 @@ def getInverseValuesGlobalExteriorTrace(self,
...
-    dim -- global dimension of the space
-    mesh -- a Mesh object, which is a partition of the domain
-    finiteElements -- a dictionary of the FiniteElement objects
-    indexed by element number
-    dofMap -- a DOF
+    """ Base class for finite elements defined on the reference element.
+
+    Parameters
+    ----------
+    referenceFiniteElement : class
```
cekees: I believe hear  you can put the actual type names and sphinx will generate links to the class definitions

```
ReferenceFiniteElement
ElementMaps
DOFMap

proteus.MeshTools.Mesh
```

alistairbntl: Cool, so this does work.  It does require some special markup.  For the mesh you need to say something like :class:`proteus.MeshTools.Mesh` for example.  The next commit will have the reference included.

proteus/LinearAlgebraTools.py
```diff
@@ -272,6 +307,126 @@ def mult(self, A, x, y):
...
+            Dummy place holder for PETSc compatibility
+        x : vector
+
+        Returns
+        -------
+        y : vector
+        """
+        self.ksp.solve(x,y)
+
+class PCDInv_shell:
```
cekees: probably want to delete "pressure mass" in the docstring, since this is a generic inverse matrix shell

proteus/LinearSolvers.py
```diff
@@ -581,10 +610,309 @@ def failed(self):
     def info(self):
         self.ksp.view()
 
-class NavierStokes3D:
+class schurOperatorConstructor:
```
cekees: We go  by the convention that derived  types (class names) are usually capitalized while object names are lower case.

proteus/NonlinearSolvers.py
```diff
@@ -435,12 +435,19 @@ def setLinearSolverTolerance(self,r):
         self.norm_r_last = self.norm_r
         self.linearSolver.setResTol(rtol=eta,atol=self.linearSolver.atol_r)
     def solve(self,u,r=None,b=None,par_u=None,par_r=None):
-        """
-        Solve F(u) = b
-
-        b -- right hand side
-        u -- solution
-        r -- F(u) - b
+        """ Solve the non-linear problem F(u) = b.
```
cekees: thanks for converting this to the new standard

proteus/Transport.py
```diff
@@ -2714,6 +2712,8 @@ def calculateElementJacobian(self,skipMassTerms=False):
                     if self.timeIntegration.diffusionIsImplicit[ci]:
                         if self.numericalFlux == None or self.numericalFlux.mixedDiffusion[ci] == False:
                             if self.sd:
+                                import pdb
```
cekees: probably want to delete these from committed code. they will cause travis to fail.

cekees:
This looks great. I need to go through the actual construction of the blocks more closely. You may want to go ahead and add some of the tests to the `nosetests` line in .travis.yml so they get run automatically.

proteus/LinearSolvers.py
```diff
@@ -809,38 +790,48 @@ def getFp(self,output_matrix=False):
...
             _exportMatrix(self.Fp,'Fp')
         #Ff(1:np,1:np) would be Fp, the pressure convection-diffusion matrix
         #
         #now zero all the dummy coefficents
         #
-        self.L.pde.q[('df',0,0)][:] = 0.0
-        self.Fp = self.getAp()
-        self.Fp.scale(self.L.pde.coefficients.nu)
-        self.Fp.__add__(self.Cp)
         return self.Fp
```
alistairbntl: Note __add__ was not actually adding self.Cp to self.Fp.  I've corrected this above, which illuminated a problem with the advection part pressure block.

proteus/LinearSolvers.py
```diff
@@ -809,38 +790,48 @@ def getFp(self,output_matrix=False):
...
                                              F_csr_rep_local[2],
                                              p4pyPETSc.InsertMode.INSERT_VALUES)
+
         self.Fsys_petsc4py.assemblyBegin()
         self.Fsys_petsc4py.assemblyEnd()
-        self.Cp = self.Fsys_petsc4py.getSubMatrix(self.linear_smoother.isp,self.linear_smoother.isp)
+
+        self.Cp = self.Fsys_petsc4py.getSubMatrix(self.linear_smoother.isp,
+                                                  self.linear_smoother.isp)
+
```
alistairbntl: `self.Cp` should be the pressure-advection term that needs to be added to the scaled pressure-lapacian to create the correct Fp operator.  Based on my simulations this does not appear to be creating the proper matrix.

cekees: Since you're just trying to set up the advection term alone, you should be able to print `self.Cp` here (for a small mesh) and check directly if `Cp_ij = - int dot(u w_j ,grad  w_i) dV ` and that it is equal to a the proper block in Fsys.

An alternative way to set up Fp is just to set it up all at once by setting the advection, diffusion, and mass coefficients corresponding to the PCD preconditioner. I'm not sure what will be easier in the long run.

PR #374, opened 2016-04-28
Copper/fix

adimako:


cekees:
This should be fine, but I suspect you want to make these changes on master.

PR #375, opened 2016-04-28
Stable/copper

adimako:


cekees:
Hi Aggelos, stable/copper is a "tracking branch", meaning we generally only merge origin/master into stable/copper.   What are you needing to do?

PR #376, opened 2016-04-28
Adimako/abs/fix

adimako:


adimako:
@cekees, @tridelat I have tested this with one of Elena's old case and I get the about the same reflection coefficients for a nonlinear wave, although the case I ran is not exactly as Elena's. I think it is a good fix for now, we should go ahead and merge this to the master node. I did the same for the stable/copper branch (along with some updates in BoundaryConditions.py and SpatialTools.py) but I get errors during checks (https://github.com/erdc-cm/proteus/pull/374). I think the error is related to the PETSC, not the commits per se, but note that this version has compiled & run normally in my account in copper.

cekees:
Just to be sure I'm following: 1) the sdf for the sponge layer center line is given by phi_s, and it should be oriented  so that phi_s > 0 is where you want the damping/bc enforcement and phi_s < 0 is where you want to transition to internal physics. 2) the transition width is 2*eps_s; 3) now the transition over 2*eps_s will be the 0.5*(1-cos(pi x)) function.
```
double x = fmax(0.0, fmin( 1.0, 0.5+phi_s/(2.0*eps_s)));//0 at phi_s = -eps, 1 at phi_s=eps
H_s = 0.5*(1 - cos(M_PI*x));
```

adimako:
1) for phi_s < 0   then x<0.5 and phi_s >0 then x>0.5, as phi_s measures from the centre of the rel zone. The function then is going from 0 to 1, passing from phi_s = 0, and if the function is antisymmertic then it should be 0.5 at phi_s =0 (see graph below)
capture

adimako:
2) Correct as eps is half of the zone lenght by definition
3) Correct.

cekees:
OK, thanks. Looks good to me.

cekees:
the pr test is failing due to time outs so I'm just going to merge as is.

PR #380, opened 2016-04-28
adds configuration changes for compiling on OS X 10.11.4 with clang 7.3.0

mfarthin:
- requires hashdist/hashstack pull request #946
- Defines `PETSC_SKIP_COMPLEX` to avoid template conflicts from `<complex>` inclusion in `petscmath.h`

- adds a compiler flag to allow looser pointer conversions


mfarthin:
If users are having trouble with accessing `pypi` through `openssl` -- for example when installing `vcversioner` they can export their system certificates to `pem` format and put them in `/etc/ssl/cert.pem`

PR #387, opened 2016-05-05
Added function in RandomWaveClass for writing timeseries

adimako:
Adding a function for writing timeseries to Random Waves


adimako:
@cekees, @tridelat This new function will automate the generation of a random time series and writing in to a file, so it can be picked up by the time series class, for speeding up generation. It also includes extending the timeseries, so it can be consistently used in the relaxation zone. I will wait to test it with the numerical tank before I merge this

adimako:
Closing and deleting branch, as everything is moved to adimako/wavetools_update

PR #391, opened 2016-05-11
Copper/update

adimako:
@cekees, this is the version I used for running the cases. As said only some tweaks in WaveTools


adimako:
Also do not merge, it is not there yet

adimako:
@cekees Failed because of memory access errors, but compiles ok in copper

adimako:
Synchonised flcbdWrappers with the master branch, same profile results

cekees:
Is this ready to merge too? Or  is it  out of date since the other branch went into stable/copper?

adimako:
@cekees, this is stale, should be removed. I have implemented these already in WaveTools

PR #393, opened 2016-05-12
added prototype of gmsh support

cekees:
This adds the ability to do

```
domain = GMSH_3D_Domain("assembly", he=10.0)
```
where `assembly.geo` is  the main gmsh file describing a  domain and `he` is the max  element diameter. The only example I have now is in 3d/dtmb on this commit, but it appears to be working:

https://github.com/erdc-cm/air-water-vv/commit/70954d5d4f5cbcc747e5469c0dcc76e3f484407d

I would prefer that the `he` get set through @tridelat 's new MeshOptions class, but other than that I'm not sure how fancy to make it. I think we should probably push most of the work  of setting up the geometry and specifying the mesh generation onto  gmsh itself (and the writer of the actual .geo file) so the focus for  people  using the gmsh pre-processing will be on learning gmsh, not a proteus interface  to it.

/cc  @tridelat  @alistairbntl @adimako

### DIFF
--- a/proteus/Domain.py
+++ b/proteus/Domain.py
@@ -301,6 +301,27 @@ shipout();
     def writeXdmf(self,ar):
         raise UserWarning("Xdmf output not implemented")
 
+class GMSH_3D_Domain(D_base):
+    """
+    A domain described by a gmsh .geo file
+
+    Parameters
+    ----------
+    geofile : str
+              The base filename of the main .geo file
+    he : float
+         The maximum element diameter (this should go way)
+    name : str
+           A name for the domain
+    units : str
+            The units of length
+    """
+    def __init__(self, geofile, he, name="GMSH_Domain", units="m"):
+        D_base.__init__(self, 2, name, units)
+        self.geofile=geofile+".geo"
+        self.polyfile=geofile
+        self.he = he
+
 class PlanarStraightLineGraphDomain(D_base):
     """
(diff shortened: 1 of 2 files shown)


tridelat:
Thank you for this, should we also add an option for PLC or other domains to be meshed through gmsh or should we generate a geo file from those domains first  with a function like `self.writeGeo()` and then create a new GMSH_3D_Domain instance?
I quite like defining my geometry through proteus as it makes it easier to manipulate with context options and to deal with rigid bodies.

cekees:
That would be good. Maybe we could just have a member  like  self.to_gmsh() that would return the GMSH_3D_Domain instance so it could be done in one call.  `domain = PLC(....).to_gmsh()`. I guess  another option if we enable other 3D domains to be capable  of writing a .geo  representation of themselves is just to add another mesh options like `use_gmsh`.

cekees:
@tridelat are you working on this branch or are you working on gmsh support  on a different branch? Not a big deal either way, but if you don't have new work done in this direction I may clean this up a little and merge so I can start using gmsh meshes.

tridelat:
@cekees I am working on making it possible to convert 2D cases from gmsh to triangle (with flags). You can merge this branch if it is more convenient and we can add the support for 2D later.

cekees:
I'm going to go ahead and merge this but will leave the branch in case you have made additional commits to it.

PR #397, opened 2016-05-19
BoundaryConditions as a cython module

tridelat:
@adimako @geomak89 this branch is for optimizing the BoundaryConditions module with cython.


cekees:
Cool. I think it will make a lot  of difference to have `twp_flowVelocity` defined compiled as C since I think the WaveTools  functions it calls already are. Then declaring the types of the loop variables in the relaxation zone loops over the elements should make that whole computation run fast. You may want to look at the Cython documentation on how you can have it generate an annotated version to see what has been optimized.

proteus/mprans/SpatialTools.py
```diff
@@ -1390,7 +1385,7 @@ def assembleAuxiliaryVariables(domain):
     -----
     Should be called after assembleGeometry
     """
-    domain.auxiliaryVariables = []
+    domain.auxiliaryVariables = {'twp': [], 'vof': []}
```
nerisaurus: I think I like where this is going [1], but I'm not entirely sure I understand how these auxiliaryVariables are being set up.

We start with two collections of auxiliaryVariables, 'twp' and 'vof', which I assume are for the twp_navier_stokes and vof models respectively.  Should we have ones for the other models used as well?  On a brief glance, I at least have a model which brings in auxiliaryVariables from `redist_n.py`, and I wouldn't be surprised if other models have the same problem.

(Maybe it would be nice to have a `getAuxiliaryVariables(*key name*)` method for the domain, which can return `[ ]` if the dictionary does not contain an element [2] )

But anyway, for each shape we keep setting `aux = domain.auxiliaryVariables` (which starts as the empty-ish dictionary we initialized above), and then adding `aux['twp']` values and... then we have `aux[-1]` which we seem to be making some assumptions about (line 1400, if I read these numbers right, sets `body = aux[-1]` and then immediately starts making broad assumptions about body's components).  I'm just not sure what happened to 'vof' (or if it's just there for later expansion), and how the -1 is meant to work after resetting domain.auxiliaryVariables above.

(After looking at it more, I've figured out that the -1 part is just referring back to what is now the 'twp' - I'm guessing it just hasn't been changed yet [but can be without breaking something down the line]?  That eases my confusion a bit, although I still have some other questions I've kept in this)

[1] A dictionary is just what I was looking to implement for gauges.
[2] This would allow each `*_n.py` file to say `auxiliaryVariables = domain.getAuxiliaryVariables(" * ")` and not worry about it breaking if there's nothing for it.  Then the user can throw in gauges (or other `*_n.py` specific auxiliary variables) and toggle them on and off from the main .py file without worrying about changing the specific `*_n.py` files each time - while also being able to affect each and every possible `*_n.py` file they might have, not just `twp` and `vof` (and we wouldn't have to add on new empty dictionary elements for every possible `*_n.py` file.

tridelat:
Yes, thanks for spotting this, there is an error that was introduced in the latest commit: it should be `aux['twp'][-1]`. The `-1` here is to access the latest auxiliary variable added in the list (a class instance of `RigidBody` in this case). But maybe it looks a bit clumsy anyway, we could do something clearer like this for the same effect:
```
body = RigidBody(shape)
aux['twp'] += [body]
body.i_start = start_flag+1
...
```

Because the auxiliary variables are already stored in a dictionary after calling `assembleAuxiliaryVariables`, each `*_n.py` files would just need something like: 
```
auxiliaryVariables = domain.auxiliaryVariables['twp']  # this should return a list
```
New keys such as `redist` or `ls` can be added to the `domain.auxiliaryVariables` dictionary that is initialised in the function `assembleAuxiliaryVariables(domain)`. I added vof there because I was using it for column gauges.

proteus/mprans/SpatialTools.py
tridelat: Yes, thanks for spotting this, there is an error that was introduced in the latest commit: it should be `aux['twp'][-1]`. The `-1` here is to access the latest auxiliary variable added in the list (a class instance of `RigidBody` in this case). But maybe it looks a bit clumsy anyway, we could do something clearer like this for the same effect:
```
body = RigidBody(shape)
aux['twp'] += [body]
body.i_start = start_flag+1
...
```

Because the auxiliary variables are already stored in a dictionary after calling `assembleAuxiliaryVariables`, each `*_n.py` files would just need something like: 
```
auxiliaryVariables = domain.auxiliaryVariables['twp']  # this should return a list
```
New keys such as `redist` or `ls` can be added to the `domain.auxiliaryVariables` dictionary that is initialised in the function `assembleAuxiliaryVariables(domain)`. I added vof there because I was using it for column gauges.

nerisaurus: I put out a question on slack to this aim, but I'll ask it here too for redundancy: can we be sure that we know all the keys that we might need?

I could put in `redist` and `ls` in the `assembleAuxiliaryVariables(domain)` dictionary initialization.  To be cautious, I could add in `dissipation`, `ls_consrv`, and `kappa`.  If any new models get tacked on, though, I'll have to return to this again (and if I'm a user who isn't closely connected to Proteus, I'm going to have to maintain a local branch with my changes each time).

If the above keys are pretty much final, then I'm fine with that.  If not, I'd rather use `domain.auxiliaryVariables.get(*key name*, [ ])` and not initialize anything unnecessarily in the dictionary (but have methods dynamically add keys for it to initialize, potentially into a list).

tridelat:
@adimako @Giovanni-Cozzuto-1989 @nerisaurus 
This PR is ready to be merged into master. Let me know if there are any objections.
(further modifications/additions will be done in other branches)

This will affect the test cases set-up in the following ways:
- boundary conditions are now only accessed as `my_shape.BC['my_tag']`, and the functionality to access them via`my_shape.BC.my_tag` has been removed (because it would not work with tags containing spaces or special characters such as `+-=@`, which could cause confusion..)
- in the `*_p.py` files, the lines accessing the BC must be changed to `domain.bc[flag].p_dirichlet.init_cython()` (instead of `domain.bc[flag].p_dirichlet`)
- to set a custom function to a boundary condition, it must be done with `my_shape.BC['my_tag'].u_dirichlet.uOfXT = my_function` (instead of `my_shape.BC['my_tag'].u_dirichlet = my_function`)
- several boundary tags have been changed for clarity (especially when dealing with 3D where, for example, a cuboid front/back/left/right does not clearly state what axis they refer to). For example, the right boundary of the Tank2D classed is now accessed via `my_tank.BC['x+']` instead of `my_tank.BC['right']`.
- some function arguments of SpatialTools may have changed but should be easy to spot when trying to run a test case

Also, after recompiling proteus, `BoundaryConditions.pyc` and `mprans/BoundaryConditions.pyc` **must be removed manually** as they are not cleaned up automatically (otherwise they will be used instead of the `.so` and `.c` files coming from cython).

cekees:
+1 for merging this. We definitely want to get  it in soon so  @nerisaurus has a chance to finish updating his work  to include the slight API changes. Just out of curiousity, did you and @adimako ever get any precise numbers on the speed increases this produces for specific test problems?

tridelat:
Okay, I will merge. We have not compared precisely to previous test cases yet, but will keep you updated on the speed increase. I will run a test case with waves + generation zone (+ floating body) previously run on copper to get precise numbers, as testing on hydra is not always reliable (speed can be affected by how much processing power is used by others).

adimako:
@adimako I only have indications for the speed up by using c definitions in WaveTools. I would recommend performing a short test (~5 mins) with profiling with the linear waves numerical tank.

adimako:
@tridelat, @cekees, @Giovanni-Cozzuto-1989 @nerisaurus @alistairbntl. As a general comment, these changes justify tagging of a version, since we are already have test cases running with the old interface. @tridelat, can you tag the previous version from this commit. I observed that random waves are unjustifiably slow in the cluster, so I will try also this out to see if it makes any difference

adimako:
@tridelat @cekees @Giovanni-Cozzuto-1989 @nerisaurus @alistairbntl I think this has removed the overheads coming from the BoundaryCalls etc. Still we have to compare like for like but this is my impression from a test I run. If so, we may apply this to WaveTools as well and then having to optimise only eta_mode and vel_mode as these cause the real computational delays, everything else is overhead between calls.

Issue #401, opened 2016-05-26
Switch log to logEvent


alistairbntl:
There are currently two commands for outputting to the Proteus log file - `log` and `logEvent`.  

As @cekees pointed out, the `log` function is a poor choice given there is a mathematical function with the same name.  I believe that all `log` functions for outputting to the simulation log file should be switched to `logEvent`.  

@cekees - Is this correct or am I missing something subtle here?

PR #402, opened 2016-05-28
Replaces log instances with logEvent.

alistairbntl:
log and logEvent are used interchangeably throughout the library.  Since `log` is ambiguous, this PR seeks to standardize all logEvents in the library.  Note, this is meant to address #401.


cekees:
Looks great. Thanks for doing that.

PR #404, opened 2016-06-01
fix case when point not  in nearest node star elements

cekees:
For now just continue checking node stars of all nodes  on the original patch boundary.

Also added a fix for a failing assertion in protues.mprans.MCorr.py arising due to slight differences in numpy and python sum.

todo
-----
- [ ] add test that triggers this case
- [ ] check if this also fixes apparent line gauge issues
- [ ] plan long term fix for point location and path intersection calculations


cekees:
@Giovanni-Cozzuto-1989 @nerisauris @tridelat and @adimako , this will effect the gauge output in air-water-vv by removing some spurious spikes at random gauge locations. It's a pure python fix that requires no new packages in the stack.

cekees:
merging this and raising unfished tasks in #426

PR #410, opened 2016-06-07
Updates to testing framework.

alistairbntl:
This PR introduces a few new features to the testing suite.
	1) It adds mesh_tests to the travis tests and corrects a	regression.
	2) It introduces a testing class for handling set-up and tear downs of test problems.
	3) It introduces nosetests attributes which allow marking tests as fast or slow to allow different     ranges of testing depending on the computing resources available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,8 @@ install:
 - ls $SSL_CERT_DIR
 
 script:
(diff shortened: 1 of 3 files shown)


alistairbntl:
Please don't merge until I've added a make test.

PR #411, opened 2016-06-07
Updates to testing framework.

alistairbntl:
- This PR introduces a few new features to the testing suite.
	1) It adds mesh_tests to the travis tests and corrects a regression
	2) It introduces a testing class for handling set-up and tear downs of test problems.
	3) It introduces nosetests attributes which allow marking tests as fast or slow to allow different   ranges of testing depending on the computing resources avaliable.
	4) Adds a `make test` option to the Makefile.


alistairbntl:
@cekees - many of the tests fail when I run the `nosetests` command on my local machine while they pass when the same command is run on the travis-ci server.  

Does this have something to do with loading tests etc?   I'd like to fix this before merging this PR.

cekees:
This  can get sort of tricky right now. I've seen cases where the module imports interact in funny ways, so you have to do hard module reloads rather than just imports in some of the setup code. nosetests also has some options for isolating tests, but  even so I've had trouble with mysterious interactions among tests.  You can enable debugging with nose (forget the option right now), which may help.

.travis.yml
```diff
@@ -32,7 +32,8 @@ install:
 - ls $SSL_CERT_DIR
 
 script:
-- nosetests -vv --traverse-namespace proteus/tests proteus/tests/ci  proteus/tests/poisson_2d
+- nosetests -vv --traverse-namespace proteus/tests proteus/tests/ci  proteus/tests/poisson_2d  proteus/tests/mesh_tests -a '!slow'
```
cekees: How does the `-a '!slow` option work?

cekees: Okay, I see you just add the @attribute('fast',...) or @attribute('slow',...) to each test to classify into fast and slow tests. Is that the idea? The makes organizing the tests into tests for travis  and other CI frameworks (e.g. buildbot) much less invasive than using a directory hierarchy.

cekees:
@alistairbntl just checking in on this. Are you thinking we should redo  this with pytest, or is it compatible with phasing out nosetests? Would be nice to start using the test attributes.

alistairbntl:
@cekees I'm trying to get the pytests stuff off the ground so this PR should wait till its working.  I'm having trouble importing modules like petsc4py when I use the pytests.  Once this is resolved hopefully we'll be in good shape.

proteus/BoundaryConditions.pyx
```diff
@@ -47,7 +47,7 @@ class BC_Base(object):
             cls.ct = context
 
 
-class BoundaryCondition():
+class BoundaryCondition(BC_Base):
```
cekees: @tridelat I accidentally committed this change while trying to debug the testing. It looked strange to me that you had this BoundaryConditions class but it wasn't derived from BC_Base. Can you check  this to make sure I switch it to what you intended?

tridelat: @cekees `BoundaryCondition` is actually not a subclass of `BC_Base`. Maybe the names are a bit confusing, I made this class for cython and for initializing variables that are originally tied to an object but do not change in the `uOfXT(x,t)` function during the simulation. 
`BC_Base` is a container that could have class instances of `BoundaryCondition` as variables that would then be manipulated through functions (like for `BC_RANS` in mprans.BoundaryConditions where `u_advective`, `vof_dirichlet`, etc are `BoundaryCondition` instances).

cekees: Thanks for looking at it. Based on our phone call I'll wait for you to double check that I haven't made  any other mistakes with BoundaryConditions/SpatialTools and put them back as they should be. You can go ahead and merge once any of your modifications pass the  tests.

PR #413, opened 2016-06-13
Adimako/docs update

adimako:


proteus/WaveTools.pyx
```diff
@@ -19,17 +20,16 @@ import sys as sys
 
 
 def loadExistingFunction(funcName, validFunctions):
-    """ Checks if a function name is present in a list of known functions,
-    returns system exit if not present
+    """ Checks if a function name is present in a list of known functions, returns system exit if not present
```
alistairbntl: @adimako - Thanks for putting together a PR to improve documentation!  Here are a few comments I have.

So I'm not sure how to handle this case.  

The standard calls for line length to be kept below 75 characters so that the comment can be read in the terminal.  Moreover, the standard states the top line comment of a function should be a single line.  If a longer description is required, it can be included in an extended comment following the top line comment.  A good example of this is shown: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html.

I'm not saying things need to be done this way, but I do find readability of the library really suffers in places because line lengths are allowed to grow so long so I'd personally like to see longer lines avoided where possible.

proteus/WaveTools.pyx
```diff
@@ -43,26 +43,56 @@ def loadExistingFunction(funcName, validFunctions):
         if func.__name__ == funcName:
             func_ret = func
     if funcName not in funcNames:
-        logEvent("WaveTools.py: Wrong function type (%s) given: Valid wavetypes are %s" %(funcName,funcNames), level=0)
+        logEvent("ERROR! Wavetools.py: Wrong function type (%s) given: Valid wavetypes are %s" %(funcName,funcNames), level=0)
```
alistairbntl: Is this the approach you use for exception handling?  If so I like it and will use the same format in the future.

alistairbntl:
So I was going through this PR and tried to render the documentation to see how it looked.  I'm not sure Sphinx is creating documentation at all for .pyx files.  I'm going to create and issue and try to find a solution.

adimako:
@alistairbntl. I have added messages specific to WaveTools to prevent stupid mistakes, e.g. aligning the wave direction with the gravity direction. Originally, I did not use the ERROR! in the beginning, but I found it difficult to locate the error message in the log file. I believe this can be caught more easily in the logfile. Any other suggestions welcome @cekees @tridelat

adimako:
@alistairbntl Thanks for the comment, I also picked up some suggestions from the link that could make documentation more readable. I will modify and push again at some point

proteus/WaveTools.pyx
```diff
@@ -420,20 +691,48 @@ class MonochromaticWaves:
...
+    param : waveDir
+            Description: Wave direction vector
+            Type: numpy array
+    param : g
+            Description: Gravitational acceleration vector 
+            Type: numpy array
+    param : N
+            Description: Number of frequency components
+            Type: integer
+    param : bandfactor
```
nerisaurus: The comments below in __init__ disagree with the descriptions of the parameters up here.  I figure that's probably just comments which have gotten rotten over many updates (these descriptions make a lot more sense), but I guess you should check just in case and remove unnecessary comments so they don't stick around.  I don't know how common these extra comments are - I just needed to check out this function in particular for a problem.

adimako:
@nerisaurus Yes, you are right, it is not complete, I will change these comments to comply with the directions from the docstring template

adimako:
@nerisaurus @alistairbntl I have changed the functions to meet alistairs suggestion (75 cols max) and to be consistent with numpy docstring. Haven't applied to the classes yet

cekees:
I'm looking at the cython documentation issue. I think it will work, just needs a little tweaking to the sphinx config and possibly the .pyx.

cekees:
OK, I'm not going to get to the cython documentation issue this week or next.  I was planning to just work  through these posts where people have gotten a reasonable level  of support for cythonized modules: http://stackoverflow.com/questions/10057476/how-do-i-use-sphinx-with-cython

proteus/WaveTools.pyx
```diff
@@ -1,68 +1,139 @@
...
+           'vel_mode',
+           'sigma',
+           'JONSWAP',
+           'PM_mod',
+           'cos2s',
+           'mitsuyasu',
+           'dispersion',
+           'tophat',
+           'costap',
+           'decompose_tseries']
```
cekees: For now, the only way I can get sphinx autodoc to autogenerate module's class and function docs is to add the __all__ module data member. 
/cc @adimako @alistairbntl @tridelat @nerisaurus

cekees:
@adimako I think this can be merged now.  You might want to enable the inheritance diagram since some of the waves classes are related, but that could wait until a later refactorization of the class hierarchy for waves.

adimako:
@cekees I will merge it, but I will still need to write the documentation for classes. So I will leave the branch open

adimako:
@cekees @alistairbntl , I still cannot see the documentation in proteustoolkit site. As for the classes, I think we could refactor some of the module function is classes by theme, e.g. numerical tools, wave spectra, spectral analysis etc.

PR #416, opened 2016-06-14
Fixed bug in RandomWavesFast

adimako:


adimako:
This has fixed a bug that caused more filtering that it should for the random wave series. Tests were updated too. As soon as they pass, I will merge to the master

PR #421, opened 2016-06-17
Lin alg tools

alistairbntl:
Adds three helper functions to LinearAlgebraTools with tests.


alistairbntl:
@cekees - I think this PR is pretty minor and should be easy review for merge.  I'm using these tools in a lot of other PRs so it would be helpful to have merged into master.

cekees:
Looks good.

PR #422, opened 2016-06-20
Laplace matrix

alistairbntl:
This is the first commit geared towards building `TransportCoefficients` classes for use in operator construction.  The idea is that Proteus should be able to build a Laplace matrix for the given mesh and FE without modifying and manipulating the existing `TransportCoefficients` class.


proteus/NumericalFlux.py
```diff
@@ -4,6 +4,7 @@
 .. inheritance-diagram:: proteus.NumericalFlux
    :parts: 1
 """
+import pdb
```
cekees: remove the pdb calls

proteus/Transport.py
```diff
@@ -5961,6 +5990,116 @@ def getMassResidual(self,u,r):
...
+            LaplaceJacobian[ci] = {}
+            for cj in range(self.nc):
+                if cj in self.LaplaceOperatorCoeff.stencil[ci]:
+                    LaplaceJacobian[ci][cj] = numpy.zeros(
+                        (self.mesh.nElements_global,
+                         self.nDOF_test_element[ci],
+                         self.nDOF_trial_element[cj]),
+                        'd')
+
+        for ci,ckDict in self.LaplaceOperatorCoeff.diffusion.iteritems():
```
cekees: just for future discussion...this is going to have a constant null space since assembling only the element integrals is equivalent to pure homogenous Neumann conditions. It won't be that hard to grab dirichlet BC object for the pressure and set Dirichlet conditions along the proper boundary segment. I believe we're already expecting the constant null space when the Laplace operator appears in the PC, but I just wanted to flag this.

proteus/TransportCoefficients.py
```diff
@@ -150,7 +150,7 @@ def __init__(self,
                 self.elementIntegralKeys.append(('a',ci,ck))
                 self.elementBoundaryIntegralKeys.append(('a',ci,ck))
                 if not self.potential.has_key(ck):
-                    warn("""a[ci=%d][ck=%d] is non-zero but phi[ck=%d] is undefined. Setting
+                    print("""a[ci=%d][ck=%d] is non-zero but phi[ck=%d] is undefined. Setting
```
cekees: switch this to logEvent

cekees:
Looks good to me. Thanks also for all the tests and documentation updates. Once you remove the print statments and pdb statements you should go ahead and merge.

alistairbntl:
@cekees Changes have been made.  As per the exception raising, I need to read up on it a bit more and will propose some ideas once I have a better handle on how it should be treated.

alistairbntl:
I've refactored and merged this PR into the schur_qp branch.

PR #425, opened 2016-06-27
Discrete Mass Matrix Operator

alistairbntl:


alistairbntl:
@cekees - the most import function in this PR is the `attachMassOperator` added to the `OneLevelTransport` class.  I'm thinking of using this idea as the template for building additional operator constructors so your input on the current set up would be greatly appreciated.

Its a first draft so please don't be afraid to make radical suggestions if you think it can be done better.

Thanks!

cekees:
So now any model can just add a mass matrix for the current mesh and the idea  would be that models could also associate a laplace operator or advection-diffusion operator as well without messing around with the the model's own coefficients or underlying Jacobian assembly. Is that  right? Seems  like a good idea to  me. @mfarthin do you have any use cases for this sort  of thing? The need  came up in the context of  preconditioners where we want to assembly a matrix for  an operator that is  close to the underlying physics but not identical to it.

proteus/Domain.py
```diff
@@ -301,6 +302,45 @@ def writeAsymptote(self, fileprefix):
...
+    nd : int
+        The elements dimension (must be 2 or 3)
+
+    Returns
+    -------
+    reference_element
+    """
+    if nd!=2 and nd!=3:
+        logEvent("ERROR - Reference element must have dimension 2 or 3")
+        sys.exit(1)
```
cekees: You could also raise an exception here. Maybe a ValueError (https://docs.python.org/2/library/exceptions.html). Nathan was  thinking about this some.

proteus/LinearAlgebraTools.py
```diff
@@ -48,9 +132,9 @@ def __init__(self,array,blockSize,n,N,nghosts=None,subdomain2global=None,blockVe
                 self.cparVec=flcbdfWrappers.ParVec(blockSize,n,N,nghosts,subdomain2global,array,1)
         self.nghosts = nghosts
     def scatter_forward_insert(self):
-        self.cparVec.scatter_forward_insert()
+       self.cparVec.scatter_forward_insert()
```
cekees: looks like inconsistent indentation here...3 spaces instead of 4?

proteus/LinearAlgebraTools.py
```diff
@@ -242,15 +325,45 @@ def SparseMatFromDict(nr,nc,aDict):
...
+    -------
+    sparse_matrix : :class:`proteus.superluWrappers.SparseMatrix`
+        superlu sparse matrix in CSR format.
+
+    Note
+    ----
+    For the superluWrapper, both the colind and rowptr should use
+    32-bit integer data types.
+    """
+    if (colind.dtype != 'int32' or rowptr.dtype != 'int32'):
```
cekees: Should be logEvent instead of print. could raise TypeError or ValueError maybe instead  of sys.exit

cekees:
Looks good except for print statements.

proteus/Domain.py
alistairbntl: So I think we should make a decision about this.  I'm following a method I saw @adimako use, but if we'd rather use built in exceptions I'm open to that too.  I think we should be consistent throughout the library with whatever we choose to use.  @nerisaurus?

nerisaurus: It's nice to have logEvents, because that allows them to be very visible in the log file.  It's very handy to have ValueErrors or other errors built off of the various exception classes, though, because we can catch and handle them in other parts of code.  

For example, we could have a statement which tries to unitSimplex(nd) if possible, but has a backup plan if it isn't able to - it can catch a specific ValueError and use that to change its behavior without necessarily crashing (while crashing if other things break).  This also makes it easy to detect if functions are throwing the correct errors in tests, since we can send in wrong results and make sure it breaks the way we want it to.  Note that you can (and should) add in messages to errors, and can even develop simple error subclasses if there's a particular type of error you want to keep separate from the rest.

It might be useful to have a logEvent/ValueError (or, again, any other type of error) combination, if we want to make sure the error gets written down.  Alternatively, having a single function which raises errors and also writes a log event might be nice (I guess you'd pass the specific error type as a function, because there isn't a fixed set of potential errors?), or having some way of catching errors and printing them in the log as if they were logEvents... I don't know how the logging code works, nor if it or python's exceptions have any good qualities for parallelism (the main concern I'd have is that one or the other might be more prone to a race condition which garbles its output if several nodes fail).

alistairbntl:
@cekees - same deal, edits are made and I'll think about exception handling.

alistairbntl:
I've refactored and merged this PR into schur_qp.

PR #429, opened 2016-07-06
updates hashstack version to stable/proteus latest commit

mfarthin:
Looks like most recent commit to `.hashdist_default` was in January


cekees:
For future reference, here's a rough outline of how to update the remote hashdist caches for the travis build.
- docker pull cekees/minv2:latest
- run that image and login to it
- pull proteus and (more importantly) proteus/stack to the new versions
- rebuild that stack and proteus manually
- hit remote push (pushes the source and build caches to a remote server)
- optionally commit and push the cekees/minv2:latest image 
I'll do that next week for this PR. We can do most of that automatically with buildbot once that setup is back online and we have some docker  images running as buildbots. The idea being that the buildbots on ERDC hardware will be doing a fair amount of heavy testing, including fresh builds of the stacks, so they can also push the build artifacts to remote caches. Travis and most builds would then be using   mostly cached source and build artifacts.

cekees:
Still running build.  My power cord got  dislodged and my machine shutdown in the middle of the build. With docker containers you have to manually commit to save state, so an accidental shutdown like that means you have to  start from scratch.

cekees:
@mfarthin and @alistairbntl the hashstack binaries are working for this, but many of the tests are failing.

cekees:
@tridelat some of the tests are failing because  if proteus.mprans.SpatialTools not being able to import proteus.mprans.BoundaryConditions. I think it has to do  with how the import needs to happen in a regular install, probably using either the `from .` syntax or adding BoundaryConditions to mprans/__init__.py.  Not a big hurry, but I won't be able to track it down for a few days.

PR #431, opened 2016-07-11
Weighted relaxation function

adimako:
@cekees @tridelat @Giovanni-Cozzuto-1989. We had been going through this a while ago, and since I am completing WaveTools report, I thought it would be nice to check the effect of the RZ function (H_s) to coefficients.

I checked the following curves

### DIFF
--- a/proteus/mprans/RANS2P.h
+++ b/proteus/mprans/RANS2P.h
@@ -834,7 +834,7 @@ namespace proteus
 					   double dmom_v_source[nSpace],
 					   double dmom_w_source[nSpace])
     {
-      double mu,nu,H_mu,uc,duc_du,duc_dv,duc_dw,viscosity,H_s;
+      double mu,nu,H_mu,uc,duc_du,duc_dv,duc_dw,viscosity,H_s,H_s1,H_s2;;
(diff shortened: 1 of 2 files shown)


adimako:
capture

adimako:
Blue dashed is the exponential distro that eleni used which performed decently. The sinfunction is the new one that it is a bit like the smoothed heavinside and the weighted one is a combination of the two. Turns out ("surprise - suprise") that the weighted one is the best, as it has a smooth transition at the beginning to avoid non-physical effects and then it matches the exponential.

adimako:
I think it is reasonable to use the weighted one from now on. Also I will post the flume cases for using them in regression tests. Such tests will be really useful for doing such an analysis

cekees:
I'm not sure why the tests are failing now, given that this is a small change. I'm relaunched the tests just to see if it was some sort of a load issue on travis.

cekees:
Thanks! @nerisaurus the tests mentioned here might be good  for you to look at it when they go up on air-water-vv.

adimako:
@cekees, this branch compiles in copper so I think it should be ok merging it. 
@nerisaurus, I have created a branch in air water-vv where you can find these tests (and other tests as well). Feel free to review and recommend changes to the structures.

adimako:
Merging this one

PR #432, opened 2016-07-13
Finalise gmsh support

tridelat:
This pull request is for finalising the gmsh implementation.

In `Domain.py`, when using `writeGeo`, the following must be defined in order to write a complete and functional geo file:
- facets in 2D to create line loops (facets still defined with vertices as in 3D)
- volumes in 3D to create surface loops
- if a Shape instance is in another shape, use the `setChildShape` or `setParentShape` functions in order to arrange the Domain properly when calling `assembleDomain` (otherwise `gmsh` will mesh over internal shapes).

To do:
- [x] add support for 2D domain
- [x] reduce the various `writeGeo`, `writeGMSH` to one function
- [x] add meshing options interface
- [x] clean up way to switch to gmsh
- [x] optimise (possibly) and move `gmsh2triangle.py` to where it belongs
- [x] clean up logic (loose print statements / single processor to generate mesh)
- [x] test branch with existing test cases
- [x] move `MeshOptions` class away from Domain.py


cekees:
Sounds good.  Thanks for working on this. One thing I'd like to do is move  the MeshOptions stuff out of Domain.py.  I'd like to avoid getting meshing information tangled up with Domain geometries. Sometimes the domain IS a mesh, but that's a special  case. Generally we want the geometry of the domain to be an independent thing. This came up in the discussion with @zhang-alvin yesterday about mesh adaptivity. That's one of the cases where you need an independent domain geometry that can be queried separately. Anyway, the MeshOptions in here are not that big of a deal right now, but it's the sort of thing that tends to accumulate complexity.

proteus/Domain.py
```diff
@@ -59,6 +63,144 @@ def getBoundingBox(self):
             self.L += [xMax-xMin]
 
 
+    def writeGeo(self, fileprefix, group_names=False):
```
cekees: Cool, so adding this in the base class will work. Does it work for all domain types that we currently have, or do we need to check the actual type (or  override writeGeo in the child classes)?

tridelat: It should work for all domain classes as the `.geo` files use the same format for 1D, 2D, or 3D domains. I just realised that `D_Base` does not have `segments`, `facets`, etc. defined in `__init__` (as I usually use SpatialTools to define a Domain), would it be okay to populate those with empty lists `[]` as a default? This way, the `writeGeo` will write what has been defined and ignore empty lists.

tridelat:
@cekees I agree, the MeshOptions should be moved somewhere else. Maybe in MeshTools? We could make different type of MeshOptions according to what mesher we are using.
I also plan to move RigidBody and its related functions out of SpatialTools at some point to keep it purely geometrical and to make a more advanced and more readable rigid body module (which could potentially become big with chrono functionality).

tridelat:
@cekees Concerning the `MeshOptions` class, should it be moved to another module or should it be a module on its own?

MeshOptions could be extended to include mesh refinement options in order to generate a .geo file through `SpatialTools`, and make it easy to modify through `Context` options for different test cases like here: [MeshRefinement.py](https://github.com/erdc-cm/air-water-vv/blob/2076ae43576fcf991dca8491b5e50727874b5bd3/2d/wavetank_gmsh_refinement/MeshRefinement.py) (this is for testing on air-water-vv).
I see 2 ways for doing this: 
 - `Domain` needs a `MeshOptions` instance as it is through the domain that the .geo/.edge/etc files are written. Or:
 - `MeshOptions` needs a `Domain` instance so it can append its mesh refinement options to the .geo file generated by the domain.

I think this branch could be merged to master soon, especially if we leave out the refinement options for now.

proteus/gmsh2triangle.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+from proteus.MeshTools import Edge
+from proteus.Profiling import logEvent
+
+def msh2triangle(fileprefix):
```
cekees: I think this could go into MeshTools.  Needs a docstring here, but I assume you're just converging a gmsh .msh file to a triangle format mesh, right?

tridelat: Yes, it creates .ele .node and .edge files from a .msh file. I will add a docstring and move it to MeshTools.

proteus/MeshTools.py
```diff
@@ -6001,3 +6001,75 @@ def getMeshIntersections(mesh, toPolyhedron, endpoints):
...
+from proteus import default_p as dp
+
+class MeshOptions:
+    """
+    Mesh options for the domain
+    """
+    def __init__(self, domain=None):
+        self.Domain = domain
+        self.he = 1.
+        self.use_gmsh = False
```
cekees: I agree with having the gmsh switch here in the MeshOptions (that's an unchecked item in your todo list, but I think you're done on that).

tridelat: Do you mean having something such as `p.domain.MeshOptions.use_gmsh` instead of `p.domain.use_gmsh` or have it completely separated as to something such as `p.MeshOptions.use_gmsh`? (the latter will break things in current tests as it requires the p files to have a `MeshOptions` class instance).
https://github.com/erdc-cm/proteus/blob/add_gmsh_support/proteus/NumericalSolution.py#L242

tridelat:
There is no test running gmsh to generate triangle files but I can create one.

tridelat:
@cekees I made a simple test for gmsh, it generates the mesh and read the first line of the triangle files to check if  it has the right number of nodes, edges, and elements. Would that be sufficient? If we want to check line by line that the output is what is expected, we would need to push pre-generated .ele .edge and .node to compare it.

proteus/NumericalSolution.py
```diff
@@ -239,26 +239,51 @@ def __init__(self,so,pList,nList,sList,opts,simFlagsList=None):
...
+                                                        verbose=10)
+                    if comm.isMaster() and p.genMesh:
+                        tmesh.readFromPolyFile(p.domain.polyfile)
+                        tmesh.writeToFile(p.domain.polyfile)
+                        logEvent("Converting to Proteus Mesh")
+                        mesh=tmesh.convertToProteusMesh(verbose=1)
+                    comm.barrier()
+                    if not comm.isMaster() or not p.genMesh:
+                        mesh = MeshTools.TriangularMesh()
+                        mesh.generateFromTriangleFiles(filebase=p.domain.polyfile,base=1)
```
adimako: File "/home/HR/asd/PROTEUS/proteus/proteus/NumericalSolution.py", line 278, in __init__
    mesh.generateFromTriangleFiles(filebase=p.domain.polyfile,base=1)
  File "/home/HR/asd/PROTEUS/proteus/proteus/MeshTools.py", line 3826, in generateFromTriangleFiles
    cmeshTools.generateFromTriangleFiles(self.cmesh,filebase,base)
I get this error, I will post the files that generate it shortly

adimako: Files are here:
https://github.com/erdc-cm/air-water-vv/tree/add_gmsh_bug/2d/add_gmsh_bug

tridelat: What command do you use to launch the simulation?

tridelat: This should be fixed with commit 15af46752ac1f4e6aa8c5434502e829db8c1433f

PR #433, opened 2016-07-13
Adimako/wave tools nl fix

adimako:
I improved the `RandomNLWaves` and `RandomNLWavesFast` classes. This should not affect any other classes. If passes tests I will go ahead an merge


adimako:
I am merging this at it passed tests

PR #436, opened 2016-07-15
Fixed bug that caused simulations to hang

adimako:
Bug that caused simulation to hang when BoundaryConditions class tried to use mwl from RandomNLWavesFast


cekees:
Looks like this should go in too, right?

Issue #428, opened 2016-07-01
What's the future of SpatialTools etc.?


nerisaurus:
@cekees @tridelat @adimako 

As I've been planning out updates to the air-water-vv repository, I've noticed a lot of new developments in:
- Domain
- WaveTools
- Gauges
- BoundaryCondition
- and SpatialTools

interfaces (I've especially been paying attention to developments in #397, mostly because the new documentation is really helpful).  

Along with some minor questions about the current implementation, my main ill-defined question is what do we want the code-flow of an ideal proteus problem to look like while it's using these tools and how can I help out? The air-water-vv repository I'm working with has a whole lot of code following dozens of different paradigms in proteus design code-flows.  Most of them are rather ad-hoc (for understandable reasons), leading to tests which quickly become obsoleted and break down (see erdc-cm/air-water-vv#105 for illustration).

I have a variety of aforementioned questions and suggestions which I'll put in individual comments below for easier discussion.  My suggestions are a lot of work that I don't want to force other people to do, but I'm mentioning them instead of working on them because I don't want to unilaterally make major changes to the functionality while there's so much simultaneous development of the tools, and while other people are still working out how to use the tools in projects.

nerisaurus:
#### `setDimensions` seems to be doing too much
In individual Shape classes (in both proteus/SpatialTools and proteus/mprans/SpatialTools) like `Rectangle` or `Tank3D`, the convention appears to be that there is one core class method called `setDimensions`.

Aside from taking an input and setting it into `self.dim` for the shape, the method also creates all the shape's geometry: vertices, segments, facets, regions, boundary tags...

I feel this is rather unintuitive.  The name setDimensions doesn't suggest that it will do this, and it ends up getting called by tons of methods in the class as they all rebuild the geometry after setting up their new parameters, with each one having to reset the dimensions when they do so (leading to the code repeatedly doing, basically, `self.dim = dim = self.dim`).  

I think some of the calls to rebuild the geometry may be a bit excessive, but I can understand that the redundancy makes it easier for users to change things and not have to worry about whether or not they reconstructed the shape - and the construction even of a complicated shape is probably not expensive compared to what will happen afterwards.  

[...]

PR #437, opened 2016-07-22
creation of BodyDynamics module

tridelat:
@Giovanni-Cozzuto-1989 
This PR is for separating SpatialTools and body dynamics related code. (SpatialTools can still be used to create a rigid body, but only geometrical functions will be kept there)
In the long run, it might be used for coupling with Chrono, mooring dynamics, etc.
- [ ] add new functions related to sliding caisson case
- [ ] verify that test cases run as before and that the new module is free of bugs
- [ ] create new subclasses of RigidBody for special case (overriding the default functions such as `getAcceleration`)
- [ ] clean up / optimise code
- [ ] move all functions related to rigid body away from spatialtools
- [ ] add automated tests for rigid bodies


proteus/mprans/SpatialTools.py
```diff
@@ -83,7 +84,7 @@ def _attachAuxiliaryVariable(self, key):
         """
         if key not in self.auxiliaryVariables:
             if key == 'RigidBody':
-                self.auxiliaryVariables[key] = True
+                self.auxiliaryVariables[key] = bd.RigidBody(self)
```
nerisaurus: Lots of codes like to mess around with `dt_init` and `cfl_target` values as well.  Should Shapes hold onto the default values (instead of, or in addition to, how BodyDynamics is the only one with control over them currently), pick up user defined values when initialized in the user code (as an optional thing to specify for a shape) or when the rigid body is declared, and pass them through here?  Or do you plan on having the user access those values directly from BodyDynamics via some method there?

cekees: I think dt_init and cfl_target need to stay out of Shapes. The RigidBody objects sometimes need to know those, but they don't need to be modifying them. Those are numerical method parameters for the TimIntegration/SplitOperator/StepControl objects.

nerisaurus: Perhaps it would be better to have Shapes hold default values of `None` for those variables.  If RigidBody's are not used, then it doesn't seem to have a need for them - and if they are, the user then would be forced to tell the Shape the parameters they have (and then Shape would pass them to BodyDynamics here).

Or, of course, a method for passing the values directly to BodyDynamics - or pulling them out of Context (but while `dt_init` is one of the most consistent variables in the tests I've seen, I'm wary of constantly pulling things out of Context and hoping the user understands that they needed those particular variables to be there in the first place - and `cfl_target`might be even more obscure [especially if it doesn't have to necessarily be proportional to `runCFL`] ).  

Alternatively, this might suggest that time stepping / CFL parameters might deserve their own class as per [Tristan's design here](https://github.com/erdc-cm/proteus/issues/428#issuecomment-230259668), allowing us to pull them out of the hat with a little less mystery and obscurity than just plain variables in Context.

In any case, it's useful to know which access method is going to get used in order to anticipate the changes needed to fit it in.

tridelat:
@Giovanni-Cozzuto-1989 
For the sliding caisson, I think it would be nice to create a subclass and override some functions of the base class `RigidBody` instead of trying to integrate everything in the base class, because the sliding caisson is a very specific case and only works for a rectangular shape it seems. The base class should be as generic as possible so it is easy to build new subclasses.
Creating a subclass can be done like this:
```python
class SlidingCaisson(RigidBody):
    def __init__(self, shape):
        super(SlidingCaisson, self).__init__(shape)  # initialises everything from RigidBody
        ...  # override variables here
        # add / override functions
```
Then in the test case, it could be created like this:
```python
my_rect = SpatialTools.Rectangle(dim, coords)
my_caisson = BodyDynamics.SlidingCaisson(my_rect)  # auxiliary  variable will still be attached
```
Let me know your thoughts.

cekees:
This PR changes 30 THOUSAND files and 1M lines of code! It looks like there as been an accidental `git add *` done... I think you may need to do some git-filter-branch cleanup to remove the accidentally added directories from both the branch and the repository history: https://help.github.com/articles/removing-sensitive-data-from-a-repository/. @zhang-alvin may be able to give you some pointers as I think he did that fairly recently on the scorec branch.

zhang-alvin:
Unfortunately, my experience with modifying repository history is limited to the simpler task of migrating files to git-lfs via a tool (git-lfs-migrate). I haven't yet attempted anything complicated with git-filter-branch beyond permanently removing a single file.

Giovanni-Cozzuto-1989:
Unfortunately something went wrong with the git add command. It's an insane number of additions! Sorry about that! I'm pretty sure I didn't use git add * or git add . commands. The weird thing is also that it didn't take a lot to push this abnormal number of files. Maybe it's worth to git reset on a previous commit and add again only the files I need. @cekees

cekees:
Thanks for cleaning that up. I think in general the files you are adding show up below the comment line when you do `git commit`, so you can usually abort the commit and fix it at that stage. One thing I noticed is that the problematic commits are still on github https://github.com/erdc-cm/proteus/commit/4e469e65cf2dc9f9f7e4e5abd50d81d508bb68d5, but they aren't in my local repository. Not sure why that is. It may be that github waits to do some garbage collection. Anyway, I'm not worried about it for now. Carry on!

tridelat:
Closed as it the changes here are now tracked by #528

PR #439, opened 2016-07-26
Introduced quality checks for maintaining consistency during window r…

adimako:
…econstruction and relaxation zone


adimako:
Needs testing

adimako:
Test file updated, passed the tests so I am merging this. Key additions:

- Included a check when `TimeSeries` is used in the relaxation zone. This will allow for consistency when forcing the analytical solution in space. More details in the report
- Included a `windout` function that returns window parameters. Written for testing purposes, I think it is a good idea to leave it in
- Included an accuracy check in `RandomWavesFast`. Max error from the original time series must be 1% otherwise the code will fail. We can disable this by setting `checkAcc = False` in the class
- Changed default values of window function, after some sensitity tests
- Added relevant tests

cekees:
+1

PR #440, opened 2016-07-29
Add bdm2

alistairbntl:
I still need to rebase and refactor but this is a working first draft of the BDM enhancement.


proteus/Transport.py
```diff
@@ -3555,9 +3559,11 @@ def calculateExteriorElementBoundaryCoefficients(self):
                 if self.coefficients.advection.has_key(ci):
                     self.ebqe[('advectiveFlux',ci)][t[0],t[1]] = g(self.ebqe[('x')][t[0],t[1]],self.timeIntegration.t)
                     for cj in self.coefficients.advection[ci].keys():
+                        #
```
cekees: I would try to minimize this kind of touching of the source that doesn't really introduce a change when it's in the context of a feature request. As we converge on formatting styles we can do those kinds of reviews separately.

proteus/Transport.py
```diff
@@ -6504,6 +6509,8 @@ def initialize(self,
                     if ts.dofMap.nDOF_all_processes != par_N:
                         mixed=True
                 par_nghost = 0
+                subdomain2global = trialSpaceDict[0].dofMap.subdomain2global
+                max_dof_neighbors= trialSpaceDict[0].dofMap.max_dof_neighbors
                 log("Allocating ghosted parallel vectors on rank %i" % comm.rank(),level=2)
```
cekees: This is where I would have expected the ParVec...TH class to get used, but I don't see it here.

proteus/cfemIntegralsModule.c
```diff
@@ -6850,6 +6900,10 @@ static PyMethodDef cfemIntegralsMethods[] = {
     cfemIntegralsCalculateWeightedShapeTrace,
     METH_VARARGS, 
     "The quadrature weighted shape  functions"},
+  { "calculateWeightedShapeTrace",
```
cekees: I think this should be "calculateWeightedPiolaShapeTrace"

proteus/postprocessing.c
```diff
@@ -3767,12 +4275,12 @@ void calculateConservationResidualPWL(int nElements_global,
 	 /*mwf debug
 	 printf("calcConsResPWL eN=%d nN=%d starR=%g\n",eN,nN,
 		starR[nN_global][eN_star]);
-	 */
+	 //	 */
```
cekees: not a big deal, but I'd removed the mixed // and */ comments within the multiline /* */ comment

proteus/MeshTools.py
```diff
@@ -3734,6 +3740,29 @@ def computeGeometricInfo(self):
...
+        tmesh = TriangleTools.TriangleBaseMesh(baseFlags="Yp",
+                                               nbase=1,
+                                               verbose=False)
+        tmesh.readFromPolyFile(polyfile)
+        mesh = tmesh.convertToProteusMesh(verbose=0)
+        mesh.partitionMesh()
+        mesh.globalMesh = mesh
+        return mesh
+    if nd == 3:
+        pass
```
cekees: You implemented Domain.unitSimplex(3) above so I guess the 3D should be just ripping out the code from NumericalSolution that calls tetgen on a 3d polyfile.

setup.py
```diff
@@ -83,7 +83,7 @@
                              library_dirs=[PROTEUS_LAPACK_LIB_DIR,
                                            PROTEUS_BLAS_LIB_DIR],
                              libraries=['m',PROTEUS_LAPACK_LIB,PROTEUS_BLAS_LIB],
-                             extra_compile_args=PROTEUS_EXTRA_COMPILE_ARGS,
+                             extra_compile_args=PROTEUS_EXTRA_COMPILE_ARGS+['-g'],
```
cekees: would remove before merging

proteus/tests/post_processing/test_bdm_projection.py
```diff
@@ -0,0 +1,199 @@
...
+        self.bdm2_obj.evaluateLocalVelocityRepresentation(0)
+
+        comparison_vec = np.load('./comparison_files/trig_velocity_rep.npy')
+        assert np.allclose(self.bdm2_obj.q[('velocity',0)],comparison_vec)
+
+
+def test_bdm_sshaped_region():
+    '''
+    Tests a post-processed velocity field projected
+    into bdm1 space.
```
cekees: does this test need to be extended to BDM2?

proteus/LinearAlgebraTools.py
```diff
@@ -131,6 +131,70 @@ def save(self, filename):
         _petsc_view(self, filename)
 
 
+class ParVec_petsc4pyTH(p4pyPETSc.Vec):
```
alistairbntl: @cekees - Hmm, I'm not sure this has anything to do with the BDM patch honestly.  I'm guessing the TH appended to the back refers to Taylor Hood.  Do you recall near the beginning of the year messing around with the parallel support for TH elements?  I think this had something to do with that.

proteus/MeshTools.py
alistairbntl: Okay, I've added the 3D case, but the updates do not appear to be rendering here after I pushed up the changes? 

Anyway, allow me to summarize what I've done, and some questions I have

1. I'm not entirely sure what I should be returning in the 1D case?  Just a line 0-1?
2. I've updated the old unit_mesh_tests to the new py.test format.
3. I've added the 3D simplex mesh.
4. So I found the code duplication for the tetgen in the NumericalSolution module with the MeshTools unnecessary.  Thus, I wrote a routine in the TriangleTools module that generates a tetrahedral mesh from a polyfile which I call in the nd==3 case.  I'm interested in what you think about (a) using this function call in NumericalSolution and (b) whether TriangleTools is really the correct home for the function.

proteus/LinearAlgebraTools.py
cekees: I did work on fixing the ParVec support for interleaved and non-interleaved (uniform and mixed) spaces. Blame show's you having added this code though. If it's not being used anywhere then I'm OK just cutting it. We shouldn't need a separate parallel vector for Taylor Hood, just different logic in the main ParVec_petsc4py.

proteus/TriangleTools.py
```diff
@@ -11,10 +11,59 @@
 import triangleWrappers
 import TriangleUtils
 import TriangleFileUtils
+import MeshTools
+
```
cekees: This looks good. I would put it in MeshTools.py though. TriangleTools  was supposed to be specifically for `triangle`.  We could add a TetgenTools module and start pulling out a lot of the stuff around tetgen calls (e.g. from NumericalSolution). Anyway, not a big deal right now. I'd just move it to MeshTools

proteus/tests/mesh_tests/test_unit_simplex.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
+"""
+
+Test module for generating unit_simplex elements
+
```
cekees: Thanks for adding this.

cekees:
I forced skpping of the wavetools testRandom case in the master branch because it's failing on most of the platforms I can test on. I was planning to discuss it with @adimako this week, but it doesn't need to hold up this PR. In fact if your merge origin/master into your branch the failure will go away.

proteus/TriangleTools.py
```diff
@@ -30,7 +30,7 @@ class TriangleBaseMesh:
 
     def __init__(self,baseFlags="zen",nbase=0,verbose=0):
         """
-        initialize the triangulateio object,
+        initialize the triangulation object,
```
cekees: `triangulateio` was the correct thing here

proteus/postprocessing.c
```diff
@@ -2637,6 +2638,9 @@ Input Variables
     nVDOFs_element,nSpace);
   */
 
+  /* printf("BDM1 test information: \n"); */
+  /* printf("nVDOFs_element: %d\n",nVDOFs_element); */
```
cekees: I would only leave code in comments like this if it's complicated debugging code that is likely to get used again.

proteus/PostProcessingTools.py
```diff
@@ -622,6 +675,7 @@ def __init__(self,vectorTransport=None,vtComponents=[0],omitFluxBoundaryNodes=Tr
                 self.solutionTestSpaceIsNotPWL[ci] = True
                 logEvent("pwl post-processing for finite element space"+str(self.vt.u[ci].femSpace))
                 #need to set up w and w*dS_f and weights to build residuals for linears out of R_{e,i}
+                #Point1
                 if self.testSpace == None:
```
cekees: While we're updating this file it would be good to replace the `== None` tests with `is None` so we don't generate so many warnings.

cekees:
It looks like bdm2 is broken for 3D. I turned it on in the two poisson tests just to check. Am I doing something wrong?

proteus/TriangleTools.py
cekees: I take that back. This is talking about the triangulation object.

PR #441, opened 2016-08-10
Structured Mesh / SpatialTools support

nerisaurus:
A cheap and easy fix to allow structured meshes using defaultRectangularDomain to integrate into the post-SpatialTools workflow.  More elaborate techniques could be used (I note that the PSLG does a redundancy check which could be useful for this) but it isn't really a pressing issue.  The domain will obviously fail to adapt to any sort of interesting spatial tools usage.


cekees:
Looks good.

PR #442, opened 2016-08-10
Boundary Condition Changes to support air-water-vv refactor

nerisaurus:
setHydrostaticPressureOulet (and the WithDepth variant) are now named in the same style as the other conditions.  setTwoPhaseVelocityInlet is uncommented because it has several air-water-vv uses (unless a WaveTools based alternative can be made).  Some minor changes are made to its previous implementation to start to try to adapt it, but I'm not quite confident in it.  A few minor cosmetic changes (error declarations, mostly) to other functions.


cekees:
What do you think we should do to test this? Maybe add a very small rectangular tank with constant inflow, hydrostatic pressure on outflow and atmospheric on top? Maybe a modified form of the broad crested weir as a "smoke" test?

nerisaurus:
If I was confident of the broad crested weir's performance, it would make a good smoke test for both of the untested boundary conditions (it has always been finicky with convergence, though, and is not fast).  

There's been floated the idea of a **steady flow** wave tank (with the numerical tanks) which should run relatively quickly (if it can be made small enough - it shouldn't have any problems with needing a lot of refinement for the waves like the other tanks).  It may either use these two boundary conditions or potentially give us a "wave" way of solving them (so a WaveTools style class that can be passed into more generic boundary conditions and give us these types of outflows/inflows).

In any case, care should be taken to test both the normal hydrostatic pressure outlet and the one with depth.

cekees:
That makes sense. A steady flow tank has some simple things like the water level and discharge that would be easy to turn into assertions for the tests.

proteus/mprans/BoundaryConditions.pyx
```diff
@@ -319,64 +319,70 @@ class BC_RANS(BC_Base):
...
+            def inlet_p_advective(x, t):
+                b_or = self._b_or[self._b_i]
+                u_p = np.sum(U * b_or)
+                # This is the normal velocity, based on the inwards boundary
+                # orientation -b_or
+                u_p = -u_p
+                if x[vert_axis] < waterLevel:
+                    return u_p
+                elif x[vert_axis] >= waterLevel:
+                    return None
```
nerisaurus: @tridelat We (me and @cekees ) suspect that the `return None` here and potentially the possibility of values escaping the `if`/`elif` statements (with no `else` to catch them) may be causing some of the problems with the [hydraulicStructures](https://github.com/erdc-cm/air-water-vv/pull/125).  Do you think this is something you can update for this boundary condition?

tridelat: @nerisaurus, this is possible, as this boundary condition is untested, it looks like it should return a `[0,0,0]` array instead of `None` like for `setUnsteadyTwoPhaseVelocityInlet`

nerisaurus: That makes sense.  The docstring says

> the condition returns the gravity as zero,
        and sets _dirichlet condition to zero, only if there is a zero inflow
        velocity component.

though.  I'm not entirely sure how this would update (the whole `U` input and other parts seems to be an older layer of notation) - it might want to be some kind of wind speed variable (probably set to 0 as default, as in `setUnsteadyTwoPhaseVelocityInlet`) above the water level, although that's not as pressing of a change (that refactoring would clean up and potentially extend the uses for this condition, while making it not return `None` will probably fix some current bugs).

Changing that and potentially closing the `if`/`elif` statements (I think they should only fail if the inputs to the function are wrong [not a number], but an `else` that raises an error warning about this is - while probably not necessary - a good assurance that bad behavior will be noticed) should hopefully solve the issues.

tridelat:
I will go ahead and merge this branch as I think BC changes were finished here and there is no conflict with master

PR #443, opened 2016-08-10
Gauge support for air-water-vv refactor

nerisaurus:
Users can now attach gauges to any RANS shape by the attachPointGauges() (or similar for line and line integral gauges) method, which uses the same inputs as a gauge call plus a label for which problem the gauges should be associated with (e.g. ls, twp, redist).  These are attached to the Domain and can be called from _n.py files in a simple manner.  Errors are raised for misnamed problems (i.e. if someone puts twp_navier_stokes despite us using the shortened name twp in this context).  _n.py files may call for their auxiliary variables whether or not they have any without bad behavior (allow people to turn gauges or other factors on and off without worrying about editing _n.py files).

### DIFF
--- a/proteus/mprans/SpatialTools.py
+++ b/proteus/mprans/SpatialTools.py
@@ -86,13 +86,23 @@ class ShapeRANS(Shape):
         if key not in self.auxiliaryVariables:
             if key == 'RigidBody':
                 self.auxiliaryVariables[key] = True
-            if key == 'RelaxZones':
+            elif key == 'RelaxZones':
                 self.auxiliaryVariables[key] = self.zones
-            if str(key).startswith('Gauge_'):
+            elif str(key).startswith('Gauge_'):
                 self.auxiliaryVariables[key] = [gauge]
-        elif str(key).startswith('Gauge_') \
-                and gauge not in self.auxiliaryVariables[key]:
-            self.auxiliaryVariables[key] += [gauge]
+            else:
+                logEvent("auxiliaryVariable key: "
+                         "{key} not recognized.".format(key=str(key)), level=1)
+        elif str(key).startswith('Gauge_'):
+            if gauge not in self.auxiliaryVariables[key]:
+                self.auxiliaryVariables[key] += [gauge]
+            else:
+                logEvent(
+                    "Attempted to put identical "
+                    "gauge at key: {key}".format(key=str(key)), level=1)
+        else:
+            logEvent("Key {key} is already attached.".format(key=str(key)),
+                     level=1)
 
     def attachPointGauges(self, model_key, gauges, activeTime=None,
                           sampleRate=0,
--- a/proteus/mprans/SpatialTools.py
+++ b/proteus/mprans/SpatialTools.py
@@ -66,7 +66,7 @@ class ShapeRANS(Shape):
         self.auxiliaryVariables = {}  # list of auxvar attached to shape
         self.It = None  # inertia tensor
 
-    def _attachAuxiliaryVariable(self, key):
+    def _attachAuxiliaryVariable(self, key, gauge=None):
         """
         Attaches an auxiliary variable to the auxiliaryVariables dictionary of
         the shape (used in buildDomain function)
@@ -75,6 +75,8 @@ class ShapeRANS(Shape):
         ----------
         key: string
             Dictionary key defining the auxiliaryVariable to attach
+
+        gauge: Gauges
 
         Notes
         -----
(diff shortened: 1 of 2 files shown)


proteus/mprans/SpatialTools.py
```diff
@@ -86,6 +88,60 @@ def _attachAuxiliaryVariable(self, key):
                 self.auxiliaryVariables[key] = True
             if key == 'RelaxZones':
                 self.auxiliaryVariables[key] = self.zones
+            if str(key).startswith('Gauge_'):
+                self.auxiliaryVariables[key] = [gauge]
+        elif str(key).startswith('Gauge_') \
+                and gauge not in self.auxiliaryVariables[key]:
+            self.auxiliaryVariables[key] += [gauge]
```
cekees: The logic and indentation here looks a little suspicious. Might want to 3rd and 4th if's into elif's and terminate the blocks with else's so it's clear what happens when the key fall's through.  I may just need to look at the context more carefully.

proteus/tests/test_spatialtools.py
```diff
@@ -223,6 +222,55 @@ def test_BC_flags(self):
...
+        assert domain.auxiliaryVariables['vof'][0].sampleRate == 0.5
+
+        tank.attachPointGauges('voff',
+                                gauges=((('vof',), ((0., 0., 0.),)),),
+                                activeTime=(0., 1.),
+                                sampleRate=0.5,
+                                fileName='point2.csv')
+        self.assertRaises(ValueError,assembleDomainRANS,domain)
+
+
```
cekees: l like the new functionality. Thanks for adding the test.

proteus/mprans/SpatialTools.py
nerisaurus: True, the logic doesn't really give any warning for failed keys (technically this only should be called by other methods, so tests should be handling the heavy lifting of checking for correct behavior, but some else statements would still be nice).  I'm not sure how hard it should fail (as the function might be called multiple times with the same values, breaking out of the if statement intentionally).

The indentation is correct for the original intention of the function, but it depends on the direction of where we take the logic.  The function was made to only accept the first `RigidBody` and `RelaxZones`.  This might be because the values in `self.auxiliaryVariables` do get used a lot elsewhere and a lot of code is run multiple times.  It could be to keep from attaching the keys multiple times, and maybe messing some stuff up with the `RelaxZones` (that is, resetting it might break things?).  The gauges accept any number of unique gauges, so they needed the `elif` to handle new additions.

cekees: Why don't we make this

```
 if key not in self.auxiliaryVariables:
             if key == 'RigidBody':
                  self.auxiliaryVariables[key] = True
+            elif key == 'RelaxZones':
                  self.auxiliaryVariables[key] = self.zones
 +           elif str(key).startswith('Gauge_'):
 +                self.auxiliaryVariables[key] = [gauge]
 +           else:
 +                   logEvent("Was not able to append %s  to auxiliaryVariables".format(str(key))
 +        elif str(key).startswith('Gauge_') \
 +                and gauge not in self.auxiliaryVariables[key]:
 +            self.auxiliaryVariables[key] += [gauge]
 +        else:
+             logEvent("Could not....")
```

nerisaurus: Just to check your indentation, you mean:
```python
if key not in self.auxiliaryVariables:
    if key == 'RigidBody':
        self.auxiliaryVariables[key] = True
    elif key == 'RelaxZones':
        self.auxiliaryVariables[key] = self.zones
    elif str(key).startswith('Gauge_'):
        self.auxiliaryVariables[key] = [gauge]
    else:
        logEvent("Was not able to append %s  to auxiliaryVariables".format(str(key)))
elif str(key).startswith('Gauge_') \
     and gauge not in self.auxiliaryVariables[key]:
    self.auxiliaryVariables[key] += [gauge]
else:
    logEvent("Could not....")
```
correct?  I think that's a good compromise between an error and invisible responses, although I'm not sure how high level the logEvent should be (it depends on how much it gets spammed by repeated calls to the function - if it's a lot, it maybe should be a higher log level).

I'd also put in a third warning (or at least break up the logic to avoid having the second warning cover as much of the error-space):
```python
if key not in self.auxiliaryVariables:
    if key == 'RigidBody':
        self.auxiliaryVariables[key] = True
    elif key == 'RelaxZones':
        self.auxiliaryVariables[key] = self.zones
    elif str(key).startswith('Gauge_'):
        self.auxiliaryVariables[key] = [gauge]
    else:
        logEvent("Phrase One")
elif str(key).startswith('Gauge_'):
    if gauge not in self.auxiliaryVariables[key]:
        self.auxiliaryVariables[key] += [gauge]
    else:
        logEvent("Phrase Two")
else:
    logEvent("Phrase Three")
```

I'd suggest for **Phrase One**:
```python
logEvent("auxiliaryVariable key: {key} not recognized.".format(key=str(key)))
```
and for **Phrase Two:**
```python
logEvent("Attempted to put identical gauge at key: {key}".format(key=str(key)))
```
(although this one may be less necessary, since its repetition is more clearly undesirable)

and finally for **Phrase Three**:
```python
logEvent("Key {key} is already attached.".format(key=str(key)))
```

cekees: Great. I like that. I'd probably make the log level 0 or 1 unless this is a keys/gauges falling through is  something you expect a well-written application to generate a lot  of.

PR #444, opened 2016-08-10
Tank2D for air-water-vv refactor support

nerisaurus:
Tank2D is now more modular and easier to extend.  Some terminology has changed to make method names more intuitive.

### DIFF
--- a/proteus/mprans/SpatialTools.py
+++ b/proteus/mprans/SpatialTools.py
@@ -836,27 +836,28 @@ class Tank2D(ShapeRANS):
         Domain class instance that hold all the geometrical informations and
         boundary conditions of the shape.
     dim: Optional[array_like]
-        Dimensions of the cuboid.
+        Dimensions of the tank (excluding sponge layers).
     coords: Optional[array_like]
         Coordinates of the centroid of the shape.
     from_0: Optional[bool]
-        If True (default), the tank extends from the origin to postive x, y, z
+        If True (default), the tank extends from the origin to positive x, y, z
     """
     count = 0
 
-    def __init__(self, domain, dim=(0., 0.), coords=None, from_0=True):
+    def __init__(self, domain, dim=(0.,0.), coords=None, from_0=True):
         super(Tank2D, self).__init__(domain, nd=2)
-        self.__class__.count += 1
-        self.name = "tank2d" + str(self.__class__.count)
-        self.dim = np.array(dim)
-        self.from_0 = from_0
+        self._nameSelf()
+        self._setupBCs()
         self.spongeLayers = {'x-': None,
                              'x+': None}
-        if coords is None:
-            self.coords = self.dim/2.
-        else:
-            self.coords = coords
-            self.from_0 = False
+        self._findEdges(dim, coords, from_0)
+        self.constructShape()
+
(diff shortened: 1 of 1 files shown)


proteus/mprans/SpatialTools.py
```diff
@@ -836,27 +836,28 @@ class Tank2D(ShapeRANS):
...
     coords: Optional[array_like]
         Coordinates of the centroid of the shape.
     from_0: Optional[bool]
-        If True (default), the tank extends from the origin to postive x, y, z
+        If True (default), the tank extends from the origin to positive x, y, z
     """
     count = 0
 
-    def __init__(self, domain, dim=(0., 0.), coords=None, from_0=True):
+    def __init__(self, domain, dim=(0.,0.), coords=None, from_0=True):
```
tridelat: maybe the `dim` argument should not have a `(0.,0.)` default value anymore in order to force the user to set it directly as `setDimensions` disappeared

proteus/mprans/SpatialTools.py
```diff
@@ -879,71 +880,124 @@ def __init__(self, domain, dim=(0., 0.), coords=None, from_0=True):
...
+                         [5 + added_vertices, 2]]
+            segmentFlags += [self.boundaryTags['y-'],
+                             self.boundaryTags['x+'],
+                             self.boundaryTags['y+']]
+            segmentFlags[1] = self.boundaryTags['sponge']
+            added_vertices += 2
+        return segments, segmentFlags
+
+    def _constructRegions(self, vertices, vertexFlags, segments, segmentFlags):
+        regions = [[0.5 * (self.x0 + self.x1), 0.5 * (self.y0 + self.y1)],]
```
tridelat: For the y coordinates, it might be preferable to put the y (or/and x) coordinates at the top or bottom (or a corner) of the region, where other objects with their own regions are unlikely to clash. In the floating caisson cases, the caisson is usually in the center of the tank for example (I think it is okay though, because it has a "hole", but there could be instances where another object could be in the middle of the tank with its own region).

tridelat:
Thank for this @nerisaurus , nice additions and it adds clarity

proteus/mprans/SpatialTools.py
nerisaurus: It could make sense.  We could also add back in `setDimensions` (I had planned to, but delayed it because I wanted to make sure there wouldn't be conflicts - or rather, that those conflicts would bring up immediate errors when they tried to call `setDimensions` wrongly... I guess I forgot to go back and add it in the end) but with its reduced scope of just setting the dimension and maybe calling `_findEdges` (since plain setters are somewhat discouraged/irrelevant in python).

proteus/mprans/SpatialTools.py
nerisaurus: That does make sense.  I don't like some of the uncertainty of most of the methods I've seen of putting it in a corner (because they have to rely on some sort of arbitrary distance from the actual corner, unlike the center definition which is purely functional), but they are a lot less likely to have issues.  Maybe `self.x0 + 0.001 * (self.x1 - self.x0)` ?

cekees:
+1 to merge

PR #448, opened 2016-08-11
Tank2D with Obstacles - new class for air-water-vv refactoring support

nerisaurus:
This class allows for a simple way to build most air-water-vv style problems where a mostly rectangular tank has obstacles (such as a weir, a sloped bottom, etc.).  The implementation is a bit messy and a more in depth gmsh interface should handle this eventually.  It extends off of the refactored Tank2D.

(a new pull request to bring it directly to master, since Tank2D resolved first and github doesn't have transitive pull requests)

### DIFF
--- a/proteus/mprans/SpatialTools.py
+++ b/proteus/mprans/SpatialTools.py
@@ -1143,6 +1143,688 @@ class Tank2D(ShapeRANS):
             self.BC['x+'].setUnsteadyTwoPhaseVelocityInlet(wave=waves,
                                                            wind_speed=wind_speed)
 
+#[temp] no tests yet!
+class TankWithObstacles2D(Tank2D):
+    """
+    Class to create a 2D rectangular tank with obstacles built out of any wall.
+
+    An obstacle is defined by a contiguous list of points which start and end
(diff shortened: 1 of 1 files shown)


cekees:
Could you post an example of the usage? You could link to where you use this in air-water-vv or link to a gist containing a script that builds a non-trivial domain and plots it.

nerisaurus:
[Here's](https://github.com/erdc-cm/air-water-vv/tree/nneri/refactor/constantSlopeBeach/2d/constantSlopeBeach/Mase_and_Kirby_shoaling) an example from the `constantSlopeBeach` set.  All hydraulicStructure ([in this PR](https://github.com/erdc-cm/air-water-vv/pull/125)) and constantSlopeBeach tests ([in this PR](https://github.com/erdc-cm/air-water-vv/pull/122)) use this, so you can look around at them for other uses.

cekees:
Just to clarify: obstacles boundaries must intersect the tank boundary?

nerisaurus:
Yes.  A ValueError is raised if they do not. Other obstacles can still be included via additional shapes.

PR #450, opened 2016-08-19
scorec/linux: PUMI and MeshAdapt

zhang-alvin:
This is a pull request to start the process of merging the private proteus-scorec branch with the proteus master branch.

**Description**
The proteus-scorec branch adds the ability to interface with the SCOREC libraries, namely PUMI (Parallel Unstructured Mesh Interface). This allows us to use tools like MeshAdapt and ParMA for mesh adaptation and predictive load balancing. An implicit error estimator was also built on top of these libraries which can be used to adapt the mesh to track the flow physics. 

The SCOREC tools require SCOREC model and mesh files. There is an available workflow for using GMSH (open-source) and Simmetrix (private) geometries and meshes. This workflow involves converting the meshes into SCOREC meshes using SCOREC tools. There are limitations with the GMSH workflow since the model/geometry is inferred from the mesh. 

**Compilation Status**
The scorec/linux branch is up-to-date with the master branch and can compile and run with and without Simmetrix libraries.  

**To-Do List**
- [x] Hashstack needs to be updated/merged with hashstack-private branch to enable compilation
- [x] There are various SCOREC tools/executables like split, convert, etc. that can be installed with modifications to the scorec.yaml and scorec-sim.yaml files, but needs to be linked to the $PROTEUS_ARCH/bin or scripts directory for ease of use.
- [x] There are some Simmetrix-only tests that need to be wrapped in a conditional in the Makefile 
- [ ] The dtmb case does not behave the same using the GMSH workflow as compared to the master branch. It appears boundary conditions are not being set right. Simple flow cases like Couette flows, however, run correctly. Added as a potential issue #500 
- [x] The set of tests in MeshAdaptPUMI/tests/ need to be refactored/updated.
- [x] Documentation strings are needed for every function.
- [x] Clean-up environment scripts
- [x] Have tests run on Travis CI properly


cekees:
Thanks. I'll get the travis stack updated so the tests can run.

proteus/MeshAdaptPUMI/ErrorResidualMethod.cpp
```diff
@@ -0,0 +1,860 @@
+#include "MeshAdaptPUMI.h"
+#include <PCU.h>
+#include <petscksp.h>
+
+#include <apf.h>
```
zhang-alvin: apf is a library within PUMI that takes care of abstractions like fields and meshes.

cekees:
Has there been a version of the scorec tools generated that doesn't require simmetrix in order to build? I'd like to start testing the gmsh workflow, but I can's seem to use cmake to "turn off" the simmetrix-dependent code.

zhang-alvin:
At the moment, if a SIM_INCLUDE_DIR path is not specified, the compilation process should use the Simmetrix-independent scorec libraries. See https://github.com/erdc-cm/proteus/blob/scorec/linux/proteus/config/default.py#L125.

Makefile
```diff
@@ -247,6 +247,28 @@ check:
...
 
+	@echo "************************"
+	@echo "SCOREC-Proteus Test"
+	${PROTEUS_ENV} ${PROTEUS_PYTHON} proteus/MeshAdaptPUMI/test/meshLoad/meshLoad.py
+	@echo "************************"
+  
+	@echo "SCOREC-Gmsh Flow Test"
+	${PROTEUS_ENV} ${PROTEUS_PYTHON} proteus/MeshAdaptPUMI/test/test_gmshLoadAndAdapt/gmshCheck.py
+	@echo "************************"
+
```
zhang-alvin: Agreed. I added an additional build target check_simmetrix in the latest commit.

Makefile
```diff
@@ -191,7 +191,7 @@ install: profile $(shell find proteus -type f) $(wildcard *.py) proteus
 	@echo "done installing standard extension modules"
 	@echo "************************"
 	@echo "installing scripts"
-	cd scripts && ${PROTEUS_ENV} PROTEUS_PREFIX=${PROTEUS_PREFIX} make
+	#cd scripts && ${PROTEUS_ENV} PROTEUS_PREFIX=${PROTEUS_PREFIX} make
```
zhang-alvin: Uncommented in commit 6696689b85b7f7333e491b25931f5b43cee1bc53. There was an issue in building the scripts on SCOREC machines because the f77 Fortran compiler was not available, so it was temporarily turned off to avoid the issue.

cekees:
What I meant is that the scorec tools themselves won't currently compile without simmetrix. I'm guessing you guys are always building on systems with simmetrix licenses, but most proteus users will not have them so we need a way to build the scorec tools without simmetrix libraries. Meant to /cc @ibaned.

ibaned:
@cekees be sure to use this repo: https://github.com/SCOREC/core. That seems to be what is in hashstack master: https://github.com/erdc-cm/hashstack/blob/master/pkgs/scorec.yaml#L7 (which btw should not need `SIM_MPI`: https://github.com/erdc-cm/hashstack/blob/master/pkgs/scorec.yaml#L32). That version should not depend on Simmetrix at all. However, hashstack-private points to https://github.com/SCOREC/core-sim (https://github.com/erdc-cm/hashstack-private/blob/roams/pkgs/scorec.yaml#L6), which is the Simmetrix-requiring repository. We just recently deprecated that, and the latest https://github.com/SCOREC/core can build with and without Simmetrix, controlled by CMake flag `ENABLE_SIMMETRIX:BOOL`.

cekees:
Any specific commit I should use for SCOREC/core? I just tried with the latest (77a3ffa5639115a069066d4ca53ae7a44f884312), and the build fails because `-Werror` is set (warnings==errors), and there are warnings.  Otherwise the build seems to be fine. I will push the updated scorec.yaml file to hashstack once I know what commit to use and have run the proteus tests.

ibaned:
Could you send me your configure flags, compiler version, and the error output? We aim to be warning-free so I'd like to fix what you're seeing.

cekees:
I generated it on commit 54fdc9e23ceda4e78054edc059e88c8f444df6f5 of https://github.com/hashdist/hashstack using the proteus.linux2.yaml profile on ubuntu 16.10 with gcc-6.2.0 toolchain. The first warning that generated the error was an unused parameter "rank", but I don't have the line number handy. If you clone proteus into the official docker Ubuntu 16.10 image, you should be able to reproduce it.  To work around it I just ran `hit build --debug` and removed the -Werror flag before running the build. CMAKE input is from the pkg/score.yaml file on that commit:

```
- name: configure
  extra: [
  '-DCMAKE_C_COMPILER:STRING=${MPICC}',
  '-DCMAKE_CXX_COMPILER:STRING=${MPICXX}',
  '-DENABLE_ZOLTAN:BOOL=ON',
  '-DENABLE_SIMMETRIX:BOOL=OFF',
  '-DZOLTAN_PREFIX:PATH=${ZOLTAN_DIR}',
  '-DPARMETIS_PREFIX:PATH=${PARMETIS_DIR}',
  ]
```

proteus/MeshAdaptPUMI/MeshConverter.cpp
```diff
@@ -0,0 +1,412 @@
...
+      mesh.nNodes_element = 4;
+      mesh.nNodes_elementBoundary = 3;
+      mesh.nElementBoundaries_element = 4;
+      break;
+    default:
+      apf::fail("dimension is not 2 or 3\n");
+      break;
+  }
+
+  std::cerr << "*******Proteus Mesh Stats*********\n";
```
zhang-alvin: I can see this being an issue with large multiprocessor runs with multiple stages of mesh adaptation, but it is also very useful information while testing/debugging as a sanity check. I'll place those lines within a condition statement that depends on a user specified flag.

proteus/MeshAdaptPUMI/MeshFields.cpp
```diff
@@ -0,0 +1,223 @@
...
+{ 
+ rho[0] = rho_p[0]; rho[1] = rho_p[1];
+ nu[0] = nu_p[0]; nu[1] = nu_p[1];
+ g[0] = g_p[0]; g[1] = g_p[1]; g[2] = g_p[2];
+ return 0;
+}
+
+int MeshAdaptPUMIDrvr::transferBCtagsToProteus(int* tagArray,int idx, int* ebN, int*eN_global,double* fluxBC)
+{
+/*
```
zhang-alvin: This is a remnant of an attempt at passing the BCs specified by the user into the MeshAdapt portion of the code. It was functional for a serial case, but there were issues when running in parallel that were not resolved. I will comment the function out until it is usable.

proteus/MeshAdaptPUMI/MeshConverter.cpp
cekees: That sounds good. If it's debugging info then use a macro and leave it turned off by default. If it's global (simulation-wide) info that is useful then put it inside something like `if(rank == 0){}` so we don't get a duplicate print statement for every MPI task.

zhang-alvin:
It's actually not a duplicate statement for each MPI task. The names of the
variables are misleading. It will print information about the local mesh
part (number of nodes, edges,etc). Either way, I think using a macro will
be a good idea.

proteus/MeshAdaptPUMI/MeshConverter.cpp
zhang-alvin: It's actually not a duplicate statement for each MPI task. The names of the
variables are misleading. It will print information about the local mesh
part (number of nodes, edges,etc). Either way, I think using a macro will
be a good idea.

cekees:
There's still something going on with the tests in the travis environment that I don't understand--a lot of tests now fail, not just your two new ones.  I verified that `make test` and the  commands from .travis.yaml on an installed version of proteus complete with no errors on both ubuntu 16.10 and debian jessie.  I think you should proceed on this branch with adding more  tests, and I'll figure out what's going on with travis.  I need to reconfigure the buildbot workers to properly use git-lfs anyway.

cekees:
I had to do several things to fix the tests, some of which show up in https://github.com/erdc-cm/proteus/blob/scorec/linux/.travis.yml and some of which show up in changes to the stable/proteus branch of hashstack.

- install git-lfs in travis build env and fetch/checkout lfs (.travis.yaml) so we have mesh inputs to tests
- build scorec libraries as shared (stack/pkgs/scorec.yaml)--the static libraries don't seem to be relocatable
- (possibly irrelevant), but I got rid of an extra installation of blas due to the way `use:` is implemented in hashdist
- switched back to full virtual machine build (also possibly irrelevant but easily tested later)

The tests now run without building any of the stack (except proteus). That is, scorec, zoltan, etc. are downloaded from the ubuntu 14.04 build mirror.

/cc @ibaned @zhang-alvin

cekees:
This branch is getting close to being ready to merge, but there are still a few things I want investigate on the changes to NumericalSolution etc.

proteus/NumericalSolution.py
```diff
@@ -7,6 +7,7 @@
 
 import os
 import numpy
+numpy.set_printoptions(threshold='nan')
```
cekees: we should remove this

proteus/NumericalSolution.py
```diff
@@ -935,6 +979,265 @@ def calculateSolution(self,runName):
...
+                        assert(m.stepController.t_model_last == mOld.stepController.t_model_last)
+                    if self.archiveFlag == ArchiveFlags.EVERY_SEQUENCE_STEP:
+                        #hack for archiving initial solution on adapted mesh
+                        self.tCount+=1
+                        for index,model in enumerate(self.modelList):
+                            self.archiveSolution(
+                                model,
+                                index,
+                                self.systemStepController.t_system_last+1.0e-6)
+                  ##chitak end Adapt
```
cekees: We should probably put this adaption block into a member function and call it here.

proteus/NumericalSolution.py
```diff
@@ -946,6 +1249,12 @@ def calculateSolution(self,runName):
         for index,model in enumerate(self.modelList):
             self.finalizeViewSolution(model)
             self.closeArchive(model,index)
+
+        #Destroy the mesh data structures in C because garbage collection doesnt take care of it
+        #import cmeshTools
+        #m = self.modelList[0]
+        #lm = m.levelModelList[0]
+        #cmeshTools.deleteMeshDataStructures(lm.mesh.cmesh)
```
cekees: I think this can be removed safely now.

proteus/cmeshToolsModule.cpp
```diff
@@ -1746,15 +1761,14 @@ static PyObject* SparsityInfo_getCSR(SparsityInfo *self,
...
-          assert(rowptr_[I]+offset < nnz);
           colind_[rowptr_[I]+offset]=*sit;
           offset++;
         }
       std::sort(&colind_[rowptr_[I]],&colind_[rowptr_[I+1]]);
       max_nonzeros = std::max(max_nonzeros,rowptr_[I+1] - rowptr_[I]);
     }
   //std::cout<<"Proteus: Maximum nonzeros in any row is "<<max_nonzeros<<std::endl;
-  return Py_BuildValue("(O,O,i,O)",PyArray_Return(rowptr),PyArray_Return(colind),nnz,PyArray_Return(nzval));
+  return Py_BuildValue("(N,N,i,N)",PyArray_Return(rowptr),PyArray_Return(colind),nnz,PyArray_Return(nzval));
```
cekees: Just to clarify: You used "N" here to ensure that `del` works from python?

zhang-alvin: That sounds right. Looking at the documentation for `Py_BuildValue`, specifying a "O" increments the reference count on an object on top of the reference created during the object construction, whereas "N" does not. This allows `del` to remove all references to an object to trigger garbage collection

proteus/cmeshToolsModule.cpp
```diff
@@ -1728,12 +1743,12 @@ static PyObject* SparsityInfo_getCSR(SparsityInfo *self,
 //     }
   int nnz=0;
   int dim[1];
-  dim[0] = int(self->columnIndecesMap.size())+1;
+  dim[0] = self->columnIndecesMap.size()+1;
```
zhang-alvin: In retrospect, the commit for the memory leak fix was sloppy. I have to double check whether this change was necessary.

proteus/NumericalSolution.py
zhang-alvin: Removed as of https://github.com/erdc-cm/proteus/pull/450/commits/15f7670c1764fdf15eac0f639a0e2d470fe6bc37

proteus/NumericalSolution.py
zhang-alvin: This was a statement used for debugging. Removed as of https://github.com/erdc-cm/proteus/pull/450/commits/15f7670c1764fdf15eac0f639a0e2d470fe6bc37

cekees:
@zhang-alvin thanks for getting this down to 37 files. @mfarthin you may want to review this now as we'd like to have this go into the 1.3 release sometime in March and merge into master before the workshop so we can all run adaptive examples then.

proteus/cmeshToolsModule.cpp
zhang-alvin: Reinserted the casts and assert statements in a0626121e540dbe69a3712b126ccadde16628012.

proteus/NumericalSolution.py
zhang-alvin: Defined the error estimation and adaptation blocks as member function as of 85487123de8079cc48e45e604e05ecf1f5727013

PR #452, opened 2016-10-03
removed time dependence from the pressure subgrid error

cekees:
@adimako @tridelat @Giovanni-Cozzuto-1989 : I was reading about subgrid scale turbulence modeling with methods very similar to what we have in RANS2P, but I noticed that they don't include the term that depends on the time step size within the pressure subgrid error term. If it's easy, could you check one of the tests that showed pressure spikes when the time step size is cut to match the output? I ran this branch on a floating_bar test and the pressure looks smoother. BTW, because the modified formula is in RANS2P.h and RANS2P2D.h you'll have to run 'make develop' (or make install if that's how you build proteus) to rebuilt the RANS2P module.

### DIFF
--- a/proteus/mprans/RANS2P.h
+++ b/proteus/mprans/RANS2P.h
@@ -1026,7 +1026,8 @@ namespace proteus
       cfl = nrm_df/(h*density);//this is really cfl/dt, but that's what we want to know, the step controller expect this
       oneByAbsdt =  fabs(dmt);
       tau_v = 1.0/(4.0*viscosity/(h*h) + 2.0*nrm_df/h + oneByAbsdt);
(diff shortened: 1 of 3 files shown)


tridelat:
I ran the same wave sloshing test case to compare:
1. the old RANS2P saving data every 0.1s
2. the old RANS2P without saving
3. the new RANS2P saving data every 0.1s
4. the new RANS2P without saving

The graphs show the pressure at the left boundary over time. This is without turbulence modelling (`useRANS=0`).  The graph at the bottom is the same as the one at the top with offsets for each curve to see the difference more easily. The pressure jumps are still there (but not exactly the same) when recording data at regular intervals, and the solution differs slightly between the old and new RANS2P (the mesh was coarse: 1 characteristic element length per waveheight). 

plot
plot2

PR #455, opened 2016-10-19
cython optimisation of BoundaryConditions modules

tridelat:
This branch is to optimise the BoundaryConditions module with cython
#### Optimisation:
- [x] Optimise RelaxationZone (08686ad: total ~80% wall clock time reduction)
  - [x] main iteration loop over elements (08686ad: 90+% wall clock time reduction)
  - [x] wave velocity function (08686ad: ~45% wall clock time reduction)
    - Possible further improvements:
      - [ ] `cimport` WaveTools when #456 is ready (waves currently imported as python objects, making it expensive to call)
      - [ ] reduce `smoothedHeaviside` cost (also imported in python)
- [x] Optimise BC_RANS
- [x] Optimise BC_Base
- [x] Convert to .pyx files to .py and .pxd files
- [x] Update tests
#### Changes for test cases:
- **smoothing** distance is now passed as an argument in `setGenerationZones` (SpatialTools) and `setUnsteadyTwoPhaseFlowVelocityInlet` (BoundaryConditions) in order to remove implicit dependencies from the `Context` that would break the simulation if specific variables were not defined (`he` and `ecH`)
- **getContext()** not necessary anymore as boundary conditions do not need information from context anymore (call to be removed in _so.py files using SpatialTools/BoundaryConditions).


tridelat:
@cekees, this is ready to be merged. It improves the speed of BC greatly, which makes a big difference in the case of relaxation zones.
`cimport WaveTools` and using the wave classes does not work because fused types of cython (http://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html, that would be used to declare a wave object from different possible wave classes) cannot be used as a member variable of a class.
So, for now, waves are accessed as a python object in the BC (but the waves' functions are still converted in C). Maybe we can think of another way to import the waves but this is still a very significant speed improvement.

proteus/BoundaryConditions.py
```diff
@@ -1,37 +1,42 @@
...
+        if b_or is not None:
+            self._b_or = b_or[b_i]  # array of orientation of all boundaries of shape
 
-    @classmethod
-    def newGlobalBC(cls, name, default_value=None):
-        """
-        Makes a new BoundaryCondition class instance attached to BC_Base.
-        This creates a new class attributed that will be added to all BC_Base
-        class instances.
+    # @staticmethod
```
cekees: There's commented out code here and in some of the other larger diffs. I've got no problem with that, just making sure you feel these comments are worth saving in the source file.

cekees:
Great. Go ahead and merge it when you're ready.  I should be able to generate the docs for this now, right? Since you worked out how to use the .py and .pxd approach?

proteus/BoundaryConditions.py
tridelat: I left it commented out there because these were useful member functions that could not work anymore once cythonised, but could potentially be revived if a solution is found.

tridelat:
Yes, I think the docs for this should work now as there is a .py file

PR #456, opened 2016-10-19
Wt optimisation

adimako:
# Optimize waveTools

[ ] Created c header where eta and u functions will be called from
[ ] Compiled and tested for Monochromatic Waves - Linear
[ ] Compile and tested for other wave conditions 
[ ] Merge changes from #455 
[ ] Monitor performance improvement.


adimako:
@cekees @alistairbntl @manuel-quezada. Our only problem with these is that we cannot use directly C++ objects from WaveTools classes to Boundary conditions module. such as this pointer https://github.com/erdc-cm/proteus/blob/WT_optimisation/proteus/WaveTools.py#L720 for calculating eta

cekees:
OK, I'm for merging this and addressing or laying out  a strategy for the remaining issue at or after the workshop.  Is the problem basically that you can't call or use C++ objects from one cython module in another cython module?  I know you can do that directly within the Python C API, so I think we can figure something out if that's the case.  For example with some of the linear algebra we create a matrix from python, which actually triggers initializations of C data structures, then those python objects get passed into an extension module, exposed as the underlying C data structures, and operated on by C API functions from other modules.  If the issue is C++ we might be able to refactor as C structs  with functions and do it that way.

PR #465, opened 2016-11-30
dt_system_fixed as maximum time step when used with CFL

tridelat:
@cekees, this is a small modification to use `dt_system_fixed` as maximum time step (if set) when using CFL.


cekees:
Looks good.

PR #472, opened 2016-12-15
Cleanup poisson tests

alistairbntl:


alistairbntl:
@cekees - Something didn't work with the Fortan complier?  Any thoughts?

setup.py
```diff
@@ -405,8 +405,6 @@
                   (os.path.join(proteus_install_path,'tests'),
                    ['proteus/tests/hex_cube_3x3.xmf',
                     'proteus/tests/hex_cube_3x3.h5']),
-                  (os.path.join(proteus_install_path,'tests','ci'),
```
cekees: To got the tests to work in both dev and install mode I had to explicitly list the data files to be copied. When you removed the bdm data set it started giving an error when copying.

PR #477, opened 2016-12-17
Wt optimisation

adimako:
- Velocity profile bug (fixed)
- Memory leak bug (pending)

### DIFF
--- a/proteus/WaveTools.h
+++ b/proteus/WaveTools.h
@@ -91,40 +91,70 @@ namespace proteus
 
   }
 
-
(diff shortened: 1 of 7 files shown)


cekees:
I think compiling with valgrind is a good idea. I will like at the approach you used and see if there is a nice way to add a debugging  option to the build that enables it across the whole proteus build if not also petsc and some of the other important dependencies. Also, I'm noticing a lot more use of the C++--11 shared pointers, which we can start doing in our C++ modules to make memory allocation safer.

PR #479, opened 2016-12-18
Stable/hydra/gcc4.4.2

adimako:
@tridelat @cekees @alistairbntl @Giovanni-Cozzuto-1989. I have managed to compile this in hydra and works. This is not to be merged but rather stay as a stable version for hydra with gcc4.4.2. I had to roll back some wrapper files (see comparisons) to make it work. 

Regarding waveTools, valgrid does not give me anything suspicious but I keep getting the memory increase after all. I do not think it is likely that [WaveTools.h](https://github.com/erdc-cm/proteus/blob/stable/hydra/gcc4.4.2/proteus/WaveTools.h) file  has memory leakage; in the few places that dynamic memory is allocated, I am ensuring that it is deleted when the pointer is useless (.

My concern is that I do not use a __dealloc__ in the wavetools classes and the c++ objects that are dynamically allocated are not destroyed. I am not sure if we construct the WaveTools class once or multiple times within the Boundary conditions module. If the latter is the case, that would explain it. What do you think @tridelat


adimako:
Apparently there is memory leaking during eta and u calls. I created a dummy loop and I was able to increase memory usage x3 by looping over 1million times over the u calls. Looking into it

adimako:
@cekees, Where should I push the stack for this version. Should I also reference the commit to .hashstack_default?

cekees:
Yes, you can store commit in file and do a pr on hastack. @tridelat has done one before. Thanks for straightening out the hydra install.                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone on the Verizon Wireless 4G LTE network.                                                                                                                                                                                                                From: Aggelos DimakopoulosSent: Wednesday, December 21, 2016 6:40 AMTo: erdc-cm/proteusReply To: erdc-cm/proteusCc: Chris Kees; MentionSubject: Re: [erdc-cm/proteus] Stable/hydra/gcc4.4.2 (#479)@cekees, Where should I push the stack for this version. Should I also reference the commit to .hashstack_default?

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/erdc-cm/proteus","title":"erdc-cm/proteus","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/erdc-cm/proteus"}},"updates":{"snippets":[{"icon":"PERSON","message":"@adimako in #479: @cekees, Where should I push the stack for this version. Should I also reference the commit to .hashstack_default?"}],"action":{"name":"View Pull Request","url":"https://github.com/erdc-cm/proteus/pull/479#issuecomment-268538438"}}}

cekees:
@adimako should we close this without merging?

adimako:
@cekees Yess I will close it. Let's keep the branch for a while though

PR #480, opened 2016-12-20
Fixed leaks - cherry pick from hydra/stable

adimako:
The leaks are fixed. I was doing something naive: starting new pointers within the C++ scope, which turns out that it is too troublesome to deallocate. 

Now the velocity is passed as a pointer through the argument list and it is updated within the function, therefore all _cpp_u functions are treated as void.

I will do one minor optimisation yet and then hopefully we merge and resolve this Bug

### DIFF
--- a/proteus/WaveTools.h
+++ b/proteus/WaveTools.h
@@ -17,7 +17,7 @@ namespace proteus
  const double 	Pi17_ =  (1.7*PI_);
 
 
- inline double fastcosh(double k, double Z , bool cosh)
+ inline void fastcosh(double * hype, double k, double Z )
  {
 
        double Kd = k * Z;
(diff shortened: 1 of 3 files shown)


adimako:
@Giovanni-Cozzuto-1989 @cekees @tridelat. This will resolve the memory leaks + reduce some calls to cosh for imrpoving speed. I do not know why it fails, I get an error saying that the log is too big?

cekees:
Can you paste in the error message about the log being too big?

adimako:
`The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated`

It did not fail for this though, a conflict caused it

cekees:
I think that's referring to the py.test log. Might want to check if there
is a print statement left inside a loop somewhere.

adimako:
Indeed, I have a print statement for printing some error values, should I remove it

PR #482, opened 2016-12-25
Update jupyter

cekees:
brings the stack,  Dockerfile, and Makefile up to date with latest release of jupyter.


cekees:
@alistairbntl and @adimako I tried using the new "request review" support on github--not sure what you see on your end. I just wanted you to take a look at the changes to WaveTools.py (@adimako) and the laplace tests (@alistairbntl) that I had to make in order to get the tests to pass.

proteus/tests/mesh_tests/test_quads.py
```diff
@@ -79,7 +79,10 @@ def teardown_method(self,method):
     def test_01_FullRun(self):
         import filecmp
         self.ns.calculateSolution('test1')
-        relpath = "comparison_files/poiseulle_xmf.output"
+        if self.ns.ar[0].global_sync:
```
cekees: @alistairbntl the reason I got differences here was that the new stack wasn't building h5py properly, which forces proteus.Archive to generate one xmf file per proc with slightly different metadata. I added back the old file and put in a check to see which file to use for comparison. I also fixed the stack so h5py is installed properly so the test should pass regardless of whether h5py is in the stack or not. BTW, we should probably be comparing the h5 files as well since that's where all the actual data is.  Might be able to use h5diff.

PR #483, opened 2016-12-30
Scorec nosimdep

cekees:
@zhang-alvin this merges in recent changes to master and the stack. We should start being able to add automated regression and unit tests to run with travis.


zhang-alvin:
@cekees Thanks for this. I'll look into creating regression tests on Travis.

PR #487, opened 2017-01-13
Fixing conf files for hydra. Fixing waveTools tests, all running now …

adimako:
…with make test.

### DIFF
--- a/proteus/config/centos.py
+++ b/proteus/config/centos.py
@@ -2,8 +2,8 @@ import os
 from default import *
 
 
-PROTEUS_MPI_INCLUDE_DIR='/apps/openmpi/1.6.4/gcc/4.4.6/include'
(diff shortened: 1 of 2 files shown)


adimako:
@cekees @tridelat @Giovanni-Cozzuto-1989 . This has compiled in hydra with practically no modifications, which is good. I changed the centos.py configuration file to include the new compiler. I also removed some print statements in the wavetool tests and removed the skip decorators, they seem to run fine with make test. Running make test gave me errors with in the test functions for timesteps. I know that we have released a version but I would be keen to merge to the master as modifications are specific to hydra and wavetools and that will save us from maintaining two branches

cekees:
Great, glad it's working.  I'm finding some issues that I'd like to fix to, so maybe we'll just make a 1.2.1 release sooner rather than later.

PR #489, opened 2017-01-14
Cleaned up poisson_2d test output.

alistairbntl:


alistairbntl:
Actually wait before you merge...I've added a bit more stuff.

alistairbntl:
@adimako - Please have a look at this to see if you're okay with the changes I made.  I'm trying to organize the test directory and remove files generated by tests.

alistairbntl:
In that case I believe that the file will not be removed.  There may be another way to handle this, but usual with the class structure the `teardown` method will remove the files, I just don't have time to refactor this right now.

adimako:
Hold on a while before you merge...

adimako:
I checked if moving the delete staments before tests but it is not possible to cover all the cases, not without digging into WT and check at which point files are written. @cekees @alistairbntl is there a way to run a script after test_wavetools that cleans up these files?

alistairbntl:
Like I said, I think the best approach is to refactor the tests using the class structure in files like `proteus/tests/matrix_constructors/test_laplace_TH_2D.py`.  I'd be happy to do this at some point, but right now I need to stay focused on something else I'm working on.

alistairbntl:
Since the tests pass I'm okay with merging as is and creating an issue to refactor and clean it up when somebody has the time.

adimako:
@alistairbntl go for it

cekees:
+1.

PR #492, opened 2017-01-24
Wave tools bugfix

adimako:
- Included Stokes drift
- Added more tests for auxiliary functions
- Fixed glitches with old tests
- Added missing documentation

@cekees @tridelat @Giovanni-Cozzuto-1989 @alistairbntl. I am not sure if I should switch back to the "unoptimised" functions because the pressure spikes are consumed relatively quickly, and the cosh profile is usually good enough when phi>-d/2. I think this is something to decide with more testing

### DIFF
--- a/proteus/WaveTools.h
+++ b/proteus/WaveTools.h
@@ -135,12 +135,14 @@ namespace proteus
      double fcos = fastcos(phase);
      double fsin = fastcos(Pihalf_ - phase);
      
+     double C = omega / kAbs;
+     double Udrift = 0.125*9.81*(4*amplitude*amplitude)/(C*depth);
      double UH=amplitude*omega*Uhype*fcos;
      double UV=amplitude*omega*Vhype*fsin;
      //Setting wave direction
      for(int ii=0; ii<nDim ; ii++)
        {
-	 U[ii] += UH*waveDir[ii] + UV*vDir[ii];
+	 U[ii] += (UH-Udrift)*waveDir[ii] + UV*vDir[ii];
        }
 
    }
--- a/proteus/WaveTools.h
+++ b/proteus/WaveTools.h
@@ -117,7 +117,7 @@ namespace proteus
      delete [] VV;
    }
  */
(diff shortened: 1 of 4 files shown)


cekees:
I'm still not sure what is happening with the fast* functions. Assuming we're using the wave generation/absorption zone, then I suppose what could be the problem here is that the fast* functions generate a velocity field which is not quite divergence free, and the pressure field is sort of having to correct for that.

adimako:
I think the spikes are caused because there are discontinuous derivatives in the fastcos function due to truncation errors (I am using a branch approximation, depending on the phase, and the spikes appear when I am switching branches). These may be amplified when a relatively small time-step is used.

cekees:
This looks fine to me.

cekees:
I was checking the WaveTools doc at https://mistral.hrwallingford.com:4443/api/proteus.WaveTools.html and noticed some of the docstrings are not in a form that renders nicely because they don't match the numpy style. I edited one line and double checked that it straightens out the webpage on my machine (`make develop; make doc`). 

FYI, To regenerate the development docs at https://mistral.hrwallingford.com:4443 You can login to http://mistral.hrwallingford.com:8080 with your github credentials, then go to the debian_jessie builder (http://mistral.hrwallingford.com:8080/#/builders/1) and "force" the build on branch WaveTools_bugfix. The class diagrams are working now.

BTW, there is a lot of good information in the docstrings in WaveTools that would format nicely using the Notes, References, and Examples sections in this example: 

https://github.com/numpy/numpy/blob/master/doc/example.py

/cc @adimako @alistairbntl @tridelat @Giovanni-Cozzuto-1989

adimako:
@cekees, is there any way to suppress class attributes?

cekees:
My reading of the sphinx docs is that they shouldn't be on based on our options (but clearly they are). You might try adding 'no-undoc-members' to the autodoc options here:

https://github.com/erdc-cm/proteus/blob/master/doc/source/conf.py#L59

Here is the relevant section of the sphinx manual:

http://www.sphinx-doc.org/en/stable/ext/autodoc.html#confval-autodoc_default_flags

Although I don't know if it has any affect on autodoc, you might also consider prefixing your data and function members by "_" if the aren't intended to be part of the API: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

PR #495, opened 2017-02-05
Update ev

cekees:
@manuel-quezada and @mfarthin I made some modifications to the rans3p classes to get things running with the full conservative level set scheme. It runs with both `useVF=0` or `1` and `openTop=True` or `False`.  The solution looks a little rough near the top with  `openTop=True` though, which makes me think I'm still got some kind of problem at open boundaries. Also running with bdf2 seems to worsen the issue.  Unfortunately this has the changes on master merged in so a lot of unrelated stuff has changed relative to your branch.


proteus/mprans/VOF3P.h
```diff
@@ -1157,9 +1159,15 @@ namespace proteus
 		      bc_u_ext = ebqe_bc_u_ext[ebNE_kb];
 		      // compute flux = flow*bc_u
 		      if (flow < 0)
-			flux_ext = bc_u_ext*flow;
-		      else 
-			flux_ext = 0;
+			{
```
cekees: @manuel-quezada we need to be sure to set the solution at the exterior quadrature points in all of the versions of the method we want to keep using going forward. Same thing goes for NCLS3P.h

PR #497, opened 2017-02-13
Sed bugfix

adimako:
@cekees I refactored it to prevent memory leaking. I found some lines commented out because it would make drag nonlinear. Yet I 'd like these lines in, as they enable a smooth transition from packed situation to a suspended one. The line in the code is below

### DIFF
--- a/proteus/mprans/SedClosure.h
+++ b/proteus/mprans/SedClosure.h
@@ -93,7 +93,7 @@ public:
     
 
     
-    return (weight*gDrag1 + (1.-weight)*gDrag2)*rhoFluid;
+    return uFluid[0] ; //(weight*gDrag1 + (1.-weight)*gDrag2)*rhoFluid;
     }
 
     inline double gs0(double sedF)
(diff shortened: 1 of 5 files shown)


adimako:
https://github.com/erdc-cm/proteus/blob/WT_optimisation/proteus/mprans/SedClosure.h#L81

proteus/mprans/SedClosure.h
```diff
@@ -576,7 +576,7 @@ namespace proteus
 
     
 
-    inline double*  mIntFluid(  double sedF,
+    inline void  mIntFluid( double * mint2, double sedF,
```
cekees: On functions like this that "return" values by setting function arguments, there's an informal (and not consistently followed) convention to put the  inputs first and the outputs last.  mint2 is the output here, right?

adimako: Correct. . I will merge for now but I will update all of them later (I did the same in WaveTools.h)

adimako: Correct. . I will merge for now but I will update all of them later (I did the same in WaveTools.h)

adimako:
@cekees I have not merged this one yet. I was about to but then I thought I would add a solitary wave theory to WaveTools. 
I used Eleni's formulation and I created the unit tests by copying / pasting the formulation in OpenFOAM source code, where we know that it works as we ahve tested it. It passed all tests, so unless we there are any objections, I say we merge, if the tests pass

cekees:
+1

PR #498, opened 2017-02-17
added support for IFISS mesh domains

cekees:
Create a domain from an IFISS mesh with 

```
domain = proteus.Domain.MeshQuadDomain_IFISS('grid_data_step_4')
```


cekees:
@alistairbntl this is a PR onto your branch, so you can merge it whenever you're ready. @bootland just pushed his fix to the IFISS mesh support.

PR #504, opened 2017-04-08
Fix test gmsh

cekees:


cekees:
@tridelat I just removed the use of the time command in the shell because the buildbot images don't have 'time' installed. I don't think the time is used by the tests, so I'm jst leaving it out for now. Could do it from Python if necessary to avoid system dependencies.

PR #506, opened 2017-04-11
Adds a new TestTools class for the air-water-vv repository.

alistairbntl:


alistairbntl:
Cool, this will need some work as we move forward.  For the time being, @lmaurel should have me as a reviewer on his air-water-vv stuff so I can track the progress and bugs in these test classes.

PR #526, opened 2017-05-10
Issue gauge domain

yuxianglin:
Hi:

I modify the Gauge.py in https://github.com/erdc-cm/proteus/blob/master/proteus/Gauges.py#L266,the first argument input is changed to (nearest_node_distance, comm.rank).

In Domain.py https://github.com/erdc-cm/proteus/blob/master/proteus/Domain.py#L1049, I modify the h to (h[0],h[1],h[2])

Yuxiang

### DIFF
--- a/proteus/Domain.py
+++ b/proteus/Domain.py
@@ -1000,7 +1000,7 @@ class PiecewiseLinearComplexDomain(D_base):
                     pf.write('\n')
                 if self.facetHoles:
                     for hN, h in enumerate(self.facetHoles[fN]):
-                        pf.write(`hN+1`+' %f %f %f\n' % h)
+                        pf.write(`hN+1`+' %f %f %f\n' % (h[0],h[1],h[2]))
             if self.holes:
                 pf.write('%d\n' % (len(self.holes),))
                 for hN, h in enumerate(self.holes):
--- a/proteus/Gauges.py
+++ b/proteus/Gauges.py
@@ -263,7 +263,7 @@ class Gauges(AV_base):
                                                                     owning_proc), 3)
         else:
             # gauge isn't on any of the elements, just use nearest node
-            global_min_distance, owning_proc = comm.allreduce(nearest_node_distance, op=MPI.MINLOC)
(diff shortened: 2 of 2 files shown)


alistairbntl:
@yuxianglin - thanks for the PR.
It was not clear to me from our discussion on the issue page, but did you have a specific case where the issue in `Domain.py` failed?  If so, can you add a test that illustrates how this PR fixes that bug?

Similarly, could you create a test for the correction to the Gauge function?  It seemed to me you had a working example where the current formulation caused an error.

yuxianglin:
@alistairbntl Oh, I forget the test, I 'll do that shortly...

yuxianglin:
@alistairbntl Sorry for the waiting, the slurm system in our department is broken recently.

I am not quite sure about the test you refer to, I am running make test and it all passes, but the test_gauge.py only test the point gauge inside the domain, should I add point outside at test_gauge.py?

yuxianglin:
About the domain, should I also write a test_domain.py to test if it success?

alistairbntl:
@yuxianglin - The work-flow for patches that fix bugs in the existing code (which includes this pull-request) often go as follows.

1. You are running some simulations and encounter a bug.
2. You identity the bug in the library and fix the code.
3. You create a new test that ensures the bug you encountered has in fact been fixed.  This will prevent future regressions from causing a similar bug to occur in the future.

So, if I understand this PR correctly, you were running some code that had a bug which you then fixed.  Next, you should add a test in the appropriate folder of the `proteus/test` directory that makes sure this bug does not crop up again in the future.  Thus, there should be a test for both the `domain` and the `gauge`.

Are you on Slack?  If you are (find me on the ERDC channel at @alistairbntl), it might be easier to continue this discussion there.  I'd be happy to walk you through setting up the test at some point it you'd like.

cekees:
These changes look good to me. It looks like you triggered a corner case where the gauge point is outside of the domain. We may also want to put in a log message that warns in this case, since most people are probably intending for the gauge to be inside the domain.

PR #530, opened 2017-05-19
make sure Dirichlet values are set for Velocity at atmospheric bdry

cekees:
We need to make sure there's a Dirichlet value for all components of velocity if the boundary is open. To allow tangential components to slip you set the DiffusiveFlux to zero for those components. @tridelat this seems to fix the nonlinear solver convergence issues I've found in the updated wavetanks from @lmaurel and seen by @cdmoore.

### DIFF
--- a/proteus/mprans/BoundaryConditions.py
+++ b/proteus/mprans/BoundaryConditions.py
@@ -169,18 +169,21 @@ class BC_RANS(BC_Base):
             orientation = self._b_or
         self.reset()
         self.p_dirichlet.setConstantBC(0.)
-        if self._b_or[0] == 1. or self._b_or[0] == -1.:
-            self.u_dirichlet.setConstantBC(0.)
-        else:
-            self.u_dirichlet.resetBC()
-        if self._b_or[1] == 1. or self._b_or[1] == -1.:
-            self.v_dirichlet.setConstantBC(0.)
-        else:
-            self.v_dirichlet.resetBC()
-        if self._b_or[2] == 1. or self._b_or[2] == -1.:
-            self.w_dirichlet.setConstantBC(0.)
-        else:
-            self.w_dirichlet.resetBC()
+        self.u_dirichlet.setConstantBC(0.)
+        self.v_dirichlet.setConstantBC(0.)
+        self.w_dirichlet.setConstantBC(0.)
+        # if self._b_or[0] == 1. or self._b_or[0] == -1.:
+        #     self.u_dirichlet.setConstantBC(0.)
+        # else:
+        #     self.u_dirichlet.resetBC()
+        # if self._b_or[1] == 1. or self._b_or[1] == -1.:
+        #     self.v_dirichlet.setConstantBC(0.)
+        # else:
+        #     self.v_dirichlet.resetBC()
+        # if self._b_or[2] == 1. or self._b_or[2] == -1.:
+        #     self.w_dirichlet.setConstantBC(0.)
+        # else:
+        #     self.w_dirichlet.resetBC()
         self.vof_dirichlet.setConstantBC(vof_air)  # air
         self.u_diffusive.setConstantBC(0.)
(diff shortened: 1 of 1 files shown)


tridelat:
@cekees I think it should be the opposite of what is currently in BoundaryConditions.py for dirichlet velocity: `None` for velocity components normal to the boundary, and `0.` for velocity components tangential to the boundary. Old cases used this (pre-BoundaryConditions module), and also used `None` BC dirichlet in some cases for all velocity components. (examples of this can be found in the current air-water-vv master branch  https://github.com/erdc-cm/air-water-vv/commit/798cb080091347ceb5247e03f8efa90d98cb3146).
I think that setting dirichlet BC for all velocity components to `0.` will make the boundary act as a wall, and prevent fluid from escaping.

tridelat:
Okay, I found the commit where this error was introduced: aa59ed469bb40a8fbeb9b9e262678b5638088657
It was definitely an unintended mistake during the cythonisation of BoundaryConditions.
It should be `0` for tangential velocity components and `None` for normal velocity components.

cekees:
What we want is the Dirichlet boundary conditions set for all three components. Then you turn that OFF on a component by adding the zero diffusive flux on top of that. The Dirichlet value needs to stay ON. The reason is that if you get velocities coming into the domain through the boundary, then you need a value of the velocity (usually zero) to be set for the advective flux.

proteus/mprans/BoundaryConditions.py
```diff
@@ -169,18 +169,21 @@ def setAtmosphere(self, orientation=None, vof_air=1.):
...
-            self.v_dirichlet.resetBC()
-        if self._b_or[2] == 1. or self._b_or[2] == -1.:
-            self.w_dirichlet.setConstantBC(0.)
-        else:
-            self.w_dirichlet.resetBC()
+        self.u_dirichlet.setConstantBC(0.)
+        self.v_dirichlet.setConstantBC(0.)
+        self.w_dirichlet.setConstantBC(0.)
+        # if self._b_or[0] == 1. or self._b_or[0] == -1.:
+        #     self.u_dirichlet.setConstantBC(0.)
```
cekees: I was in a hurry and didn't have time to fix these commented out sections, but these are where you want to turn on  the diffusive flux if the (and leave the Dirichlet conditions alone).

cekees: That is, turn on the diffusive flux for the normal components.

cekees:
Just to clarify further, the weak Dirichlet boundary conditions are "upwinded" so setting a Dirichlet condition on the velocity does not necessarily turn the boundary into a wall. If you set the diffusive flux in the normal direction to zero then the Dirichlet condition is not enforced on outflow, which is what you want for allowing fluid out.  The tangential component values will be enforced by penalty, which stabilizes the flow through the boundary.

tridelat:
I see, thank you for the clarification. 
This was not clear as a lot the old air-water-vv test cases (that do not use the BoundaryConditions module) are using dirichlet conditions only on some velocity components for atmosphere boundary condition (and the others are set to `None`) or sometimes do not use any dirichlet condition on the velocity (such as here: https://github.com/erdc-cm/air-water-vv/blob/master/2d/linear_waves_flat/linear_waves_flat_01_GAZ/twp_navier_stokes_p.py , where there is no dirichlet condition for velocity on the `"top"` boundary, and both diffusive components of the velocity are set to `0.`), hence the confusion.

cekees:
@adimako there are similar considerations for downstream outflow boundaries, the only difference being that pressure is hydrostatic instead of constant. It looks like mprans.BoundaryConditions currently doesn't do anything in the body of `setHydrostatic...` and that the weir problems in @lmaurel's branch are using some parameters that in the function interface for master. Is there another branch somewhere with new work on hydrostatic outflow boundaries?

cekees:
will go ahead and merge the atmospheric fix but raised #531 for downstream hydrostatic boundaries

PR #532, opened 2017-05-23
PiecewiseLinearComplexDomain: gmsh to tetgen conversion

tridelat:
This PR is for conversion of 3D gmsh mesh to tetgen mesh with relevant flags using the `use_gmsh` switch on `PiecewiseLinearComplexDomain`.


proteus/MeshTools.py
```diff
@@ -6338,9 +6338,10 @@ def setOutputFiles(self, name='mesh', poly=True, ply=False, asymptote=False, geo
         self.outputFiles['geo'] = gmsh
 
 
-def msh2triangle(fileprefix):
+def msh2triangle(fileprefix, nd):
```
zhang-alvin: Since this function can convert a .msh file into Tetgen readable files, I think the function name should be modified to reflect this capability (msh2simplex?).

tridelat: Thanks for the feedback, I renamed it to `msh2simplex` on commit b28443b

tridelat:
The test fails for the mesh in 3D because the gmsh version is different on the Travis build and my working computer (this affects the number of vertices and elements in the 3D mesh).
The tests are currently comparing the number of entities in the tetgen files (.ele, .node, .face, .edge) to the known (hardcoded) number of entities given by gmsh for a specific geometry. There might be a better way to test it so it works regardless of the version. Otherwise, we can skip those tests for now, or settle down on the gmsh version to use.

adimako:
@tridelat Waiting to fix this one to release v1.3.2

tridelat:
Thanks for fixing the test @cekees ! The relative 2% difference should make it work with most versions. Is it ready to be merged?

adimako:
@cekees @tridelat I am merging this one for v1.3.2

PR #533, opened 2017-05-26
modified hydrostatic outlet BC with "exact" profile

cekees:
- switched to dirichlet and advective/diffusive flux to match the way we modified atmospheric bc
- added pressure profile based on the "exact" integral of the fluid density based on the smoothed Heaviside of the free surface level set. The difference this makes can be seen below, where the non-exact pressure profile induces a strong airflow from the top through the outlet

outletpressureeffect_0 33
outletpressureeffect_4 00


cekees:
Here are the profiles with the new (old) exact pressure profile
exactoutletpressureeffect_0 33
exactoutletpressureeffect_4 00

Issue #535, opened 2017-06-05
Matrix_constructor tests


zhang-alvin:
If I do a `make test` in my local master branch build, I run into the following error:  

```
___________________________________________________________ ERROR collecting proteus/tests/matrix_constructor/test_laplace_reference_C0P1_3D.py ____________________________________________________________
test_laplace_reference_C0P1_3D.py:12: in <module>
    import laplace_template_C0P1_3D as tp_3D
import_modules/laplace_template_C0P1_3D.py:67: in <module>
    ns = NumericalSolution.NS_base(so,[p],[n],so.sList,ip.opts)
../../../linux2/lib/python2.7/site-packages/proteus-1.3.1-py2.7-linux-x86_64.egg/proteus/NumericalSolution.py:314: in __init__
    check_call("mv {0:s}.1.neigh {0:s}.neigh".format(fileprefix), shell=True)
../../../linux2/lib/python2.7/subprocess.py:540: in check_call
    raise CalledProcessError(retcode, cmd)
E   CalledProcessError: Command 'mv reference_triangle_3d.1.neigh reference_triangle_3d.neigh' returned non-zero exit status 1
```
This is caused by a recent change in [NumericalSolution.py:](https://github.com/erdc-cm/proteus/commit/c2f6e09030d799a2a77ddf23ebe6d529fe535fb4#diff-3be390be0e2fb5c5bfa10e373dfc5a21R310) by @tridelat .

The Travis CI builds seem to be passing because the matrix constructors tests aren't in the setup.py file. You can check the build [log](https://travis-ci.org/erdc-cm/proteus/builds/237953557) and tests in test/matrix_constructors are not touched. From the commit history, it looks like the tests were created by @alistairbntl.

alistairbntl:
@zhang-alvin - thanks for noticing this.  I did not realize these tests were in the master branch.  I've created a branch that will fix this (although it will require a few iterations).

zhang-alvin:
PR was merged. Closing issue.

PR #537, opened 2017-06-07
Adds proper imports to setup.py.

alistairbntl:
Address issue #535.


alistairbntl:
@tridelat , @cekees - This pull request is designed to address issue #535.   I've added the matrix constructor tests to `setup.py`.  In so doing an error was generated in `NumericalSolution.py` near line 313.  I've added a try/except for the `.neigh` and `.edge` moves because one of the added tests generates a reference simplex that does not produce those file exceptions.  I'm not sure this is the best way to handle this, so please let me know if you have any suggestions.

At this point, however, a new error is popping up in one of the 3D gmesh tests.  (see the failure on the travis build above).  I'm not too familiar with this part of the code or what is triggering the assert failure.  @tridelat do you have any ideas?  Could this be related to the change I've made in `NumericalSolution.py`?  I don't see how but I am very likely missing something.

alistairbntl:
@cekees - think this is good to go?

PR #539, opened 2017-06-14
fixed some bugs in log output

adimako:


adimako:
@cekees @tridelat I will go ahead and merge this, it has not introduced changes in calculations anyway

PR #540, opened 2017-06-15
Adimako/return windows

adimako:


adimako:
@cekees @tridelat I will merge this. Again, minor modifications, not really changing anything, just, making some public arrays available at the constuctor stage

PR #542, opened 2017-06-21
Mwf/second order sw es add 2dm

mfarthin:
@manuel-quezada this has some code for reading in triangular meshes in the "XMS" format. This is the format for the malpasset example, although you'll need to change the suffix to `3dm` from `2dm`.

There is an example routine `test_2dm` in the `tests/poisson_2d` file that shows how to setup a problem using one of these meshes. The `square4x4.3dm` file is a sample mesh in the format. You can also add a `*.bc` file with material types for the mesh nodes. It still needs more testing, but seems to be working correctly


cekees:
I believe we don't need this now--what we need to do is rebase @manuel-quezada's SWE work off master. Closing  but leaving the branch for now.

PR #543, opened 2017-06-21
Add fem dem

cekees:
adds terms to RANS3PF to include multiple solid obstacles represented as individual signed distance functions


cekees:
- adds initial multi-solid level set IBM method in 2D and 3D
- simplifies Dockerfile and enables automated builds on cloud.docker.com by doing only the proteus part here and building on this sequence of incremental containers:
  - https://github.com/erdc-cm/debian_base
  - https://github.com/erdc-cm/jupyterhub
  - https://github.com/erdc-cm/stack_base
- add `make hashdist_package` target to automate building the hashdist binary of proteus, which is now built by the buildbots and uploaded. This is so other repos can now skip building proteus (e.g. https://github.com/erdc-cm/air-water-vv/blob/cekees/general_update_review/.travis.yml#L49)
- add ipympl and ipymesh widgets
- updates the hashstack commit

PR #544, opened 2017-06-24
Mutable Context

cekees:
Starting point for refactoring/extending proteus.Context.


cekees:
@jamijares @arnsong @milad-rakhsha the mutable context is master now in case you want to create contexts with dynamic state (e.g. in notebooks and in coupling to chrono and mosaic).

PR #547, opened 2017-06-26
Adimako/improvements

adimako:
Minor improvements to WaveTools, including a new class for a steady current with ramptime


cekees:
@adimako I changed "solitary wave" to "steady current" in the doc string for the new function and went ahead with some additions for the next PATCH release (1.3.3).

PR #548, opened 2017-06-29
SCOREC 2D Reconstruct Part 1 of 2: 2D Error Estimate and Adapt

zhang-alvin:
This PR expands the error estimation and mesh adaptation capabilities to the 2D domain. Additionally, if the model/mesh is loaded via the core/apf library, the region materials can be specified explicitly with the addition of a `regList` that labels each region in the domain. Two tests are added to check these features:

- test_2DmultiRegion
- test_2DparallelLoadPUMI

These features are a subset of what's in the scorec_reconstructMesh branch and hence constitutes part 1.


zhang-alvin:
That's right.

PR #549, opened 2017-06-29
SCOREC 2D Reconstruct Part 2 of 2: 2D reconstruction

zhang-alvin:
This PR enables the ability to read in 2D serial/parallel proteus mesh data structures, reconstruct the mesh into a SCOREC-library capable mesh, and then perform error estimation and mesh adaptivity while preserving the node, boundary, and region material types. 

To enable this feature, the following lines of code are added to a case file (e.g. `dambreak_Ubbink_medium.py`):
```from proteus.MeshAdaptPUMI import MeshAdaptPUMI
adaptMesh = True
adaptMesh_nSteps = 10
adaptMesh_numIter = 2
#
MeshAdaptMesh=MeshAdaptPUMI.MeshAdaptPUMI(hmax=1.0, hmin=0.001, numIter=adaptMesh_numIter,sfConfig="ERM",logType="on",targetError=100,targetElementCount=8000)
useModel=False
```
Examples will be provided in air-water-vv/scorec_reconstructTest branch 

Limitations:

- Parallel mesh reconstruction is compatible only with partitioning by elements. 
- There are sometimes difficulties with using both the VOF and LS methods concurrently, especially if the mesh is coarsened. Generally, `useOnlyVF=True`must be set. 
- There are some artifacts of a successful attempt at basic serial mesh-to-model classification for SpatialTools objects but has since been rendered useless because of a shift in how the reconstruction process is done. The penalty for a lack of mesh-to-model classification is that to ensure that mesh edges that should be classified on model edges will not disappear during a mesh coarsening step, every exterior mesh entitiy (vertex,edge, etc.) will be considered a model entity. The mesh-to-model classification code will be updated in a future PR. 

To-Do List:

- [x] Resolve the material bug observed with the 4 rank dambreak colagrossi case
- [x] Code clean-up
- [x] Add doc strings


cekees:
Maybe what we should do is that @adimako and I can try to run some air-water-vv tests with this branch, and @tridelat can do a test merge into his branch and try the caisson. What do you think?

proteus/MeshAdaptPUMI/MeshConverter.cpp
```diff
@@ -420,3 +430,687 @@ int MeshAdaptPUMIDrvr::updateMaterialArrays(Mesh& mesh,
...
+  stitchMesh(m);
+  m->acceptChanges();
+}
+
+}
+
+/**************************************************************************/
+
+
+//The following functions are used to facilitate and perform a reconstruction
```
cekees: Most of this looks really close to working in 3D as well. How far off would you estimate the 3D support for this workflow?

proteus/MeshTools.py
```diff
@@ -1548,13 +1548,16 @@ def convertFromPUMI(self, PUMIMesh, faceList,regList, parallel=False, dim=3):
...
           PUMIMesh.constructFromParallelPUMIMesh(self.cmesh,
               self.subdomainMesh.cmesh)
-          for i in range(len(faceList)):
-            for j in range(len(faceList[i])):
-              PUMIMesh.updateMaterialArrays(self.subdomainMesh.cmesh,(dim-1), i+1,
-                  faceList[i][j])
-          for i in range(len(regList)):
-            for j in range(len(regList[i])):
-              PUMIMesh.updateMaterialArrays(self.subdomainMesh.cmesh,dim, i+1, regList[i][j])
+          if(PUMIMesh.isReconstructed()):
```
cekees: Might want to put some informative logEvent statments around this

proteus/Domain.py
```diff
@@ -514,7 +605,7 @@ class PlanarStraightLineGraphDomain(D_base):
     2D domains described by planar straight line graphs.
     """
 
-    def __init__(self, fileprefix=None, vertices=None, segments=None, holes=None, regions=None, vertexFlags=None,
+    def __init__(self, fileprefix=None, vertices=None, segments=None, facets=None, holes=None, regions=None, vertexFlags=None,
```
zhang-alvin: I added the facets object primarily because I wanted to have a general way to treat model boundaries if 3D mesh-to-model classification was implemented. I'll remove it for now.

proteus/MeshAdaptPUMI/MeshConverter.cpp
zhang-alvin: I'll start the issue. I can probably get something working in serial in a few days, but I expect some complications in parallel which may take a week, so let's say in a week and a half I'll aim to get something working in 3D parallel.

zhang-alvin:
The test merge with the mbd_chrono branch and master was able to successfully reconstruct the multi-region floating caisson case. I think this PR is good to go now.

PR #551, opened 2017-07-06
moveMesh: stiffening of elements with quality metrics

tridelat:
To do:
- [ ] add user defined switches/coefficients for quality control
- [x] use scaling factor between relative size of element and distorsion/dilation from initial element shape
ex: scale each metrics with the min/max values of the current mesh (or of the previous time step mesh to limit number of loops), and give a "score" for each element to scale its Young's modulus
- [x] add E (and/or quality metrics) to archive for visualisation purposes


tridelat:
Note: on 689be68, the logic for estimating distorsion and dilation of elements is implemented but the Young's modulus is not yet weighted according to it

cekees:
You have done some testing with the new weighted Young's modulus enabled, correct? You just haven't pushed here yet?

tridelat:
@cekees yes, there has been some modifications,  I will push updates soon

tridelat:
With the current version, after a simulation of 7s with free oscillation of a floating caisson:
Dilation of elements: stiffening based on size only (left), dynamic stiffening using quality metrics (right). (same scale, dark red is maximum dilation)
dilation_diff
Distortion of elements: stiffening based on size only (left), dynamic stiffening using quality metrics (right). (same scale, dark red is maximum distortion)
distortion_diff

tridelat:
Closing for now, as stiffening elements according to their quality does not fully solve the problem of mesh deterioration, and this PR has been up for over a year.
See #652 for another moving mesh approach that controls element volume.

Keeping the branch for now

PR #553, opened 2017-07-07
Adimako/bug 2ndorder

adimako:
- [x] include new functions in RandomNLWaves and change calculation method
- [ ] set-up tests


cekees:
added #601 so we can merge this when you are ready @adimako

adimako:
@cekees I am not sure that we will eventually merge this, depending on the outcome of the tests, but I will definitely run some tests with a much finer mesh and a much longer flume to see what is going on.

PR #534, opened 2017-05-27
Adds a viscosity scaled pressure mass matrix solver.

alistairbntl:
This is the first of several PR that will begin integrating new preconditioning methods in Proteus for Stokes and Navier-Stokes solves.

PR #558, opened 2017-07-10
switched back to using .py,.pyx, and .h files

cekees:
@manuel-quezada I went ahead and did this. I need to do some checks before merging.


cekees:
@Giovanni-Cozzuto-1989 @lmaurel this fixes the issue in master that you discovered from the air-water-vv tests. It was caused by a change to the preconditioners API that I didn't catch in the recent PR #534 from @alistairbntl . We'll eventually need to clean up the LinearSolvers.py and set up some backward compatible names (e.g. for SimpleNavierStokes2D, 3D, etc.--the old Schur complement support)

alistairbntl:
Thanks for pointing this out Chris.

My main comment is that I don't like the names `SimpleNavierStokes3D` or `SimpleNavierStokes2D` for the linear smoother classes.  The reason is that I don't think they describe the method being used (eg. are they using a Schur complement method or something else?).    

If the default method is `ASM`, then I'd rather call the linear smoother something like `petsc_ASM`.  I realize this will upset some backwards compatibility, so we will need to keep these options in place until we reach a version 2.0.  That said, my preferred method for handling this would be to alias `SimpleNavierStokes` with `petsc_ASM` while issuing a warning that the class will be deprecated in future versions.

Also, can I run the `air-water-vv` tests locally so that I can catch a regression like this next time before it merging to `master`?

cekees:
The SimpleNavierStokes preconditioners allow you to use the any of block preconditioners provided by petsc from the command line including the basic Schur complement preconditioners not ASM. They're simple in that they just provide the ability to do block preconditioning on the p/v block structure but don't provide any physics-based scalings or custom approximations like you need for PCD.  I think the name is fine but not really the issue.  The issue is that if you're going to change a name or an API (as with additional setUp() arguments) the change shouldn't break existing tests/applications. air-water-vv isn't set up to run automatically on PR's yet but you can run it manually.

adimako:
@alistair, we will shortly have tests scripts in air-water-vv that will allow you to run the tests with a very short simulation time, just to see that the cases are not broken

alistairbntl:
@cekees - I agree completely with your comment regarding breaking existing tests/applications.

I guess my broader point is that I am not sure the `SimpleNavierStokes` classes should be used in the `air-water-vv` module, at least as currently configured.  Specifically, I don't see any `petsc` option configurations in the module that use a Schur complement approach to preconditioning.  Rather, all I see are configurations for global `lu` or `asm` solvers.

As a result, if the user does not specify a petsc input file, `SimpleNavierStokes` sets up a fieldsplit preconditioner that uses `ilu` on the sub-blocks (which I doubt a user would ever want in practice).  If, however, a user specifies one of the petsc options files in the `air-water-vv` repository, it seems petsc overwrites the field-split settings and applies the global `lu` or `asm` approach.

If `SimpleNavierStokes` is designed to provide a general class for petsc input files with field-splitting configurations then I'm okay with that.  It is less clear to me that we should be using the same class when we plan to override it with an entirely different approach.

One option would be to have a class that is catered exclusively for handling a general petsc option defined preconditioners and another for petsc option defined preconditioners that use a field-split approach.

alistair:
@adimako you probably after a different alistair :)

PR #559, opened 2017-07-11
Fix comparision with None

nehaljwani:
If the value of `flag_elementBoundaries` is not `None`, python throws the following
error:
```
ValueError: The truth value of an array with more than one element is
ambiguous. Use a.any() or a.all()
```

### DIFF
--- a/proteus/PostProcessingTools.py
+++ b/proteus/PostProcessingTools.py
@@ -218,7 +218,7 @@ class VelocityPostProcessingAlgorithmBase:
 #        TODO:
 #         get rid of temporary creation, just add a scalar argument like daxpy
         flux = -1.0*self.vt.ebq_global[('totalFlux',ci)]
-        if flag_elementBoundaries == None:
+        if flag_elementBoundaries is None:
             cfemIntegrals.updateExteriorElementBoundaryFlux(self.vt.mesh.exteriorElementBoundariesArray,
                                                             self.vt.mesh.elementBoundaryElementsArray,
                                                             self.vt.mesh.elementBoundaryLocalElementBoundariesArray,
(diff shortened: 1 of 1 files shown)


nehaljwani:
This probably happens when numpy 1.13.0 is used.

https://github.com/numpy/numpy/blob/master/doc/release/1.13.0-notes.rst#futurewarning-to-changed-behavior

The travis builds probably use 1.11.0

cekees:
I think you're right. We've been gradually fixes these comparisons.

alistairbntl:
I think this looks go to go.  Does anybody have any final concerns?

zhang-alvin:
A cursory grep of the code shows many instances of such comparisons (`== None`). Is there a cheap, preemptive way of identifying all such statements that should be changed to avoid similar errors in the future? On the extreme side, is there a reason we shouldn't just change all such comparisons to `is None`? `==` compares values whereas `is` compares addresses (see  [stack overflow](https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python)).

alistairbntl:
Good point @zhang-alvin .  I don't see any reason why we can't safely change all instances of `== None ` to `is None`.  It really shouldn't be that much work either.  
I think the job of switching over all the instances is beyond the scope of this patch, so let's create an issue!  That should guarantee it gets done!  ;-)

alistairbntl:
Perhaps I was being too cavalier in assuming all instances of `== None` should be switched to `is None`.  I think it would be appropriate in nearly all cases, but let's continue this discussion over in the issue tracker #562.

nehaljwani:
Should I update this PR capturing all instances or wait for another patch superseding this?

alistairbntl:
I think we should just go ahead merge this and submit another patch with all the instances at a later time.

zhang-alvin:
Agreed, we can go ahead with the merge.

PR #561, opened 2017-07-12
Adds a Chebyshev semi-iteration solver.

alistairbntl:
- Add a new class to create and apply a Chebyshev semi-iteration
     - Introduces several tests to verify resonable output
     - Adds a new function for turning dense numpy arrays
       into petsc4py matrices (really only useful for testing)


alistairbntl:
@bootland - Would you mind reviewing my documentation for this class and letting me know if you think I've misrepresented or missed anything important.   @manuel-quezada - I figured this may be of interest to you given your work with the lump mass-matrices.  I'm not 100% how this method will work in that situation but it might be worth trying it out.  Thanks!

proteus/LinearSolvers.py
```diff
@@ -13,6 +13,7 @@
 import Quadrature
 from petsc4py import PETSc as p4pyPETSc
 from math import *
+import math
```
cekees: Would be nice to get rid of the from math import * on the previous line.

proteus/LinearSolvers.py
```diff
@@ -2465,3 +2466,142 @@ def _createOperator(self,coeff,matrixDict,A):
...
+    for problems including the mass matrix).  This makes the method 
+    particularly well suited for solving sub-problems that arise in 
+    more complicated block preconditioners such as the Schur 
+    complement, provided one has an aprior bound on the systems 
+    eigenvalues (see Wathen 1987 - Realisitc eigenvalue bounds for 
+    Galerkin mass matrix).
+
+    When implementing this method it is important you first
+    have tight aprior bounds on the eigenvalues (denoted here as 
+    alpha and beta). This can be a challenge but, the references
```
cekees: but, -> ,but

proteus/LinearSolvers.py
```diff
@@ -2465,3 +2466,142 @@ def _createOperator(self,coeff,matrixDict,A):
                                                                           self.model.csrColumnOffsets[(ci,cj)],
                                                                           matrixDict[ci][cj],
                                                                           A)
+
+
+class IterativeMethod:
```
cekees: What's the difference in an iterative method and a LinearSolver or  a Preconditioner? Just don't want to proliferate distinct class hierarchies without a good reason.

alistairbntl:
This is now part of #598.

Issue #562, opened 2017-07-13
Replace all instances of `== None` with `is None`


alistairbntl:
To address https://github.com/numpy/numpy/blob/master/doc/release/1.13.0-notes.rst#futurewarning-to-changed-behavior , most instances of `==None` should be replaced with `is None`.
I believe that the current behavior of `==None` checks whether the object is simply `None`, but in the future `==None` will do an element wise comparison.  I think this means `[None, None, None] == None` would return `true` while `[None, None, None] is None` would return `false`.  I want to be sure this is what's happening, but if so I don't think there is any harm in changing `==None` to `is None`.

zhang-alvin:
Good to point out the future behavior. The current hashstack `d3d27601f2de3c4538f11972f68973b142bdbc92` seems to be using a version of numpy that doesn't have the elementwise comparison implementation (version 1.12). 
```
>>> import numpy
>>> A = numpy.array([None,None,None])
>>> A == None
__main__:1: FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.
False
>>> A[0] == None
True
```

If that's the case, then none of the current instances of `== None` should rely on this (future) elementwise comparison feature and would instead explicitly define a loop to do elementwise comparison. Long story short, I think we'll be okay in doing such a change; I'll try out the substitution in a new branch and see if Travis gives any issues.

cekees:
I agree. My understanding is that the way we've been using it is equivalent to `is None` or `is not None` so a global replacement of these shouldn't change any behavior. I certainly can't think of any case where I've done anything using numpy arrays that would have elements that are None. If one of you two wants to take that on please go ahead. Thanks for looking into it.

zhang-alvin:
Closing issue, now that #563 is merged in. Will reopen if similar issue arises in the future.

PR #563, opened 2017-07-13
equalsNone2isNone_patch

zhang-alvin:
Associated with Issue #562.

This PR is a patch that substitutes all instances of `== None` and `!= None` with `is None` and `is not None`, respectively. This is done to hopefully avoid any further issues like that seen in PR #559.


alistairbntl:
@zhang-alvin - I think this looks good.  
Has this been tested this against anything in the `air-water-vv` repo?  If not I think it would be a good sanity check before merging.  @lmaurel - perhaps you could run this branch against some of you tests to verify that everything is kosher (keep in mind this does not include patch #564.)

zhang-alvin:
Good call on sanity checks @alistairbntl. I ran both the dambreak_Colagrossi case and compared it with the results acquired prior to the modifications associated with this PR (same mesh):

dambreakcolagrossi_beforeandafterpr

I'll try a slightly more complicated case like the floating bar case just in case.

lmaurel:
@zhang-alvin @alistairbntl I ran my tests on this branch and the results are fine for me.

alistairbntl:
@lmaurel - Thanks for running that.  I think this is good to go!

PR #564, opened 2017-07-13
Fixes a fatal error for some logEvent calls.

alistairbntl:
-  This patch corrects a fatal error that occurs in a ksp-log call when the initial solution of a nonlinear
      iteration is in fact the true solution.


alistairbntl:
@cekees , @lmaurel - This PR should fix the bug @lmaurel reported earlier on the Proteus channel.  Basically it looks like the initial time step had the exact solution so the residual was zero.  I hadn't accounted for this when reporting the relative residual on the initial linear solve.

lmaurel:
The bug is fixed for me, it works properly.

alistairbntl:
@cekees - are you okay with this patch?  I think it would be helpful for the Wallingford group to merge this soon.  I'm not sure if that could mess up undoing patch #559 if we need to.

PR #565, opened 2017-07-14
Fixed bug in NonlinearSolver

Giovanni-Cozzuto-1989:
@cekees , @lmaurel , @alistairbntl 
This patch is for correcting an error occurring when linearSolver.setResTol in NonLinearSolvers is called.
One argument (maxIts) was missing.

### DIFF
--- a/proteus/LinearSolvers.py
+++ b/proteus/LinearSolvers.py
@@ -418,7 +418,7 @@ class KSP_petsc4py(LinearSolver):
         self._setMatOperators()
         self.ksp.setOperators(self.petsc_L,self.petsc_L)
 
-        self.setResTol(rtol_r,atol_r,maxIts)
+        self.setResTol(rtol_r,atol_r)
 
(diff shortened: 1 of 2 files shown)


alistairbntl:
@Giovanni-Cozzuto-1989 - There I go changing the API again.  Hold off on merging this, I think I'd better change this function to keep it consistent with the old API.  Can you reference the simulation that originally generated this error.

lmaurel:
@alistairbntl I had this error when I ran the wavesloshing case (https://github.com/erdc-cm/air-water-vv/tree/lmaurel/general_update_review/2d/benchmarks/wavesloshing) in the branch lmaurel/general_update_review.

alistairbntl:
@Giovanni-Cozzuto-1989, @lmaurel - Thank you for brining this to my attention.  I've changed the pull request because the original API should be preserved rather than adding an additional (and it turns out unnecessary) additional input argument into the `setResTol` function.

PR #567, opened 2017-07-14
Include SolitaryWave in pxd

adimako:


adimako:
@yuxianglin try this patch if it fixes things

PR #570, opened 2017-07-17
Cekees/mbd/chrono

cekees:
@adimako this contains an important fix to the Darcy-Forcheimer terms in 2D.
@zhang-alvin it changes the implementation of the "isotropic" mesh adaption option as explained in the comment in SizeField.cpp. 
@tridelat this is on top of your mbd/chrono branch because I was trying to run the adaptivity on the caisson but I ended up addressing these issues first. The diffs won't look so bad if you merge origin/master into your branch.


proteus/mprans/RANS2P2D.h
```diff
@@ -2031,7 +2031,7 @@ namespace proteus
 						w,
 						q_velocity_sge[eN_k_nSpace+0],
 						q_velocity_sge[eN_k_nSpace+1],
-						q_velocity_sge[eN_k_nSpace+2],
+						q_velocity_sge[eN_k_nSpace+1],//cek hack, should not be used
```
cekees: @adimako the velocity in 2D only has two components so you have to use this hack (or we could change the interface to only use u and v).

proteus/mprans/RANS2P2D.h
```diff
@@ -862,11 +862,11 @@ namespace proteus
       /* duc_du = u/(uc+1.0e-12); */
       /* duc_dv = v/(uc+1.0e-12); */
       //semi-implicit quadratic term
-      uc = sqrt(uStar*uStar+vStar*vStar*+wStar*wStar); 
+      uc = sqrt(uStar*uStar+vStar*vStar);
```
cekees: @adimako this was generating nan's becase wStar was indexing one past the end of the velocity array.

proteus/MeshAdaptPUMI/SizeField.cpp
```diff
@@ -17,13 +17,9 @@ static void SmoothField(apf::Field* f);
    thickness of refinement near the interface */
 static double isotropicFormula(double phi, double hmin, double hmax)
 {
-  static double const epsilon = 0.02;
-  phi = fabs(phi);
   double size;
-  if (fabs(phi) < epsilon)
+  if (fabs(phi) < 5.0*hmin)
```
cekees: @zhang-alvin This makes a band of 10 elements of diamter hmin around the zero level set. See below for calling the smoothing so it doesn't reduce the size of the band. With this one could hope to take 5 steps at CFL=1 before the free surface leaves the refinement zone, but we have to be a little more careful than that. I'm using CFL=0.33 and an interface epsFact of 1.5 so the transition region is 3 elements wide.

cekees:
@adimako I just  pushed an updated stack for this branch that builds zoltan with explicitly forced parmetis support, since the scorec tools rely on zoltan being build that way. I don't know if this will help with your hydra issues or not.

PR #572, opened 2017-07-18
BLD: Remove zoltan dependency for ease of compilation

zhang-alvin:
Zoltan, while a useful tool, is not very necessary for the adaptivity functionality. To make compilations less error prone across multiple platforms, we should move towards a stack without zoltan.


cekees:
@zhang-alvin we could just make this an option to the adaptivity. It looks like @tridelat fix on #585 probably fixes the issue related to zoltan, so we can also just close this.

adimako:
@cekees Agreed. I 'd rather have zoltan in

PR #577, opened 2017-07-22
Update adapt

cekees:
- [x] add user-defined anisotropic size field description
- [x] add non-trivial size field computations from the proteus models (e.g. mimic the 'isotropic' but from within MCorr.h
- [x]  verify that all options are working in serial and parallel
etc.
- [x] get gauges to work with the adaptivity process


zhang-alvin:
The `isotropicProteus` additions look alright for now. Anisotropy can be specified through a tensor field called a `size-frame`. It won't be too complicated to add that link between Proteus and PUMI.

zhang-alvin:
Regarding the placement of the "adapt" routine in the `NumericalSolution.py` made by an older [PR](https://github.com/erdc-cm/proteus/pull/570/files#diff-3be390be0e2fb5c5bfa10e373dfc5a21R1384 ) (#570) is generally fine. 

The original placement of the function was motivated by a desire to know exactly when adaptation was going to occur and this is tied to the `n0.adaptMesh_nSteps` variable. For example, if I wanted to adapt every 0.1s and there was a fixed time stepping scheme of 0.05s then I know `n0.adaptMesh_nSteps = 2`. My understanding is that the current placement with a time-stepping scheme of `minAdaptiveModelStep` would yield adaptations in less predictable intervals since it may take for example 4-5 adaptive time steps to iterate from `t_n` to `t_{n+1}` which for the same configuration as above, would already yield 2 adaptations in a single time step (as opposed to 2). 

The newer placement is closer to what is desired in the end goal I think with error-driven adaptivity, whereas the older placement was easier to control the timing of adaptivity. It'll probably be useful to just have an option to switch between the two.

cekees:
I understand what you were trying to do. I like the idea of having a dt_adapt, but the way to do that correctly is to check the the current time and compare to a saved time when the mesh was last adapted. The outer loop over the tnList is primarily for archiving purposes so it frequently changes depending on what is needed for a given call. For example I might run the simulation 20 times saving only 10 time steps and then turn around and save 2000 time steps to get a nice animation of the dynamics. You would want those two use cases to result in the same or nearly the same sequence of calls to the adaption. Likewise in many of the benchmarks where we need to start running with adaptivity, we don't actually want to archive the solution much at all (i.e. tnList=[0,T]) because all the validation is done with the Gauge output.

So, to be clear: Don't add an option to switch between the two placements. The old placement is wrong and doesn't actually achieve what you want.  If you want to add the ability to adapt at i*dt_h_adapt then add an option to default_so.py like `dt_h_adapt`  and use it with the current placement.

zhang-alvin:
Ah, I see. That makes sense and sounds simple enough to implement.

cekees:
@zhang-alvin the 3D reconstruction isn't working for me. Can you point me to a working example? It looks to me like the code to handle 3D is missing in NumericalSolution.py.

zhang-alvin:
I was running the marin case in my tests. Attached is the modified case file for the marin case from air-water-vv commit 5f910f88239dd7fafd0bdfcb28521ddf60d924c3. The setup is admittedly very fragile as there are some inbuilt assumptions. For example, since the dambreak cases used a region material flag of 1, there is an assumption that region material types should start with 1. If you index the region material types starting with 0, you will run into the error message that you posted. I can remove all assumptions altogether at the cost of another loop.

[marin.txt](https://github.com/erdc-cm/proteus/files/1227125/marin.txt)

cekees:
@zhang-alvin please be sure to pull this branch and start working/branching off it as it contains a lot of manually resolved conflicts.

Issue #554, opened 2017-07-07
3D Mesh Reconstruction for SCOREC tools-based Adaptivity


zhang-alvin:
This is a reminder/task list for the next iteration of the reconstruction feature which will be for 3D domains. 
The user will submit as input a proteus mesh and will be able to add several lines to the input file to turn on mesh adaptivity. The input mesh should not have any physical/material interior boundaries.

Components: 
[ ] 3D serial
[ ] 3D parallel

zhang-alvin:
#589 was merged, will close this issue.

PR #579, opened 2017-08-02
3D Mesh Reconstruction

zhang-alvin:
3D mesh reconstruction with naive exterior entity classification. This PR aims to address #554 and allows for the reconstruction of a proteus mesh into a SCOREC-readable mesh data structure. Exterior entities are classified as model entities for simplicity and material types are preserved during this reconstruction.

This is functional in both serial and parallel. The marin case was run as a test example in parallel up until T=1s for a moderate discretization (~30k elements). The following is an isovolume comparison at that time for a baseline case and one using the same mesh but one that is reconstructed and passed back into Proteus.
The isovolume is colored by node type to indicate the preservation of the node material types through reconstruction.
reconstruction_vs_noreconstruction


zhang-alvin:
Duplicated in #589 with comparison to the `update_adapt` branch as opposed to `master` branch.

Issue #580, opened 2017-08-02
Round-off makes time data indistinguishable when running in large time values


adimako:
I was running a wave tank for a long time to get the statistics (800 sec) and I noticed that in the probe file, the some of the time stepping is indistiguishable, e.g. 700.11 will be reapeated for 5-10 times.

This is because the time step is ~0.001 and the last decimal place is truncated. I was able to write a short script to receive the times from the log file, but I think it is better if we allow for more decimal places. 

I guess I will have to edit this https://github.com/erdc-cm/proteus/blob/master/proteus/Gauges.py#L726. 

Any suggestions before I go ahead for a PR would be welcome

PR #581, opened 2017-08-02
Include more decimal places

adimako:
In response to issue https://github.com/erdc-cm/proteus/issues/580


adimako:
@cekees @alistairbntl either of you will do

PR #584, opened 2017-08-07
Wall treatment for turbulent flows

Giovanni-Cozzuto-1989:
PR created for implementing wall functions in BoundaryConditions module:
- [x] Create class for calculating log-law BC based on GLOBAL velocity
- [x] Preliminary use of dirichlet boundaries to impose directly the velocity (simplified method)
- [x] Implement functions for extracting velocity from the model
- [x] Implement the auxiliaryVariables in SpatiaTools for multiple wall functions definition
- [x] Extract directly k values from the model rather than calculate them
- [ ] Check and fix the issue when BC point and extraction point belong to different partitions
- [ ] Use diffusive boundaries to impose directly the shear stress
- [x] Fix k production at near-wall region
- [x] Fix instability for ke model (kw model seems to be stable)


proteus/mprans/BoundaryConditions.py
```diff
@@ -584,10 +641,20 @@ def ux_dirichlet(x, t):
...
+            self.k_advective.resetBC()
+            self.k_diffusive.resetBC()
+        if dissipationInflow is not None:
+            self.dissipation_dirichlet.uOfXT = inlet_dissipation_dirichlet
+            self.dissipation_advective.resetBC()
+            self.dissipation_diffusive.resetBC()     
 
 
-# FOLLOWING BOUNDARY CONDITION IS UNTESTED #
+# FOLLOWING BOUNDARY CONDITIO IS UNTESTED #
```
cekees: Is CONDITIO Italian:)

Giovanni-Cozzuto-1989: It's latin actually ahah :)

Giovanni-Cozzuto-1989: Just one thing Chris. This morning I had a meeting with @adimako . 
We agreed on the best configuration to use at this stage. I'll pull soon it.
I'd like if you can still give a look deeper in the extraction module for velocity (and kappa as well). Randomly I get nan value when I use a big number of processors.

adimako:
- [ ] Clean up and put documentation
- [ ] Include tests
- [ ] Upload test case in air-water-vv

adimako:
I would say that these are needed for the PR to be completed, but that will take some time more. We discussed for Giovanni to switch to the 3-phase solver from end of this week and kinda do these tasks on the side.

adimako:
If we want to merge first and do these later, it is also fine

Giovanni-Cozzuto-1989:
I cleaned up deleting the profiling log statements and adding few descriptions for how each function works.

Giovanni-Cozzuto-1989:
@tridelat Could you check the extraction functions (for velocity and kappa respectively)?
It should be something tricky with them I'm still missing.
When I run a simulation with more than 12 processors sometimes it works whilst in the other cases I get "nan".

Giovanni-Cozzuto-1989:
@adimako I pushed the new test for BC.
I added inlet, oulet and wall functions tests.

Issue #566, opened 2017-07-14
Scorec compilation fails in hydra


adimako:
We get a "cannot find aph.h" error. I think @tridelat has solved this already? Any advice from @zhang-alvin would be appreciated as well.

zhang-alvin:
Can you post the branch commit and the stack commit? You would normally find the `apf.h` header file in the `proteus/$PROTEUS_ARCH/include` (e.g. proteus/linux2/include) directory. You can check if the missing files are there first.

adimako:
@zhang-alvin. It is not there. I think it is a stack problem. This file is not at the PROTEUS_ARCH and not in my stack as well. Where is it coming from?

adimako:
It has always been like this. We have been patching this by removing the PUMI libs in setup.py, but now that your stuff is in, we can't keep doing this

zhang-alvin:
The header files should be coming from the `scorec` package in the PROTEUS_ARCH.yaml file:
https://github.com/hashdist/hashstack/blob/stable/proteus/examples/proteus.linux2.yaml#L92

Once the `scorec` package is compiled, the header files are automatically copied to the PROTEUS_ARCH/include directory.

adimako:
https://github.com/HRWallingford/hashstack/blob/stable/hydra/gcc6.1.0/examples/proteus.centos.yaml

This is what we are using. The only difference is the `use:host-mpi `command. In the pkgs file, we use a different version of scorec 

```
sources:
  - url: https://github.com/tjcorona/core.git
key: git:374c1f0640d2a96242e7a5fd01fd27bd5a935b11
`````

zhang-alvin:
The version of scorec shouldn't be an issue. @tridelat mentioned that the `use:host-mpi` was more of a hack to avoid compiling the `scorec` tools altogether as that redirects hashdist to unpackage the `host-mpi.yaml` file instead. I suggest removing that line. That should prompt the actual compilation of the `scorec` package and very likely, you'll run into an error regarding `zoltan` when you're compiling Proteus.

adimako:
@zhang-alvin Correct, but what do I do with the zoltan error

zhang-alvin:
If @cekees's update to the build stack does not work, I'd go ahead and remove the zoltan dependency. It's useful as a third-party graph partitioner, but the adaptivity tools should be able to function without it.

adimako:
[...]

PR #585, opened 2017-08-08
[FIX] config file for installing SCOREC on centos

tridelat:
In response to #566 ,
I can install MeshAdapt with this config file on hydra. 
@adimako @Giovanni-Cozzuto-1989 @lmaurel , could you also try installing MeshAdapt with this config file?


cekees:
This looks right to me. You could remove the commented out zoltan since you pull it in using the hashdist package info.

cekees:
Merging this. If it doesn't fix the builds for @adimako and @Giovanni-Cozzuto-1989 on hydra then it seems more likely to be an environment thing.

Issue #591, opened 2017-08-09
Mesh Adaptivity incompatible with Gauges


zhang-alvin:
For a 2D serial dambreak_Ubbink case from air-water-vv, I receive the following error while reinitializing the data structures post-adapt:

```Traceback (most recent call last):
  File "/home/alvin/Projects/proteus/linux2/bin/parun", line 4, in <module>
    __import__('pkg_resources').run_script('proteus==1.3.3', 'parun')
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 739, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1494, in run_script
    exec(code, namespace, namespace)
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/EGG-INFO/scripts/parun", line 484, in <module>
    runName + '_run_prof')
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/proteus/Profiling.py", line 190, in __call__
    return func(*func_args, **func_kwargs)
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/proteus/NumericalSolution.py", line 1403, in calculateSolution
    self.PUMI_adaptMesh()
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/proteus/NumericalSolution.py", line 1058, in PUMI_adaptMesh
    self.PUMI2Proteus(mesh)
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/proteus/NumericalSolution.py", line 746, in PUMI2Proteus
    self.auxiliaryVariables[model.name]= [av.attachModel(model,self.ar[index]) for av in n.auxiliaryVariables]
  File "/home/alvin/Projects/proteus/linux2/lib/python2.7/site-packages/proteus-1.3.3-py2.7-linux-x86_64.egg/proteus/Gauges.py", line 666, in attachModel
    self.identifyMeasuredQuantities()
[...]

PR #592, opened 2017-08-17
Gauge compatibility fix with adaptivity

zhang-alvin:
Addresses #591 
This enables the resetting of the point/line gauges post adaptivity. Since the gauges have a somewhat complex reference tree/graph by the solution is computed, the strategy used here is to not reinitialize the gauges altogether but instead to take the existing gauges and removing the incompatible entities such as the `nearest_node` entries.

The commit history/diff is a little wonky because I forgot to branch from `update_adapt` after merging in the anisotropic capabilities.


zhang-alvin:
I ran the dambreak_ubbink benchmark from air-water-vv commit 5f910f88239dd7fafd0bdfcb28521ddf60d924c3. 

Ran 4 cases to T=1s:

Cases with adaptivity adapt every 10 steps.

A) Baseline_noadapt, h=0.01
B) Baseline_noadapt, h=0.005
C) WithAdapt, h=0.01 and then uniformly adapts to h=0.005 
D) WithAdapt, h=0.005 and then adapts such that the mesh is approximately the same except for fixing some elements 

Each case seems to have a different transient behavior. For the adaptivity cases, that might be caused by the reinitialization of the level-set post-adapt.

Pressure gauge values for A) B) C) in single plot. 
pressure_baselines_v_adapt

Pressure gauge values for A) B) D) in single plot
pressure_baselines_v_adaptiso

I also ran a multigauge/line gauge air-water-vv/nonlinearWaves  numerical tank case. The simulation proceeds post-adaptivity, but haven't looked into how to interpret/post-process gauge outputs.

proteus/MeshAdaptPUMI/SizeField.cpp
```diff
@@ -59,8 +59,9 @@ int MeshAdaptPUMIDrvr::calculateSizeField()
     changes in the smoothed size are less than 50% of hmin.
    */
   double err_h_max=hmax;
-  int its=0;
-  while (err_h_max > 0.5*hmin && its < 200)
+  //int its=0;
+  //while (err_h_max > 0.5*hmin && its < 200)
+  for(int its=0;its<200;its++)
     {
```
cekees: Did you meant to switch this to a forced 200 iterations?

proteus/NumericalSolution.py
```diff
@@ -750,7 +751,38 @@ def PUMI2Proteus(self,mesh):
...
+                        av.pointGaugeVecs[ptGaugeIdx].destroy();  
+                        av.pointGaugeMats[ptGaugeIdx].destroy();  
+                        av.dofsVecs[ptGaugeIdx].destroy();
+                      av.pointGaugeVecs = []
+                      av.pointGaugeMats = []
+                      av.dofsVecs = []
+                      av.field_ids=[]
+                      av.isGaugeOwner=False
+                  except AttributeError:
+                    pass
```
cekees: might be good to log a message on these exceptions unless these are expected frequently and not a problem.

proteus/Gauges.py
```diff
@@ -310,13 +311,17 @@ def initOutputWriter(self):
...
         numLocalQuantities = sum([len(self.measuredQuantities[field]) for field in self.fields])
         self.localQuantitiesBuf = np.zeros(numLocalQuantities)
-
         if self.gaugeComm.rank != 0:
             self.globalQuantitiesBuf = None
             self.globalQuantitiesCounts = None
         else:
-            self.fileName = os.path.join(Profiling.logDir, self.fileName)
-            self.file = open(self.fileName, 'w')
+            if self.adapted:
```
cekees: Do we need to close some files somewhere before opening the same logfile adaption? Or are they being closed properly already?

proteus/MeshAdaptPUMI/MeshConverter.cpp
```diff
@@ -941,6 +941,18 @@ int MeshAdaptPUMIDrvr::reconstructFromProteus(Mesh& mesh, Mesh& globalMesh,int h
...
+  int regStartMaterial = 100;
+  int rID = 0;
+  while(ent = m->iterate(entIter)){
+    if(mesh.elementMaterialTypes[rID] < regStartMaterial)
+      regStartMaterial = mesh.elementMaterialTypes[rID];
+    rID++;
+  }
+  m->end(entIter);
+
+
```
cekees: This fixes the issue with usage of material type==0, right?

proteus/NumericalSolution.py
cekees: I'm I'm following correctly, you'r preserving the av objects throughout the simulation and making the `attachVariables` function aware of whether adaption occured. You remove all of the existing gauges so `attachModels` will find the new locations.

proteus/MeshAdaptPUMI/SizeField.cpp
zhang-alvin: Yes, I was running into issues where a collective call was not completing. My understanding is that some ranks seemed to be stuck in the while loop but other ranks have left, which is an issue since the smoothing process is a collective operation built on the following abstract operator: 

https://scorec.rpi.edu/~dibanez/core/classapf_1_1CavityOp.html#a5d04176d38cd0e0a54e9c6e9a1a6f45a

I'm coding up a mesh gradation function currently, so this section  of code will likely be unnecessary soon regardless.

proteus/NumericalSolution.py
zhang-alvin: I expect there to be a better way to do this but the reason the try/except structure is there is because there are auxiliary variables (AV) not related to gauges that to my knowledge don't need to be modified. For example, there is an AV associated with relaxation zones for some problems and they do not have a `isAdapted` variable in their class so they will trigger the `AttributeError` exception. So they are expected frequently.

zhang-alvin: Your summary for what I'm trying to do with the gauges post-adapt sounds right. There is a web of references to the gauge class instance that I didn't want to modify without deeper knowledge of the issue. And since all that should change with the gauges is largely the nearest nodes to perform the computations, I aimed to have that recalculated for pointGauges.

proteus/Gauges.py
zhang-alvin: Good point. It wasn't throwing an error but I'll need to add a close file statement somewhere before the resetting of the gauges.

proteus/MeshAdaptPUMI/MeshConverter.cpp
zhang-alvin: Yes, now it shouldn't matter what you set as the region flag in the domain construction. Although oddly, it's only with the floating bar case that I've seen that the region flag set is not equivalent to the elementMaterialType.

zhang-alvin:
Below is a comparison of what happens when I run the Colagrossi case with a refinement of 32 and 64 followed by a case where I start off with a refinement 64 case and then allow for adaptivity around the interface while coarsening everywhere else to that of refinement 32.  The snapshots are at T=0.501s. It looks like the adaptivity coupled with the transfer of the level-set to the new mesh introduces perturbations to the interface that grow as the simulation proceeds.

Refinement 32: 
refinement32_t0p501

Refinement 64
refinement64_t0p501

Adaptive Case
adapt64to64_t0p501_withcoarsen

cekees:
Is this in parallel? I don't think that large hump should be there on the uniform mesh simulations.

zhang-alvin:
This is in parallel, and again I seem to have forgotten to turn off the conservative flux `twp_navier_stokes_n.py`. I'll rerun the case.

zhang-alvin:
https://github.com/erdc-cm/proteus/pull/592/commits/9e5e8557dd84179ca1bd9bc3bbe5be5734e2dd88 should address #593 

Running the Colagrossi case in parallel with element partitioning with no conservative flux and with nLayersOverlap > 0 did not remove the observed hump. Running it in parallel with node based partitioning yielded no hump.

https://github.com/erdc-cm/proteus/pull/592/commits/0b2e5c7606efaf6fa5533d948b3562cb3c3cd632
adds a mesh grading function that ensures that the size-ratio of any adjacent vertices is less than a hard-coded grading factor of 1.5 and is such that the smaller value in the ratio is never the one that is modified.

Below is the before and after of the mesh in the initial step with a close-up of the post adapt mesh and then a comparison of the size-fields before and after grading (blue is smaller values). The minimum value for each size field is in fact the same. 
aftergrading_adapt
aftergrading_adapt_zoom
beforegrading_aftergrading

cekees:
@zhang-alvin Are the hump issues resolved with the fix to the boundary conditions? Would be good to get the adaptivity branches cleaned up and merge into master this week.

zhang-alvin:
@cekees, the fix hasn't been implemented yet. I'll be implementing it today and will hopefully have it done by tomorrow. I'll aim to have it mergeable by the developer call.

zhang-alvin:
Meshes:
a) Refinement around interface of 64, 32 everywhere else
b) Refinement 32 mesh
c) Refinement 64 mesh

a) vs b)
adapted32to64_noadapt32
a) vs c)
adapted32to64_noadapt64
Zoomed in mesh comparison of a) vs b)
adapted32to64_noadapt32_meshcomparison

PR #594, opened 2017-08-26
SpatialTools: fix tank3D and region issues

tridelat:
Fixes issues with tank3D regions, and the start region flag when assembling domain.

### DIFF
--- a/proteus/mprans/SpatialTools.py
+++ b/proteus/mprans/SpatialTools.py
@@ -387,7 +387,8 @@ class Tank3D(ShapeRANS):
                              'x+': 4,
                              'y-': 5,
                              'z+': 6,
-                             'sponge': 7}
+                             'sponge': 7,
+                             'wall': 8}
         self.b_or = np.array([[0.,  0., -1.],
                               [-1., 0.,  0.],
                               [0.,  1.,  0.],
@@ -406,14 +407,16 @@ class Tank3D(ShapeRANS):
                                        b_or=self.b_or, b_i=4),
                    'z+': self.BC_class(shape=self, name='z+',
                                        b_or=self.b_or, b_i=5),
-                   'sponge': self.BC_class(shape=self, name='sponge')}
+                   'sponge': self.BC_class(shape=self, name='sponge'),
+                   'wall': self.BC_class(shape=self, name='wall')}
         self.BC_list = [self.BC['z-'],
                         self.BC['x-'],
                         self.BC['y+'],
                         self.BC['x+'],
                         self.BC['y+'],
                         self.BC['z+'],
-                        self.BC['sponge']]
+                        self.BC['sponge'],
+                        self.BC['wall']]
         # self.BC = BCContainer(self.BC_dict)
         for i in range(6):
             self.BC_list[i].setTank()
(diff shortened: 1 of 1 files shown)


Giovanni-Cozzuto-1989:
@tridelat I think everything is fine. Just two (maybe silly) questions:
1) why do you need the number of volumes equals to the number of facets? Shouldn't be 5 for this case?
2) why don't you have also along z-axis the possibility to create sponge sides (z- znd z+)?

proteus/mprans/SpatialTools.py
```diff
@@ -471,161 +473,159 @@ def setDimensions(self, dim):
...
+        v_i += 4
+        r_i += 1
+        nb_sponge += 1
+        f_i = len(facets)
+        # x+
+        vertices += [[x1+x_p, y0, z0], [x1+x_p, y1, z0],
+                     [x1+x_p, y0, z1], [x1+x_p, y1, z1]]
+        vertexFlags += [bt['z-'], bt['z-'],
+                        bt['z+'], bt['z+']]
+        volumes[0][0] += [f_i]  # add to volume
```
Giovanni-Cozzuto-1989: Could you explain me how volumes works?

proteus/mprans/SpatialTools.py
```diff
@@ -471,161 +473,159 @@ def setDimensions(self, dim):
...
+                facetFlags += [bt['wall']]
+            else:
+                facetFlags += [bt['y+']]
+            volumes[0][0][-1] = f_i+3 
+            volumes += [[[f_i+i for i in range(6)]]]
+        v_i += 4
+        r_i += 1
+        nb_sponge += 1
+        f_i = len(facets)
+        # x-
```
Giovanni-Cozzuto-1989: Don't you need also to generate z- and z+ cubes?

tridelat: It is assumed that the gravity acceleration is aligned with the Z axis, like for most simulations, otherwise it adds an extra layer of complications

proteus/mprans/SpatialTools.py
tridelat: Volumes are only foto build a gmsh mesh, they are a facet loop basically

proteus/mprans/SpatialTools.py
adimako: - Having the gravity in the z direction only can be restrictive as some drawing will have the vertical in x or y.
- But I understand the logic. In this case you can put an error message if gravity is detected in x or y (non zero values). This will prevent users from doing something stupid

tridelat: I think it makes sense to leave it like this, and if someone wants to use it in another direction, they can rotate the shape by using `my_tank.rotate(angle, axis)` that is available in SpatialTools for any shape.

adimako: Yes but if they do not realise that they need to align the gravity with z and e.g. give g=[9.81,0,0] will it still work?

tridelat: I mean that they can rotate the tank to be aligned with the gravity, it can be added to the docstring as a warning, but I think it is find as it is, as the `rotate` function should be quite straightforward to us

adimako: I guess it should be fine for now, but let's keep an eye on the feedback we get from other users

PR #595, opened 2017-08-29
Ibm dev

cekees:
Just putting this up so we can start moving toward master.


proteus/mprans/RANS3PF.py
```diff
@@ -879,12 +933,44 @@ def preStep(self, t, firstStep=False):
...
+
+        self.phi_s[:]=1e10
+        self.phisField=np.ones(self.model.q[('u',0)].shape,'d')*1e10
+        if self.granular_sdf_Calc is not None:
+            print ("updating",self.nParticles," particles...")
+            for i in range(self.nParticles):
+                for j in range(self.mesh.nodeArray.shape[0]):
+                    vel = self.granular_vel_Calc(self.mesh.nodeArray[j,:],i)
+                    sdf,sdNormals = self.granular_sdf_Calc(self.mesh.nodeArray[j,:],i)
+                    sdf+=vel* self.model.dt_last
```
cekees: @Milad-Rakhsha I'm not sure about line 945. If the velocity of the particle is `vel`, then I suspect you mean `np.dot(np.array(vel),np.array(sdNormals))` and `-=` insteadf of `+=` if your intention is to translate the particle.

Milad-Rakhsha: Right, This is totally incorrect and should not have been committed, I was just testing to see if we could simply advect the sdf with the velocity field to have a better estimation of  for mesh adaptivity around phi_s for moving objects,
but I suspect that doing something like `np.dot(np.array(vel),np.array(sdNormals))` would not work too
I guess something like 
vel = self.granular_vel_Calc(self.mesh.nodeArray[j,:],i)
sdf,sdNormals = self.granular_sdf_Calc(self.mesh.nodeArray[j,:]-vel* self.model.dt_last,i)
`
should do the job

PR #575, opened 2017-07-22
Alistairbntl/schur mprans

alistairbntl:
Adds a Two-Phase PCD preconditioner for MPRANS2D module with some parallel support.

PR #598, opened 2017-09-15
Cekees/pstab

alistairbntl:


proteus/LinearAlgebraTools.py
```diff
@@ -231,10 +315,38 @@ def scatter_reverse_add(self):
 class ParVec_petsc4py(p4pyPETSc.Vec):
     """
     Parallel vector using petsc4py's wrappers for PETSc
+    ARB Question - I think this WIP has been completed.  Do you agree ckees?
```
cekees: Agreed. This works for mixed spaces now too.

proteus/LinearSolvers.py
```diff
@@ -719,18 +750,409 @@ def __init__(self, linear_smoother, pde_type='general_saddle_point'):
...
+                petsc_a[LAT.ParInfo_petsc4py.proteus2petsc_subdomain[i],LAT.ParInfo_petsc4py.proteus2petsc_subdomain[j]] = nzval[k]
+        for i in range(transport.dim):
+            for j,k in zip(colind_petsc[rowptr_petsc[i]:rowptr_petsc[i+1]],range(rowptr_petsc[i],rowptr_petsc[i+1])):
+                nzval_petsc[k] = petsc_a[i,j]
+        return SparseMat(transport.dim,transport.dim,nzval_petsc.shape[0], nzval_petsc, colind_petsc, rowptr_petsc)
+
+    def initializeTwoPhaseCp_rho(self):
+        """Initialize a two phase scaled advection operator Cp.
+
+<<<<<<< HEAD
```
cekees: this appears to have unresolved conflicts in it.

cekees:
Looks like 7e10d80 was not a complete merge (it committed unresolved conflicts).

proteus/LinearAlgebraTools.py
```diff
@@ -194,6 +202,82 @@ def petsc_create_diagonal_inv_matrix(sparse_petsc):
     diag_inv.setDiagonal(1./sparse_petsc.getDiagonal())
     return diag_inv
 
+# TODO - figure out whether this function is needed
```
cekees: This looks like a duplicate of the function of the same name below.

proteus/LinearAlgebraTools.py
cekees: The comment should be removed.

proteus/LinearAlgebraTools.py
```diff
@@ -301,15 +413,85 @@ def scatter_reverse_add(self):
...
+        self.proteus2petsc_subdomain = None
+        self.petsc2proteus_subdomain = None
+        self.nzval_proteus2petsc = None
+        self.dim = None
+        self.mixed = False
+
+    def print_info(cls):
+        import Comm
+        comm = Comm.get()
+        print 'comm.rank() = ' + `comm.rank()` + ' par_bs = ' + `cls.par_bs`
```
cekees: Use logEvent, not print. If you want to log on all processors then use `-A, --logAllProcesses`. There are several reasons 1) this version of print goes away in Python 3, 2) if somebody accidentally uses this on a large multi-processor job it can easily fill up their stdio buffer, 3) we want the log to contain all relevant information

proteus/LinearAlgebraTools.py
```diff
@@ -301,15 +413,85 @@ def scatter_reverse_add(self):
...
+        A map from the local unknown to the global unknown.
+    blockVecType : str
+    pde : :class:`proteus.Transport.OneLevelTransport`
+        The Transport class defining the problem.
+    par_nc : int
+    par_Nc : int
+    proteus_jacobian : :class:`proteus.superluWrappers.SparseMatrix`
+        Jacobian generated by Transport class's initializeJacobian.
+    nzval_proteus2petsc : :class:`numpy.ndarray`
+        ?
```
cekees: ?

proteus/LinearAlgebraTools.py
```diff
@@ -345,6 +527,7 @@ def __init__(self,ghosted_csr_mat=None,par_bs=None,par_n=None,par_N=None,par_ngh
         comm = Comm.get()
         logEvent("ParMat_petsc4py comm.rank= %s blockSize = %s par_n= %s par_N=%s par_nghost=%s par_jacobian.getSizes()= %s "
                  % (comm.rank(),self.blockSize,par_n,par_N,par_nghost,self.getSizes()))
+#        import pdb ; pdb.set_trace()
```
cekees: remove

proteus/LinearAlgebraTools.py
```diff
@@ -724,8 +983,166 @@ def apply(self,A,x,y):
...
+        self.A10 = A10
+        self._create_Sp()
+        self._options = p4pyPETSc.Options()
+        self._create_constant_nullspace()
+        
+        self.kspSp = p4pyPETSc.KSP().create()
+        self.kspSp.setOperators(self.Sp,self.Sp)
+        self.kspSp.setOptionsPrefix('innerSp_solve_')
+        self.kspSp.setFromOptions()
+        self.Sp.setNullSpace(self.const_null_space)
```
cekees: Can you explain the route to setting self.const_null_space? Which object has the responsibility of deciding if it has one or not?

proteus/LinearAlgebraTools.py
```diff
@@ -724,8 +983,166 @@ def apply(self,A,x,y):
...
+    def _create_Sp(self):
+        self.A00_inv = petsc_create_diagonal_inv_matrix(self.A00)
+        A00_invBt = self.A00_inv.matMult(self.A01)
+        self.Sp = self.A10.matMult(A00_invBt)
+#        self.Sp.scale(-1.)
+        self.Sp.aypx(-1.,self.A11)
+    
+
+class TwoPhase_PCDInv_shell(InvOperatorShell):
+    r""" Shell class for the two-phase PCD preconditioner.  The
```
cekees: Thanks for putting the time into the doc string. It's good to use a consistent notation for these preconditioners.

proteus/LinearAlgebraTools.py
```diff
@@ -724,8 +983,166 @@ def apply(self,A,x,y):
...
+        self.Qp_visc = Qp_visc
+        self.Qp_dens = Qp_dens
+        self.Ap_rho = Ap_rho
+        self.Np_rho = Np_rho
+        self.alpha = alpha
+        self.delta_t = delta_t
+
+        self._options = p4pyPETSc.Options()
+        self._create_constant_nullspace()
+        # TODO ARB - Need to implement the Chebyshev semi-iteration for
```
cekees: Didn't you already do that?

proteus/LinearAlgebraTools.py
```diff
@@ -983,3 +1400,141 @@ def norm(self,y,type):
...
+            if self.save_iterations:
+                self.iteration_results.append(self.x_k.getArray().copy())
+        
+# The implementation that is commented out here was adopted from
+# the 1996 text Iterative Solution Methods by Owe Axelsson starting
+# on page 179.  As currently written, this algorithm requires
+# inputing the inverse diagonally preconditioned matrix A and b.  I'm
+# not sure this is the best approach, but I'd like to leave this code
+# in place for now in case it is useful in the future.
+
```
cekees: Makes sense.

proteus/LinearSolvers.py
```diff
@@ -411,7 +417,11 @@ def __init__(self,L,par_L,
         self.petsc_L = par_L
         self.csr_rep_local = self.petsc_L.csr_rep_local
         self.csr_rep = self.petsc_L.csr_rep
-        self.bdyNullSpace = bdyNullSpace
+        # HACK FOR CLOSED BDY PROBLEM - NEED TO CHANGE
```
cekees: We may need to put some more thought into this in general. The reason I put a link to the pde in the LinearOperator object is to pass information about the pde, specifically information on the operator's null space. This is going to be an issue for all the new work on projection schemes as well because we'll often be solving the pressure increment/correction poisson problem with a null space.

proteus/LinearSolvers.py
```diff
@@ -523,37 +533,43 @@ def _defineNullSpaceVec(self,par_b):
...
-        par_n = par_b.getLocalSize()
-        par_nghosts = par_b.nghosts
-        subdomain2global = par_b.subdomain2global
-        proteus2petsc_subdomain = par_b.proteus2petsc_subdomain
-        petsc2proteus_subdomain = par_b.petsc2proteus_subdomain
-        # information for pressure uknowns
-        pSpace = self.par_L.pde.u[0].femSpace
-        pressure_offsets = pSpace.dofMap.dof_offsets_subdomain_owned
-        n_DOF_pressure = pressure_offsets[rank+1] - pressure_offsets[rank]
-#        assert par_n == n_DOF_pressure
```
cekees: remove or uncomment?

proteus/LinearSolvers.py
```diff
@@ -523,37 +533,43 @@ def _defineNullSpaceVec(self,par_b):
...
-                                           par_n,
-                                           par_N,
-                                           par_nghosts,
-                                           subdomain2global,
-                                           proteus2petsc_subdomain=proteus2petsc_subdomain,
-                                           petsc2proteus_subdomain=petsc2proteus_subdomain)
-        null_space_basis.scatter_forward_insert()
-        self.tmp_array[:] = self.tmp_array[petsc2proteus_subdomain]
-        self.global_null_space = [null_space_basis]
+#         par_N = par_b.getSize()
```
cekees: is this whole section a duplicate of the above?

proteus/LinearSolvers.py
```diff
@@ -719,18 +750,433 @@ def __init__(self, linear_smoother, pde_type='general_saddle_point'):
         pde_type :  str 
             Currently supports Stokes and navierStokes
         """
+        import proteus
+        # TODO - find away to get rid of this import statement
```
cekees: Why is it necessary? Can't see it being used. Is this a testing issue?

proteus/MeshTools.py
```diff
@@ -4837,6 +4837,37 @@ def refine(self):
         self.meshList.append(TriangularMesh())
         childrenDict = self.meshList[-1].refine(self.meshList[-2])
         self.elementChildren.append(childrenDict)
+    # def generateFromExistingCoarseMesh(self,mesh0,refinementLevels,nLayersOfOverlap=1,
```
cekees: this is implemented above. Remove this?

proteus/NumericalSolution.py
```diff
@@ -270,13 +270,15 @@ def __init__(self,so,pList,nList,sList,opts,simFlagsList=None):
                     tmesh = TriangleTools.TriangleBaseMesh(baseFlags=n.triangleOptions,
                                                         nbase=1,
                                                         verbose=10)
+                    # ARB Before merge figure out the best way to deal with problem (eg in parallel
```
cekees: This is fixed by the simple change below, right? Just read the mesh on all processors so it is consisent?  We could raise an issue on refactoring the 2D mesh generation with triangle, gmsh, etc. and do this all better in the refactor.

cekees: /cc @zhang-alvin

proteus/ShockCapturing.py
```diff
@@ -44,6 +44,7 @@ def initializeElementQuadrature(self,mesh,t,cq):
                         cq[('da_sge',ci,ck,cj)]=cq[('da',ci,ck,cj)]
 
     def calculateNumericalDiffusion(self,q):
+        import pdb ; pdb.set_trace()
```
cekees: remove?

proteus/SplitOperator.py
```diff
@@ -150,6 +150,57 @@ def setFromOptions(self,soOptions):
 
 Sequential_FixedStep = SO_base
 
+class Sequential_tnList(SO_base):
```
cekees: How does this differ from Sequential_FixedStep_Simple

proteus/SubgridError.py
```diff
@@ -1506,8 +1506,6 @@ def __init__(self,coefficients,nd):
             coefficients.stencil[3].add(3)
     def calculateSubgridError(self,q):
         if self.nd == 2:
-        #    import pdb
```
cekees: remove?

proteus/Transport.py
```diff
@@ -6577,11 +6569,12 @@ def initialize(self,
                         for i in range(transport.dim):
                             for j,k in zip(colind_petsc[rowptr_petsc[i]:rowptr_petsc[i+1]],range(rowptr_petsc[i],rowptr_petsc[i+1])):
                                 nzval_petsc[k] = petsc_a[i,j]
-                        assert((nzval_petsc[nzval_proteus2petsc] == nzval).all())
-                        assert((nzval[nzval_petsc2proteus] == nzval_petsc).all())
-                        assert (proteus_a[petsc2proteus_subdomain,:][:,petsc2proteus_subdomain] == petsc_a).all()
-                        assert((proteus_a == petsc_a[proteus2petsc_subdomain,:][:,proteus2petsc_subdomain]).all())
-                        petsc_a = np.arange(transport.dim**2).reshape(transport.dim,transport.dim)
+                                # TODO - ARB: these asserts need to be changed to reflect new petsc_a and proteus_a structures
```
cekees: fix or raise issue and do for next release?

proteus/mprans/RANS2P.py
```diff
@@ -391,6 +400,56 @@ def attachModels(self,modelList):
         if self.epsFact_solid is None:
             self.epsFact_solid = numpy.ones(self.model.mesh.elementMaterialTypes.max()+1)
         assert len(self.epsFact_solid) > self.model.mesh.elementMaterialTypes.max(), "epsFact_solid  array is not large  enough for the materials  in this mesh; length must be greater  than largest  material type ID"
+        arbTest = True
```
cekees: This should be refactored to allow passing in a function for phi and vof. We do this in a lot of other places. @manuel-quezada needs this for RANS3P as well.

proteus/transportCoefficients.c
```diff
@@ -2425,6 +2425,7 @@ void TwophaseNavierStokes_LS_SO_2D_Evaluate(const int nPoints,
   for (k=0;k<nPoints;k++)
     {
       /** \todo optimize coefficients */
+      printf("Here1\n");
```
cekees: remove

proteus/transportCoefficients.c
```diff
@@ -2461,7 +2462,7 @@ void TwophaseNavierStokes_LS_SO_2D_Evaluate(const int nPoints,
       
       dmom_v_adv_v[k*2+0]=u[k];
       dmom_v_adv_v[k*2+1]=2.0*v[k];
-
+      printf("Here2\n");
```
cekees: remove

setup.py
```diff
@@ -451,6 +451,10 @@
                     'proteus/tests/hex_cube_3x3.h5']),
                   (os.path.join(proteus_install_path,'tests','linalgebra_tests'),
                    ['proteus/tests/linalgebra_tests/sparse_mat_1.txt']),
+                  (os.path.join(proteus_install_path,'tests','solver_tests','import_modules'),
+                   ['proteus/tests/solver_tests/import_modules/quad_mass_matrix.npy',
+                    'proteus/tests/solver_tests/import_modules/sol_10.npy',
+                    'proteus/tests/solver_tests/import_modules/sol_20_lst.npy']),
```
cekees: in the future we want to track these binary files with git lfs instead of git. @zhang-alvin has done this fairly extensively.

proteus/SplitOperator.py
alistairbntl: As I understand the logic, of `Sequential_FixedStep_Simple`, `self.dt_system` is first set based on the length of the initial time step, which works fine.  
For subsequent time steps, `self.dt_system` starts at the value used in the previous iteration.  Thus, in `Sequential_FixedStep_Simple`'s  `stepExact_system` function (which uses the version written in the `SO_base` base class, see - https://github.com/erdc/proteus/blob/master/proteus/SplitOperator.py#L57 ), if `self.dt_system` is not large enough to reach the next time step, `self.dt_system` is set to step half the distance to `tExact`.  
In the case of wanting to replicate `tnList` exactly, it seems to me that we would want to set `self.dt_system` to give the exact step.  Thus, the difference is that `Sequential_tnList` sets `self.dt_system` to arrive at the next element of `tnList` in all cases (see - https://github.com/erdc/proteus/blob/c3159a489aae5f15266010b53feb4e355bc26017/proteus/SplitOperator.py#L167 )

alistairbntl:
@cekees - I think I've addressed most of your comments above and yes, I see this PR replacing #575. 

Also, I have not tested the non-conservative for for 3D so I'm a little worried about that part, and it certainly needs a second pair of eyes.

proteus/SplitOperator.py
cekees: Got it. I don't think there's any reason to have both versions, so I removed the definition of Sequential_FixedStep_Simple and made it an alias of Sequential_tnList so it doesn't break anything.

cekees:
I think we should test this on 2D and 3D problems in air-water-vv quite a bit before we merge this. I did find some mistakes in  RANS2P.h in the new NONCONSERVATIVE option and spent a good bit of time making sure RANS2P2D.h and RANS2P.h were synchronized and cleaned up where possible.  @alistairbntl could you check it on all the cases for the 2D results in the paper? I'm running on the 3D marin problem on my laptop, and will try this on some HPC systems.

proteus/mprans/RANS2P.h
```diff
@@ -347,3137 +454,5157 @@ namespace proteus
...
-	   int nDOF_mesh_trial_element,
-	   int nDOF_trial_element,
-	   int nDOF_test_element,
-	   int nQuadraturePoints_elementBoundary>
-  class RANS2P : public RANS2P_base
-  {
-  public:
-    const int nDOF_test_X_trial_element;
-    CompKernelType ck;
+    int nSpace,
```
cekees: Why did you do this?

cekees:
The tests are passing on my machine and the 3D RANS2P code is running with both NONCONSERVATIVE=0 & 1. Just waiting for the travis testing to get straightened out.  I think it has to do with their queue's.

cekees:
Just to make sure I'm following: Now the LevelModel __init__ can take a bdyNullSpace option to flag if the boundary conditions will generate a null space. That option can be set with the flag `boundaryCreatesNullSpace` in the physics  module (defined in default_p). @manuel-quezada you will probably want to use that in PresInc.py.

manuel-quezada:
What I did was to create a flag called fixNullSpace in the Coefficients of
PresInc.py. It is set to False by default to keep the code to work as it
was by default. This is the cheap trick of making the 0-th DOF to be set to
zero strongly. We can do this or let the user provide some desired mean
value pressure and then we do the adjustment accordingly.

alistairbntl:
First, I've re-run everything using the non-conservative form using the latest commit and I haven't come across any issues yet.  I'm going to try  another 2d problem this afternoon, but so far so good.

Second, yes there is now a flag `bdyNullSpace` in the `LevelModel` class which is set in the _p file.  This bdyNullSpace attribute of the LevelModel is then set in KSP_Solver and used if needed.

alistairbntl:
Is PresInc solve a Laplacian type problem?

manuel-quezada:
Yes, it solves the Poisson problem with Neumann BCs so the null space has
the constant vector in it.

cekees:
@alistairbntl I've pushed my merged branch (baa56b7), so you're good to go.

PR #602, opened 2017-09-19
Mquezada/rans3 pf with ev

manuel-quezada:
This branch deals with RANS3PF. It mainly:

- Fixes minor problems with the opt. proj. scheme such as using a 2nd order uStar extrapolation. 
- Changes the formulation to not divide by rho the momentum equation. 
- Adds a calculateResidual and calculateJacobian functions for the entropy viscosity methods. 
- Adds arguments and flags to the python code trying to use by default the original SUPG method. 
- Modifies SimTools to compute the L2 norm of the pressure if required in the batch file. 
- Adds a test in ProjScheme_with_EV to perform a convergence test (for SUPG and EV). 
- Allows the user to provide analytical functions for the density, viscosity and force terms. 
- Allows stop considering the pressure for debugging on the momentum equation via one flag. 
- Allows the possibility of not integrating by parts div(U) during the pressure increment. 

There might be more minor changes in other files required for different things. Please ignore all commits with respect to NCLS. I used an old branch but at the end I checked out the NCLS from master. I will start a new branch for NCLS with EV.


Giovanni-Cozzuto-1989:
@manuel-quezada are you going to merge it to master branch?
Or do you need to do more tests? The thing is I need to work on some sediment benchmark and I'm having some troubles using master branch.
Do you think I should use this one?

cekees:
@Giovanni-Cozzuto-1989 I'll try to get it merged into master before tomorrow. I think it's a good idea to work off of these changes. I also need to merge in the IBM_DEV branch, which has some other relevant changes to RANS3P

Giovanni-Cozzuto-1989:
@cekees , it sounds good! Thanks

PR #610, opened 2017-09-25
Mquezada/ncls with ev

manuel-quezada:
This pull request is meant for the NCLS+re-distancing with EV. The idea is to solve the transport equation with a penalization term (by Coupez) to re-distance as the solution is transported. Afterwards, it checks if the Eikonal equation must be solved to further re-distance the level set. I consider two types of level set function: distance function and a saturated distance function. I add two stabilization methods: one through edge based Entropy Viscosity (EV) and a second via a smoothness indicator. If no extra flags or parameters are passed it uses by default the old implementations (SUPG). This is done to allow other codes that use NCLS to continue working as before without any change. This pull request: 

- adds functions to calculate the Residual and Jacobian with edge based EV methods. 
- allows a non-EV stabilization based on the smoothness of the solution. 
- allows the possibility of using a term (by Coupez) to re-distance as the solution is transported. 
- adds a solver: ExplicitConsistentMassMatrixWithRedistancing, to solve the Eikonal equation as well. 
- adds extra explicit solvers (to be used in VOF and for an L2 Projection during the mass correction). 
- fixes a bug when archiving phi_s in NumericalSolution. 
- makes some changes in SSP within TimeIntegration.py
- makes non-automated tests for pure advection (without re-distancing) in 1D and 2D. 
- makes non-automated tests for pure re-distancing and advection+re-distancing. 
- makes an small change in Advection_DiagonalUpwind_IIPG_exterior to allow for periodic BCs.


manuel-quezada:
@cekees I have thought about the automated tests and I think it might be better not to include all the tests in this pull request but just one representative that would check nothing got broken with new code. Then I would put these tests and others on the repository for the multiphase paper. If you agree, I will remove this pull request and make a new one with such change.

cekees:
What I was thinking we'd do is use `proteus.Context` to set up some options for these cases with defaults that just run a very short/small version of the test and compare to a saved version of the output. That way we catch anything that would  cause the test problem to go "stale", which is a major issue.  Here is a simple example of how to use `proteus.Context`.

Defining options with defaults:
https://github.com/erdc/air-water-vv/blob/master/2d/benchmarks/dambreak_Colagrossi/dambreak_Colagrossi.py

Using the context:
https://github.com/erdc/air-water-vv/blob/master/2d/benchmarks/dambreak_Colagrossi/dambreak_Colagrossi_so.py

PR #615, opened 2017-10-01
Mquezada/vof with ev

manuel-quezada:
This pull request is meant for VOF and NCLS+re-distancing with EV. 

For the **NCLS+EV**, the idea is to solve the transport equation with a penalization term (by Coupez) to re-distance as the solution is transported. Afterwards, it checks if the Eikonal equation must be solved to further re-distance the level set. I consider two types of level set function: distance function and a saturated distance function.

For both, the NCLS and VOF,  I add two stabilization methods: one through edge based Entropy Viscosity (EV) and a second via a smoothness indicator. If no extra flags or parameters are passed it uses by default the old implementations (SUPG). This is done to allow other codes that use NCLS or VOF to continue working as before without any change. 

Here are other specifics of the pull request:

- adds functions to calculate the Residual and Jacobian with edge based EV for NCLS and VOF.
- allows non-EV stabilization based on the smoothness of the solution.
- allows the possibility of using a term (by Coupez) to re-distance the LS as the solution is transported.
- adds a solver: ExplicitConsistentMassMatrixWithRedistancing, to solve the Eikonal equation as well.
- adds extra explicit solvers (to be used in VOF and for an L2 Projection during the mass correction).
- fixes a bug when archiving phi_s in NumericalSolution.
- makes some changes in SSP within TimeIntegration.py
- makes a small change in Advection_DiagonalUpwind_IIPG_exterior to allow for periodic BCs.
- Makes tests for NCLS and VOF with these implementations.


manuel-quezada:
@cekees I didn't do the automated tests but I did tests for NCLS and VOF using the Context in such a way that we can test every feature by changing stuff form the Context.

manuel-quezada:
@cekees I did the automated tests for NCLS and VOF

cekees:
+1

cekees:
This is ready to go. I found some issues with the tests that I've fixed on a newer branch, but will add them later.

adimako:
Is this something worth benchmarking with a wave tank case?

PR #617, opened 2017-10-03
Alistairbntl/test markers

alistairbntl:


alistairbntl:
I've add some notes explaining how to use the test markers as well as suggesting some naming conventions to follow here in the Proteus-Tests wiki: https://github.com/erdc/proteus/wiki/Writing-Proteus-Tests

PR #621, opened 2017-10-06
merged stable/topaz with master

zhang-alvin:
Updating the stable/topaz proteus branch with a merge with master. 

[] Will need to update .hashstack_default when `stable/topaz` is updated in the hashstack repo.


cekees:
I think we should hold off on merging in the stable/hpc branches...

Issue #626, opened 2017-10-16
Systematic solver failure for 3D case on latest master


tridelat:
The latest master causes failures once per time step on the twp model (RANS2P) when running a 3D case, and this considerably increase the computational cost of running a simulation. It happens for a simple air/water 3D cuboidal domain with a still water level (free slip walls and atmophere BC on top).
I use the asm solver and run the case as follows:
`parun [case]_so.py -O [path to pets.options.asm`

It runs fine up to 5d5f1f5bb6e7934af33e7551749f11dd9e709186.
The next 2 commits, a052d8755c80549b7c0a3600cdfc52e7212706ee and 767d4e83ec6f1c34191ba08fdeb67fa0372ea701, the case does not run due to an error in `calculateJacobian()` of RANS2P (189 positional arguments given instead of 190).
From commit d99cf7539a717f67b9db446a938424261622fe33, it runs but gives me the failures at each time step. 

@cekees @alistairbntl @manuel-quezada , any idea what could cause this issue?

alistairbntl:
@tridelat  - I've been working on the Marin problem the past few days and I've noticed a few issues too.  Perhaps most important, I see that the Newton iteration is not converging quadratically.  I suspect this is related to your issue and likely means the Jacobian and residual are inconsistent.
What is your setting for the NONCONSERVATIVE_FORM in the RANS2P Coefficients?

manuel-quezada:
I agree with @alistairbntl that this sounds like an inconsistency between the residual and the Jacobian on RANS2P. Do you have any 2D example in some repository or branch we can run?

manuel-quezada:
These models: RANS2P and RANS3PF have the possibility of considering different terms to model different physics. In other words, we have the chance to add terms to the equations if we consider stuff like turbulence, Darcy's forces, particles (in the case of RANS3PF) and other things. In addition, we can consider different types of boundary conditions. All these "extra" terms are on top of the standard terms for the NS equations. First, we need to try to identify which of these terms is the one that is creating the problem (or the inconsistency between the residual and the Jacobian). I suggest we try to turn off everything that is not part of the standard form of the NS equations. If that works then we can start putting back these terms one by one.

cekees:
[...]

PR #628, opened 2017-10-19
fixed bugs in 3D from the addition of NONCONSERVATIVE option

cekees:
addresses #626

### DIFF
--- a/proteus/mprans/RANS2P.h
+++ b/proteus/mprans/RANS2P.h
@@ -334,6 +334,27 @@ namespace proteus
                                    int* csrColumnOffsets_eb_w_w,                                   
                                    int* elementFlags,
 				   int* boundaryFlags)=0;
+    virtual void calculateVelocityAverage(int nExteriorElementBoundaries_global,
+                                          int* exteriorElementBoundariesArray,
+                                          int nInteriorElementBoundaries_global,
+                                          int* interiorElementBoundariesArray,
+                                          int* elementBoundaryElementsArray,
+                                          int* elementBoundaryLocalElementBoundariesArray,
+                                          double* mesh_dof,
+                                          double* mesh_velocity_dof,
+                                          double MOVING_DOMAIN,//0 or 1
+                                          int* mesh_l2g,
+                                          double* mesh_trial_trace_ref,
+                                          double* mesh_grad_trial_trace_ref,
+                                          double* normal_ref,
+                                          double* boundaryJac_ref,
+                                          int* vel_l2g,
+                                          double* u_dof,
+                                          double* v_dof,
+                                          double* w_dof,
+                                          double* vel_trial_trace_ref,
+                                          double* ebqe_velocity,
+                                          double* velocityAverage)=0;
     virtual void getTwoPhaseAdvectionOperator(double* mesh_trial_ref,
                                               double* mesh_grad_trial_ref,
                                               double* mesh_dof,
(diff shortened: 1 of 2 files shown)


proteus/mprans/RANS2P.h
```diff
@@ -953,165 +953,31 @@ namespace proteus
...
-        mom_u_diff_ten[2] = 1.0 / rho ;
-
-        mom_v_diff_ten[0] = 1.0 / rho ;
-        mom_v_diff_ten[1] = 1.0 / rho ;
-        mom_v_diff_ten[2] = 1.0 / rho ;
 
-        mom_w_diff_ten[0] = 1.0 / rho ;
-        mom_w_diff_ten[1] = 1.0 / rho ;
-        mom_w_diff_ten[2] = 1.0 / rho ;
+	    //u momentum Hamiltonian (advection
```
cekees: @alistairbntl tI think this is the root cause  of what @tridelat found. The additional terms for the nonconservative form still need to be initialized for conservative form.

proteus/mprans/RANS2P.h
```diff
@@ -5240,9 +5103,9 @@ namespace proteus
 
                 if (NONCONSERVATIVE_FORM)
                   {
-                    bc_mom_w_adv_ext[0] -= MOVING_DOMAIN*bc_dmom_w_acc_w_ext*z_ext*xt_ext;
-                    bc_mom_w_adv_ext[1] -= MOVING_DOMAIN*bc_dmom_w_acc_w_ext*z_ext*yt_ext;
-                    bc_mom_w_adv_ext[2] -= MOVING_DOMAIN*bc_dmom_w_acc_w_ext*z_ext*zt_ext;
+                    bc_mom_w_adv_ext[0] -= MOVING_DOMAIN*bc_dmom_w_acc_w_ext*w_ext*xt_ext;
```
cekees: @alistairbntl this needed to be the z component of velocity (w_ext), not the z coordinate.

alistairbntl: Thanks for pointing this out.  These only apply for the moving-mesh problems correct?  I need to familiarize myself a bit more with the methods for the moving-mesh.

proteus/mprans/RANS2P.h
alistairbntl: Indeed, I can see why this would be a problem,  I suppose without initializing these terms, the arrays could be populated with non-zero values.  This would explain the somewhat erratic failures I was seeing.

proteus/mprans/RANS2P2D.h
```diff
@@ -334,6 +334,27 @@ namespace proteus
                                    int* csrColumnOffsets_eb_w_w,                                   
                                    int* elementFlags,
 				   int* boundaryFlags)=0;
+    virtual void calculateVelocityAverage(int nExteriorElementBoundaries_global,
```
alistairbntl: Have you just moved the location of the two-phase operators here?  I think that's a good idea, but I just want to be sure I haven't missed a change to how they're being calculated.  I guess the same question applies for the 3D code above.
I think you've made some changes to the `calculateVelocityAverage` function but I'm not familiar with this function so I don't have anything to add about those changes.

cekees: Yes, I just did that to make it easier to diff RANS2P.h with the prior version and with RANS2P2D.h.

proteus/mprans/RANS2P.h
cekees: Yes, exacyly

cekees: Yes, exacyly

PR #631, opened 2017-10-21
Migrate .h5 files to git lfs and rewrite history

zhang-alvin:
Used git-filter branch to remove non-tracked h5 files from the history, which means that some tests in older commits that rely on comparison files will no longer work. 

Re-added the h5 files with tracking with git-lfs. 

Other files to consider:

```
remote: warning: File notebooks/Presentations/FEMDEM/cylinder.gif is 77.73 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
remote: warning: File notebooks/Presentations/FEMDEM/dambreakib.gif is 64.95 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

This PR will need to be merged manually via command line.


cekees:
Thanks! Do you need help with the conflicts?

zhang-alvin:
@cekees After the first pass-through of migrating files, I saw that there were several fairly large image files that (100+MB total) that I migrated in my second pass-through. 

It would be good to have a look over. I can do the merge at the end of the day.

cekees:
Ok, sounds good.

cekees:
@zhang-alvin you can delete this branch if there are no more commits.

PR #633, opened 2017-10-21
Fix conflicts in #631

cekees:
@zhang-alvin I also made one change to try to get `make test` and the travis's test setup to both work. Otherwise this uses the changes from your PR#631


zhang-alvin:
@cekees Not sure why the test is failing, but I think this is good to go.

Issue #620, opened 2017-10-06
Default Dirichlet Behavior


alistairbntl:
It seems that in some cases, when the Dirichlet boundary flags are not set properly, the Dirichlet boundary conditions can be set across the entire domain rather than just on the boundary.  When this happens, the solution is simply an interpolation of the boundary condition functions.
The default behavior should probably warn the user that the boundary flags have not been set properly.  I will update this issue once I have a simulation which replicates the error.

PR #634, opened 2017-10-22
Reproduce the issue #620

wacyyang:
This pull request is to show the issue #620 . I agree with that maybe it is better to simplify the given Dirichlet b.c. from the user point of view. In this example, the user wants to enforce Dirichlet b.c. everywhere on the boundary and doesn't have to know all flags.


alistairbntl:
@wacyyang - I have one question and one comment with regards to this PR at the moment.

1. It seems you've added a lot of tests.  I haven't looked at them in detail, but do all these tests replicate #620?  If not, can you highlight which of these tests replicate the issue?

2. In order for the tests to be picked with travis, you will need to add the new test module into the `packages` list in the `setup.py` file (located in the home `proteus` directory).  Also, supplementary files used in tests (e.g. .h5 etc.) will need to be added to the `data_files` module.  There are plenty of examples in `setup.py` you can follow.  If you've already done this I apologize, I just didn't see it in the PR.

I'll have a closer look at this once I hear back from you.

wacyyang:
@alistairbntl There is only one test in the folder of */Users/yy/proteus/proteus/tests/yy_practice/ode* which reproduces the issue #620. This test is only used to produce the issue #620, and not a test worth to put into test module. You can run it with command *parun -l 5 -v NS_convergence_so.py -b L2_batch.py* in this folder and uncomment two lines in the file *twp_navier_stokes_p.py* to reproduce the issue. If you are on another branch, you can just checkout this folder to run it. The issue is about the boundary label, and I am worried the change of this part may have a big effect on other parts. We can discuss more details about this issue #620

PR #635, opened 2017-10-22
Composite quadrature rule

wacyyang:
In this pull request, I add 
+ `class CompositeTriangle(Q_base)` which contain all quadrature points and weights in all uniformly subdivided triangles;
+ a file *test_CompQuad.py* to test this quadrature rule. There are two tests in this file, corresponding to the integral over the region below a straight line and inside a circle.


cekees:
Take a look at how I refactored the doc and the functionality to work with our documentation generater (sphinx)  and our testing framework  (py.test). 

It would be good to test something beyond just using the rule. See for example what @tridelat does in test_spatialtools.py. You could compute the order of converengence of the composite quadture rules with respect to hk ans assert that it's close to the expected value, for example. Also you could check that the rules are exact for functions for which they should be exact.

wacyyang:
In this commit, several changes are made:

- Unittest.Testcase is used. And one can use pytest to run it.
- The expectation for test_line and test_circle is the average convergence > 1
- For polynomials the usual quadrature rules should be equal to the composite rule. Low order polynomials like 1, x, xy, xy^2  are tested.

wacyyang:
In this commit, I am trying to add a composite quadrature rule for the reference tetrahedron. The Freudenthal-Bey refinement algorithm is used to get 8 sub-tetrahedron for each tetrahedron. A test file is also added, where 

- Unittest.Testcase is used. And one can use pytest to run it.
- The expectation for test_plane and test_sphere is the average convergence > 1. In face, for those two tests, the convergence is > 1.5
- For polynomials the composite quadrature rule gives the same results as the usual gauss quadrature rules. Low order polynomials like 1, x, xy, xyz are tested.

cekees:
This is looking good. We should move the tests into the root proteus/tests directory. You can do that with `git mv` on any files in the yy_pracitice subdir.  We don't need the  yy_practice subdir.

wacyyang:
I agree with you. In this commit, I have moved the `test_composite_quad` folder to `proteus/proteus/tests/test_composite_quad`

cekees:
This is looking good. Let's do a comparison on the 2D vortex test: `proteus/tests/levelset/vortex2D`

wacyyang:
In this commit, I am beginning to use 2D composite quadrature rule for the 2D vortex test. One comparison about the solutiond at the final time is shown in this figure
screen shot 2017-10-27 at 3 28 04 pm
The 0-level-set becomes much smooth. I will do more tests on this problem. One suggestion is that maybe we can use composite quadrature only around the level set curve by labeling cells before the loop and considering the composite quadrature rule for those particular cells. The cost of current computation does not increase a lot.

PR #637, opened 2017-10-27
Add Shifted Boundary Method for embedded solids

cekees:
This PR from @lnouveau adds an alternative method of evaluating the boundary integrals for Dirichlet boundary conditions enforced via the Nitsche approach.  Proteus uses the Nitsche approach at solid interfaces represented implicitly through a signed distance function (level set).  Currently the boundary integrals are evaluated  using the "smoothed  Dirac" trick for converting boundary integrals  to volume integrals as in many Immersed Boundary Methods. The SBM instead evaluates the boundary integrals along existing mesh element boundaries using interpolation of  the solution field, consistent with the FEM interpolation.  Degrees of freedom inside the shifted/surrogate boundary are "deactivated" by post-processing the residual/Jacobian vector.  The method is described for  Poisson and Stokes here in [Main and Scovazzi 2017](https://www.sciencedirect.com/science/article/pii/S0021999117307799)

# TODO
- [x] Add pressure at surrogate boundary to  complete force calculations
- [x] Verifiy drag/lift on cylinder benchmark
- [ ] Vortex solution on cylinder or stokes solution to verify convergence
- [ ] Add Hessian terms to recover  cubic convergence for quadratic velocity field
- [ ] Add loop over all particle  level sets or use single level set for all particles
- [ ] Transfer to  RANS3PF (3D version)

/cc @Milad-Rakhsha @dtoconn @arnsong


proteus/mprans/RANS3PF2D.h
```diff
@@ -3317,10 +3321,33 @@ namespace proteus
 		    globalResidual[GlobPos_u] += beta_adim*P_tangent[0]*(Gxphi_i*dt1 + Gyphi_i*dt2);
 		    globalResidual[GlobPos_v] += beta_adim*P_tangent[1]*(Gxphi_i*dt1 + Gyphi_i*dt2);
 		   
-
 		  }//i
+
+		//
+		// LEO Forces
+		//
+		double p_ext = 0. ; // !!!! TO CHANGE BY REAL VALUE !!!!
```
cekees: @wacyyang this is where the pressure is not correct

PR #640, opened 2017-11-03
Mquezada/fix minor bugs in sw es

manuel-quezada:
Branch to clean and fix a number of bugs in the SWEs code. This also changes some of the defaults in different tests. This branch produces the results in the paper. Still things to do: 
-Solve a problem with SSP when running the malpasset test. 
-Add local limiting technique.


proteus/mprans/SW2DCV.h
```diff
@@ -54,9 +53,9 @@ namespace proteus
 			    int NNZ, //number on non-zero entries on sparsity pattern
 			    int numDOFs, //number of DOFs
 			    double* lumped_mass_matrix, //lumped mass matrix (as vector)
-			    double* h_lstage, //DOFs of solution at last stage
-			    double* hu_lstage, 
-			    double* hv_lstage, 
+			    double* h_old, //DOFs of solution at last stage
```
cekees: Are this permanent, or just hacks to do FE?

proteus/mprans/SW2DCV.h
```diff
@@ -2039,7 +2028,7 @@ namespace proteus
 	  // CALCULATE EDGE BASED CFL //
 	  //////////////////////////////
 	  double mi = lumped_mass_matrix[i];
-	  edge_based_cfl[i] = fabs(dLowii)/mi;
+	  edge_based_cfl[i] = 2*fabs(dLowii)/mi;
```
cekees: Is this  a safety factor?

proteus/mprans/SW2DCV.h
```diff
@@ -2067,9 +2056,9 @@ namespace proteus
 	  ij=0;
 	  for (int i=0; i<numDOFsPerEqn; i++)
 	    {
-	      double hi = h_dof_lstage[i]; // solution at time tn for the ith DOF
-	      double hui = hu_dof_lstage[i]; 
-	      double hvi = hv_dof_lstage[i]; 
+	      double hi = h_dof_old[i]; // solution at time tn for the ith DOF
+	      double hui = hu_dof_old[i]; 
+	      double hvi = hv_dof_old[i];
```
cekees: Similar to above, is this a permanent change or hack for the RK stages?

proteus/mprans/SW2DCV.h
```diff
@@ -3269,7 +3248,7 @@ namespace proteus
       //FOR FRICTION//
       double n2 = std::pow(mannings,2.);
       double gamma=4./3;
-      double xi=2.;
+      double xi=10.;
```
cekees: should put a comment on what xi is

proteus/mprans/SW2DCV.h
manuel-quezada: They are permanent. The idea is to program calculateResidual as if a FE step is being done and modify the parameter that is being passed depending on which stage of the SSP method is being considered. The modification is done here:

https://github.com/erdc/proteus/blob/mquezada/fix_minor_bugs_in_SWEs/proteus/mprans/SW2DCV.py#L161

proteus/mprans/SW2DCV.h
manuel-quezada: We don't know the exact condition on the time step so that the "low-order" method is stable. The paper we consider for that method states that there is a constant C such that if dt <= C*dLii/mi for all i then the method preserves the monotonicity constraints. Then having or not the 2 factor is irrelevant. At the end of the day we have to recompute the time step if positivity is violated.

proteus/mprans/SW2DCV.h
manuel-quezada: This is permanent. The vectors h_dof_old, etc are updated after every stage of the SSP method.

proteus/mprans/SW2DCV.h
manuel-quezada: The factor xi alleviates the cfl restriction. I need to see the behavior after incorporating the local bounds recently incorporated into the paper.

cekees: Got it. What I mean is that you should put a comment in the code if it's something that is subject to change and not (yet) set by the theory or numerical experiments.

PR #652, opened 2017-11-22
Moving mesh with monitor function

tridelat:
This PR is for working on the moving mesh with monitor function. It is now fully working in series or in parallel with recovered gradients.

Checklist:
- [x] tested in series
- [x] tested in parallel
- [x] fix mesh nodes on exterior boundaries when doing pseudo time steps
- [x] find element containing coords when doing pseudo time step
- [x] gradient recovery at nodes
- [x] add area function to reaction term
- [x] add tests
- [x] (add quality metrics)
- [x] optimise overall code (convert to cython/C++)
- [ ] finish documentation
- [x] allow model to be solved several times on same time step
- [ ] retrieve LS on original mesh when solving for several times with nonlinear solver class

Things to add/improve:
- let nodes slide on exterior boundaries if they are not corner nodes
  -> problem here when nodes are moving on a boundary while imposing dirichlet BC

@wacyyang @cekees


tridelat:
This is what can be obtained right now with a monitor function of:
f = min(1., max(abs(d-0.25)/0.25, 0.1)) with d the distance from [0.5,0.5]

image
left: original mesh
right: mesh after solving model 3 times

tridelat:
Latest commit 20826c3 improved parallelisation by changing how processors communicate (a quick test shows ~7+ times faster parallel code for pseudo-time stepping with ~100 processors), which makes it pretty much ready to be merged in master after adding to the documentation.
There is still some optimisation that could be done mainly with the calls to the `evaluateFunAt...()` that are still in python.
@cekees On specific cases, I get some differences in results when using the python2 to python3 merge of master (I have a separate branch `tridelat/moveMesh_monitor_python2`), I am not sure why (there might be an operation that is not behaving exactly the same in python2 and python3 in those files)

tridelat:
Latest commit 20826c3 improved parallelisation by changing how processors communicate (a quick test shows ~7+ times faster parallel code for pseudo-time stepping with ~100 processors), which makes it pretty much ready to be merged in master after adding to the documentation.
There is still some optimisation that could be done mainly with the calls to the evaluateFunAt...() that are still in python.
@cekees On specific cases, I get some differences in results when using the python2 to python3 merge of master (I have a separate branch tridelat/moveMesh_monitor_python2), I am not sure why (there might be an operation that is not behaving exactly the same in python2 and python3 in those files)

PR #656, opened 2017-11-27
Submodules and test running automatically (master  to master)

Pedrohrw:


.travis.yml
```diff
@@ -47,6 +46,8 @@ install:
 
 script:
 - py.test --boxed -v linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus/tests --cov=linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus
+- ls linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/
+- py.test --boxed -v linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/air-water-vv/Tests --cov=linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus
```
cekees: I think here you just want to do `py.test --boxed -v air-water-vv`. The reason for the complicated path for the main proteus tests is to have it test the installed version of proteus on the installed tests. For air-water-vv I think we can just run proteus on air-water-vv directly. Otherwise we'd have to add __init__.py files to the air-water-vv hierarchy and install it as a python package. 
cc @adimako @alistairbntl

PR #658, opened 2017-12-01
Chrono MBD, AddedMass, and FEA cable coupling expansion

tridelat:
- [x] add custom ChLink
- [x] add AddedMass model
- [x] distance search (node star) for cable nodes
- [x] add test for AddedMass
- [x] add more tests to cover Chrono coupling

- Moved to #714  allow for cable node to slide along other cable


cekees:
@tridelat and @adimako is there some way I can help move this forward?

cekees:
I can't reproduce the error that travis is showing for  the commit but not the pr. There is no difference in the two commits, so I don't understand what's happening.  What are you thinking about for the other entries on the to do list? I think we could get by with setting the cable node location using some external information.

tridelat:
Same here, the test passes locally. I am not sure what is going on but I know that sometimes the FEA module of Chrono has troubles when we run it in parallel (returning nans), and the problem is inconsistent and appears at random (it either happens right away on the first time step or not). 
This happens even if we don't actually parallelise the Chrono code (each processor is running the exact same code, but some of them can return nans). The official parallel version of Chrono also does not work with FEA, so it might be linked to that. I am not sure how we could solve that.

On the to-do list, it would be relatively straightforward to add the custom ChLink (there is some commented out logic that is already there), but not for the sliding nodes (so we might want to drop this one for now). We can add some more tests too.
I am thinking about doing a git rebase from the commit that rewrote entire files (ChRigidBody.pyx and ChMoorings.h) because of line endings, but that might be tedious.

cekees:
@manuel-quezada we haven't changed anything related to RANS3PF on this branch and it is up-to-date with master, but the surface tension tests are failing. Any idea what is happening? The expected values aren't matching up to the tolerances when running locally just those tests, so it seems like a numerical difference. There are some changes to NumericalSolution and NonlinearSolvers on this branch, but I don't see how they would be having an effect.

cekees:
@tridelat there's a bit of a tricky issue with the surface tension test failures that @manuel-quezada and I are looking at. It's triggered by the compilation options on this branch. Since the tests should pass regardless of the optimization options, we should hold off on merging this branch until we understand what is happening.

cekees:
@manuel-quezada I found a similar issue in RDLS.h on the dimension of the gradient reconstruction. Can you double check my change to RDLS.h on this branch?

manuel-quezada:
@cekees I checked the change and it looks correct to me.

cekees:
I raised #715 because I want's able to fix it for this PR.

PR #659, opened 2017-12-05
update buildNodeStarArrays() in MeshTools

wacyyang:
This pull request is to make *nodeStarArray* work correctly for quadrilateral mesh or general mesh containing different shapes of cells. Here the loop of element is used in order to make sure all pairs of nodes are considered.


wacyyang:
Now I have the right indentation. Thanks

wacyyang:
@cekees Sorry about that. I should be more careful. I am using Eclipse and find a way to show all tabs and spaces, and replace all tabs by spaces. So I think I will not make such mistakes any more.

PR #660, opened 2017-12-13
Extend MCorr to use composite quadrature rule

wacyyang:
In this pull request, the original term in the mass correction step 
<a href="https://www.codecogs.com/eqnedit.php?latex=\hat{H}_{n&plus;1}" target="_blank"></a> 
is replaced by 
<a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^n)\bm{V})" target="_blank"></a> so that the composite quadrature rule can play a role here.
Since this is an explicit term now, only the function `getResidual` is changed in the implementation.

Here are some details. 

1.   In order to compare to the original method, a new label `mass_correction_reference` is introduced. 
2. Three new parameters are passed into the c code:  <a href="https://www.codecogs.com/eqnedit.php?latex=\Delta&space;t" target="_blank"></a> from VOF module;  <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)" target="_blank"></a> created in `postStep` of the previous step; the velocity field <a href="https://www.codecogs.com/eqnedit.php?latex=\bm{V}" target="_blank"></a> from VOF module.
3. Those new parameters in c code are called `dt, q_H_old, q_v`.
4. The first term of <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^n)\bm{V})" target="_blank"></a> is added into `r` in `evaluateCoefficients_explicit_comp_quad`; the 2nd term is introduced as `f` computed in the same function. 

Some extension can be done for the next step.

1. Add boundary integral terms. Note that for Vortex2D problem, the boundary integral is zero. 
2. Treat the 2nd term of  <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^n)\bm{V})" target="_blank"></a>  implicitly as <a href="https://www.codecogs.com/eqnedit.php?latex=\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^n&plus;u)\bm{V})" target="_blank"></a>. One has to compute <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n&plus;u)" target="_blank"></a> and update the `calculateJacobian` to include the term <a href="https://www.codecogs.com/eqnedit.php?latex=\int{\delta}_\epsilon(\phi^n)\phi_i\phi_j" target="_blank"></a>
3. More tests will be done since small difference is observed now.


proteus/mprans/MCorr.h
```diff
@@ -69,6 +69,56 @@ namespace proteus
 				   int* exteriorElementBoundariesArray,
 				   int* elementBoundaryElementsArray,
 				   int* elementBoundaryLocalElementBoundariesArray)=0;
+    virtual void calculateResidual_explicit_comp_quad(//element
```
cekees: Don't copy-paste a new residual function for this. Just add a switch (e.g. MOD_MCORR) and add only what you need. Copy-pasting like this is generally a bad idea for a small change, because it  makes it very difficult to see what you've changed.

proteus/mprans/MCorr.h
```diff
@@ -69,6 +69,56 @@ namespace proteus
...
+    				   int* u_l2g,
+    				   double* elementDiameter,
+    				   double* nodeDiametersArray,
+    				   double* u_dof,
+    				   double* q_phi,
+    				   double* q_normal_phi,
+    				   double* ebqe_phi,
+    				   double* ebqe_normal_phi,
+    				   double* q_H,
+				   double* q_H_old,///YY
```
cekees: we want q_phi_old, not q_H_old.

proteus/mprans/MCorr.h
```diff
@@ -510,7 +560,28 @@ namespace proteus
...
+			      const double& H_old,///YY
+				  const double v[nSpace],///YY
+    			      const double& u,
+    			      const double& porosity,
+    			      double& r,
+    			      double& dr,
+				  double f[nSpace],
+				  double df[nSpace])
+    {
+          r = porosity*smoothedHeaviside(epsHeaviside,phi+u) - H_old;
```
cekees: not `H_old`, but rather `porosity*smoothedHeaviside(epsHeaviside, phi_old)`

proteus/mprans/MCorr.h
```diff
@@ -510,7 +560,28 @@ namespace proteus
...
+          dr = porosity*smoothedDirac(epsDirac,phi+u);
+          for (int I=0; I < nSpace; I++)
+          {
+        	  	  f[I] = v[I]*porosity*u;
```
cekees: The explicit flux is `porosity*smoothedHeaviside(epsHeaviside,phi_old)`. Remember that here, `u` is just the correction to  the signed distance (`phi`).

proteus/mprans/MCorr.h
```diff
@@ -510,7 +560,28 @@ namespace proteus
...
+        	  	  df[I] = v[I]*porosity;
+          }
```
cekees: `df` is zero if you  use the explicit flux.

proteus/mprans/MCorr.py
```diff
@@ -656,6 +666,59 @@ def getResidual(self,u,r):
             self.mesh.exteriorElementBoundariesArray,
             self.mesh.elementBoundaryElementsArray,
             self.mesh.elementBoundaryLocalElementBoundariesArray)
+        elif self.coefficients.mass_correction_reference==1:
```
cekees: same as above, don't duplicate the call, just pass in the flag, mass_correciton_reference

proteus/mprans/MCorr.py
```diff
@@ -102,6 +105,12 @@ def attachModels(self,modelList):
             self.ebq_H_vof = modelList[self.VOFModelIndex].ebq[('u',0)]
         else:
             self.ebq_H_vof = None
+
+        #: YY
+        self.q_H_phi_old = numpy.zeros_like(modelList[self.VOFModelIndex].q[('u',0)])
```
cekees: we want `self.q_phi_old = numpy.zeros_like(modelList[self.levelSetModelIndex].q[('u',0)])`

wacyyang:
Thanks for your comments. I'll make those changes now.

proteus/MeshTools.py
```diff
@@ -4904,10 +4901,13 @@ def __init__(self,
                     self.meshList[0].nodeNumbering_subdomain2global.itemset(node,node)
                 for element in range(self.meshList[0].nElements_global):
                     self.meshList[0].elementNumbering_subdomain2global.itemset(element,element)
+		
+		self.meshList[0].buildNodeStarArrays()
```
cekees: Can you check the whitespace and tabbing in this file?. There should be no tabs and indent levels should be 4 spaces. Several lines here don't look like the right amount of indentation. 

BTW, These changes are not related to the mcorr changes, but that's OK. Next time just make sure you branch off of origin/master.

proteus/mprans/MCorr.h
cekees: Can you delete this declaration now that the changes are applied only to calculateResidual?

proteus/mprans/MCorr.h
```diff
@@ -510,7 +561,33 @@ namespace proteus
       r = porosity*smoothedHeaviside(epsHeaviside,phi+u) - H;
       dr = porosity*smoothedDirac(epsDirac,phi+u);
     }
-
+    inline
```
cekees: There should just be one evaluateCoefficients function

proteus/mprans/MCorr.h
```diff
@@ -969,7 +1046,406 @@ namespace proteus
 	    }//kb
 	}//ebNE
     }
+    inline void calculateElementResidual_explicit_comp_quad(//element
```
cekees: likewise one calculateElementResidual function

proteus/mprans/MCorr.py
```diff
@@ -102,6 +106,12 @@ def attachModels(self,modelList):
             self.ebq_H_vof = modelList[self.VOFModelIndex].ebq[('u',0)]
         else:
             self.ebq_H_vof = None
+
+        #: YY
```
cekees: You don't need to mark your contributions with YY. Git tracks that for you. See `git blame`.

proteus/mprans/cMCorr.pyx
```diff
@@ -59,6 +58,54 @@ cdef extern from "MCorr.h" namespace "proteus":
 				   int* exteriorElementBoundariesArray,
 				   int* elementBoundaryElementsArray,
 				   int* elementBoundaryLocalElementBoundariesArray)
+        void calculateResidual_explicit_comp_quad(double dt,
```
cekees: likewise remove this, there should only be one calculateResidual

proteus/MeshTools.py
wacyyang: Thanks. I will keep this in mind.

wacyyang:
In this commit, the functions ***evaluateCoefficients_explicit_comp_quad***, ***calculateResidual_explicit_comp_quad***, and ***calculateElementResidual_explicit_comp_quad*** are deleted. Some indentions are corrected. Maybe it's still not good enough. I am trying to run some tests in order to get good convergence.

wacyyang:
@cekees Thanks for your correction of some bugs. I noticed that in **evaluateCoefficients** <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^{n&plus;1}_{rd})\bm{V})" target="_blank"></a> is used instead of <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^n)\bm{V})" target="_blank"></a>. Is this right?  Using this commit, I can run the vortex2D problem to the tolerance 10e-12.

cekees:
It turns out that the nonlinear solver failures on the first step were due to a bug with the new flux term. Specifically, the initial values weren't correct so the first corrected step had very large error.  It would be good to run this version with the new terms on and off for a range of hk's and meshes. You can compare those to the uncorrected level set and vof algorithms by setting `applyCorrection=False` and `applyRedistancing=False` (so that's 3 configurations by say 3 hk's by 3 meshes--see refStudy.sh script for example of how to script the runs). I think the next step after that is to extend to an implicit flux as @manuel-quezada did.  I suspect the next step after that should be to enable a bdf or crank-nicholson time discretization in the mass correction. You can also enable the composite quadrature for P2 elements by adding the required line to ModelFactory.h

wacyyang:
Thanks. I will do such comparisons.

manuel-quezada:
@wacyyang keep me on the loop on this so I can help whenever you have any problem.

wacyyang:
@cekees @manuel-quezada In these commits, 
I am trying to implement **(mass_correction_reference=2)**. It means the term <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot({H}_\epsilon(\phi^{n&plus;1}_{rd}&plus;u)\bm{V})" target="_blank"></a> is used in computing residual and Jacobian.  Can you help me to double check it? Thanks.

wacyyang:
In this commit, I include theta-method **(mass_correction_reference=4)** for the time discretization  of mass correction problem. It means the term <a href="https://www.codecogs.com/eqnedit.php?latex={H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot[(1-\theta)H_\epsilon(\phi^{n})\bm{V}^n&plus;\theta({H}_\epsilon(\phi^{n&plus;1}_{rd}&plus;u)\bm{V}^{n&plus;1})" target="_blank"></a> is used to compute residual and Jacobian. I have compared **(mass_correction_reference=4, theta=1)** and the implicit metho**(mass_correction_reference=2)**. They produce same solutions. It seems that Crank-Nicolson method **(mass_correction_reference=4, theta=0.5)** gives a better result 
screenshot 2017-12-21 12 23 50
screenshot 2017-12-21 12 23 04

cekees:
That looks good. What was the target CFL?

wacyyang:
It is *runCFL=0.3*. Compared to the original method **(mass_correction_reference=0)** which keeps the location or mass center but has many pieces, the Crank-Nicolson method **(mass_correction_reference=4, theta=0.5)** has one piece as whole but a little displacement of mass center. I am thinking about maybe combining them together like <a href="https://www.codecogs.com/eqnedit.php?latex=\gamma&space;H^{n&plus;1}&plus;(1-\gamma))\left\{{H}_\epsilon(\phi^n)-\Delta&space;t\nabla\cdot[(1-\theta)H_\epsilon(\phi^{n})\bm{V}^n&plus;\theta({H}_\epsilon(\phi^{n&plus;1}_{rd}&plus;u)\bm{V}^{n&plus;1})\right\}" target="_blank"></a> will produce a good result.

wacyyang:
@cekees I meet an error when trying to use P2 finite element **pDegree_ls=2**. After digging into the source code, I find it is related to https://github.com/erdc/proteus/blob/80b0db3a6ff0d05f437d43f1f8dcdc126315ea6f/proteus/mprans/MCorr.h#L2518 as indicated by its comment. Maybe we can use element diameter instead. What is your suggestion?

cekees:
Yes, just use `elementDiametersArray` and switch the index to eN. In fact we don't even need this bit of code for the new method because we don't project H into the finite element space. Except in the last variant you added with the gamma, but I think we're going to loose accuracy whenever we project H. I'm guessing the center of mass problem is related to the space and time truncation error in NCLS and MCorr.

wacyyang:
I will add elementDiameters into *setMassQuadrature* just as an extension and finish all tests first. I agree with you that adding gamma is not a good idea to solve the problem about the location of the mass center. Numerically, I can see the pollution of <a href="https://www.codecogs.com/eqnedit.php?latex=H^{n&plus;1}" target="_blank"></a> on the computation and that is why the last picture of the result on gamma=0.95 still doesn't have a good improvement. For smaller gamma, it looks even worse.

wacyyang:
Using P2-FE **(pDegree_ls=2)** and Crank-Nicolson method **(mass_correction_reference=4, theta=0.5)** I got this result. It shows big improvements including the location of the mass center :beers: More tests on using composite quadrature rules will be done. 

screenshot 2017-12-22 11 29 34

wacyyang:
I just pushed a commit where I am trying to implement backward Euler method in MCorr module. Unfortunately, it is still not convergent. I think I have some misunderstanding about this part (time integration, step controller ...)  Do you have any suggestions? Thanks for your help.

wacyyang:
As you suggested, I have run the rotation2D problem using different methods. The results are shown here

screenshot 2017-12-29 19 24 49

1. White-BDF3, Blue-BDF2, Green-BDF1
1. Since the final time is _T=1_, the difference between BDF1, BDF2, BDF3 is small.
2. As for the comparison of 3 equivalent CN+BDF1 (original method **mass_correction_reference=4**, using BackwardEuler **mass_correction_reference=5**, simplified  version **mass_correction_reference=6**), all have exact same information on the final data as shown in paraview

screenshot 2017-12-29 19 36 37
However, for vortex2D problem, the final solutions are different although the difference on the computation of _r, dr_ at every quadrature point is <10e-10.

PR #661, opened 2017-12-13
implementation of cubic Lagrange FE spaces in hexes

manuel-quezada:


cekees:
@manuel-quezada the tests are passing. The reason the coverage check didn't pass is that you just added a new python file that is not tested, which makes the overall coverage decrease. It's basically just trying to enforce "if you add new code, then you should also include tests for it". I don't intend to be pedantic about these coverage checks passing. Trying to increase the overall coverage is a worthy goal, however.

manuel-quezada:
@cekees great. I am glad the other tests are passing now. About the coverage check, yes this pull request is still a dummy one. I will be populating it with more commits as I proceed with this implementation.

PR #662, opened 2017-12-15
clean up mprans

manuel-quezada:
Use emacs to open each .h, .py and .pyx file in mprans and do the following: 

- with set-mark-command select the whole file. 
- untabify and delete-trailing-whitespace.


manuel-quezada:
@cekees , I untabify and run delete-trailing-whitespace in all .h, .pyx and .py files within mprans. Let's see if the checks pass and if you are satisfied we can merge this in.

PR #663, opened 2017-12-17
2nd clean up of some models in mprans

manuel-quezada:
Send new functions to the end of their respective classes. @cekees , can you  check if this is what you had in mind? If so and if all tests pass then we can merge it in.


proteus/mprans/cMCorr.pyx
```diff
@@ -6,498 +7,479 @@ from proteus.Transport import OneLevelTransport
...
+                               double* q_r,
+                               double* q_porosity,
+                               int offset_u, int stride_u,
+                               double* globalResidual,
+                               int nExteriorElementBoundaries_global,
+                               int* exteriorElementBoundariesArray,
+                               int* elementBoundaryElementsArray,
+                               int* elementBoundaryLocalElementBoundariesArray,
+                               double* H_dof)
+        void FCTStep(int NNZ,
```
cekees: I moved these to the bottom as we discussed, but I thought you hadn't modified MCorr* with FCT/EV. Did you mean to cut these out once you created MCorr2?

manuel-quezada: I think that MCorr2.h shouldn't be there. In github it says that the file was created 2 days ago. This makes me think that I added it in the first cleanup of mprans from one of my local branches. In other words, in some local branch I was experimenting with something and created the MCorr2.h file and two days ago I accidentally added it while I was doing a clean up of all .h files in mprans. So we can delete MCorr2.h

cekees: OK, so your  FCT and mass matrix stuff that you've added to MCorr is supposed to be in there, correct?

manuel-quezada: Correct. My changes to the mass correction model are in MCorr.h (and the respective pyx and py files) so we must delete MCorr2.h and cMCorr2.pyx.

proteus/mprans/cMCorr2.pyx
```diff
@@ -1,3 +1,4 @@
+# A type of -*- python -*- file
```
cekees: @manuel-quezada I just want you to check that there's nothing important in here before you remove it, assuming this was mistakenly added. Otherwise this is ready to go in.

manuel-quezada:
@cekees I removed the MCorr2.h and cMCorr2.pyx files. I double checked they were indeed not needed. In any case there are tests for the MCorr with EV so lets wait and  make sure they don't fail.

PR #664, opened 2017-12-18
Improves FluxBoundarConditions class documenation.

alistairbntl:


alistairbntl:
I've two fairly involved PR in the works, so I'm breaking off the parts of those PRs that are self-contained and creating branches directly off master.  This is the first of several to come.

PR #668, opened 2017-12-22
Clean up SpInv operator class and adds a test.

alistairbntl:
- Makes improvements to SpInv class documentation.
       - Adds a test for SpInv_shell.  Also introduced a test fixutre
       	 function and refactored an old LSC_shell test.
       - Cleans up white space in LinearAlgebraTools module.

### DIFF
--- a/proteus/LinearAlgebraTools.py
+++ b/proteus/LinearAlgebraTools.py
@@ -118,7 +118,7 @@ def superlu_sparse_2_dense(sparse_matrix,output=False):
 
     Parameters
     ----------
-    sparse_matrix : 
+    sparse_matrix :
     output : str
         Out file name to store the matrix.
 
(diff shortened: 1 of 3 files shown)


alistairbntl:
Is the codecov/patch is failing because only 10% of the diff is hit but the target is 35.12%?  If so, I suspect this is because I edited some whitespace in the `LinearAlgebraTools` module but only added a test for the `SpInv_shell` class.  If that is the case, I think we should ignore the `codecov/patch` and merge because overall coverage is increased.  That I still don't understand this coverage stuff too well and I may be missing something important.

Also, it is not clear to me why I'm getting coverage changes in files that I haven't touched.  It sounds like these 'unexpected coverage changes` happen, though its not clear to me why the `test_boundaryconditions` module would be affected by the PR in anyway.

cekees:
Yeah, I think when you edit the docstrings it shows up as changes that are not tested, because the docstrings are valid Python.  The coverage checks are mainly informational for now, not a requirement to merge.

PR #675, opened 2017-12-31
Modifies the two-phase pcd to allow Dirichlet pressure conditions.

alistairbntl:
* When the two-phase pcd schur complement approximation
      is used with strong Dirichlet pressure conditions, the
      null space projection introduces updates to the pressure
      solution that cause poor iterative solver performance.
      This PR introduces corrections to avoid this.


alistairbntl:
@cekees - I'm not sure if you have any thoughts about where the inviscid exception should be thrown, so I just added a `printf` warning statement.

cekees:
That's OK with me since it's an error that  one shouldn't be hitting anyway.

PR #676, opened 2018-01-03
WaveTools: SteadyCurrent fix

tridelat:
A small fix for the `SteadyCurrent` class in WaveTools where `eta` is set to `0` instead of `mwl` so the mprans.BoundaryConditions module imposes the VOF correctly in `setUnsteadyTwoPhaseVelocityInlet`.
@cekees @adimako


adimako:
Updated tests

adimako:
Fixed and closed, thanks @tridelat

PR #680, opened 2018-01-05
Gco fix wall function

Giovanni-Cozzuto-1989:
This PR is for enabling calculation of both velocity field and velocity gradient close to the wall for y+<11.25.
In this case the point is in the viscous sublayer and y+=u+ can be applied for imposing boundary condition on velocity.

Previous version of the code returned this error message: 'Prescribed near-wall point outside log-law region!'


Giovanni-Cozzuto-1989:
@cekees @adimako 
This PR was created to enable the calculation of velocity in proximity of the wall when the point is in the viscous layer rathen than the log-law region. 
Please, check if you are happy with it and feel free to merge it.

PR #683, opened 2018-01-10
Compare several methods for cylinder2D problem

wacyyang:
In this PR, we are going to collect and compare several methods to solve NSE for the benchmark problem *Test 2D-2*. The first step is to use exact same setting, compute drag/lift force and compare results with other literature.  Those methods include:
- [x] conforming mesh with rans2p method
- [x] conforming mesh with rans3p method
- [x] immerse boundary method
- [x] shift boundary method


proteus/tests/cylinder2D/conforming_rans2p/twp_navier_stokes_cylinder_2d_p.py
```diff
@@ -0,0 +1,125 @@
...
+    elif flag in [boundaryTags['obstacle'],boundaryTags['front'],boundaryTags['back']]:
+        return lambda x,t: 0.0
+    
+
+
+dirichletConditions = {0:getDBC_p,
+                       1:getDBC_u,
+                       2:getDBC_v}
+
+def getAFBC_p(x,flag):
```
wacyyang: @cekees I am not sure about this line. Is this correct?

cekees:
@wacyyang, could you look over these changes and make sure we don't need to keep any of them before we close this PR?

wacyyang:
@cekees I just removed some debug lines, tested rans2p and rans3p methods again and got right drag/lift coefficients. So I think it is safe to merge it into master branch now.

cekees:
@wacyyang Let's hold off on merging this. I may need to enable one more space/quad pair.

manuel-quezada:
Is the last commit trying to fix the problems that @wacyyang found when trying to use PresInc.h?

cekees:
I was trying to fix a segmentation fault @arnsong was getting and trying to provide a basic 3D test that works with all the optimized models.

manuel-quezada:
Yesterday @wacyyang and I had what looked like a segmentation fault but we realized that indeed was a problem of not having the correct space within ModelFactory. The macro in ModelFactory was displaying the error but the code not was aborting immediately. Therefore, we were not seeing the message from ModelFactory right away. For that reason we incorporated the abort() lines within ModelFactory. The idea is to abort immediately after realizing we don't have the correct space. That solved our problem.

cekees:
I saw that. It's  a big improvement.

cekees:
The tests aren't running for me locally.   You should see them run when you do `make test` in the proteus directory. @manuel-quezada has some experience with these kinds of issues.

cekees:
Looks like a path issue in `setup.py`  for the new data  files.
https://travis-ci.org/erdc/proteus/builds/342120672#L3954

PR #684, opened 2018-01-16
fix stepExact=False case for Sequential_MinAdaptiveModelStep

cekees:
The `systemStepExact` option in the so-file input controls whether the adaptive time stepping steps exactly to the user-defined steps in tnList. While the `stepExact=True` case safeguards against taking very small times steps, the VMS tau formula we use  is a function of the step size, which can result in significant perturbations of integral quantities like draft and lift. Setting `systemStepExact=False` allows smoother variation of the time step but doesn't archive exactly at the steps in tnList. This PR fixes the initial time step size for adaptive time stepping by using tnList[1], but after that will ignore the tnList for the purpose of time step selection. I also added a hook in parun to grab `systemStepExact` from the n-file instead of the so-file (e.g. for using it on simple problems like the cylinder @wacyyang ). @tridelat and @adimako I noticed in the input files I got from @tridelat that the non-exisitent `stepExactSystem` was being set. Using `systemStepExact=False`  may fix some pressure gauge noise that @Giovanni-Cozzuto-1989 has found. 

@manuel-quezada I did verify that removing the dt term from the SUPG formulas also fixes the problem, but I didn't have time to test it more generally. Might be worth looking at these oscillations in the RANS3P code where we can compare to EV.

### DIFF
--- a/proteus/SplitOperator.py
+++ b/proteus/SplitOperator.py
@@ -17,7 +17,7 @@ class SO_base:
 
     The base class implements sequential splitting with a fixed time
     step based on the input parameter `default_so.dt_system`. If
-    `default_so.stepExactSystem` is True then the time step will be
+    `default_so.systemStepExact` is True then the time step will be
     reduced when needed to match the output times in
     `default_so.tnList`, otherwise the output will be the first time
     step after each step in `tnList`.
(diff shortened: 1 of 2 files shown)


tridelat:
@cekees Thanks for spotting this, there was also noise (jumps) in the integrated forces around rigid bodies due to time stepping varying too much in order to match exactly the `tnList` (e.g. from `1e-3` to `1e-6`), even though this did not seem to affect the overall results in this case.
I pushed cdfaf81 directly here because the documentation of the SplitOperator mentions `stepExtactSystem` instead of `systemStepExact`, which might have led to the confusion.

Issue #688, opened 2018-02-01
optimize the proteus build


cekees:
whoever does this well will be a hero. I think the way to do it right is create a compiler macro like PROTEUS_BUILD_ALL_SPACES and then go in RANS2P.h, etc. and use that macro to enable all the existing quadrature/fem space combinations while of PROTEUS_BUILD_ALL_SPACES is undefined only build 1 quadrature rule and FEM space pair per FEM Space (e.g. P1 with 3rd order gauss, P2 with 5th order gauss, etc.) @mquezada have done this sort of thing during development, but what I'm talking about would allow the default `make install` to be fast and `make install PROTEUS_BUILD_ALL_SPACES=1` to support a full range of options.

adimako:
@manuel-quezada manuel has done something about it, suggest closing it @cekees

cekees:
Yes, there are two resolutions

1) build with PROTEUS_OPT="-O0 -UNDEBUG" FC=gfortran CC=mpicc CXX=mpicxx make ..
2) Set the FULL_BUILD macro to 0

PR #689, opened 2018-02-01
Mquezada/optimize build

manuel-quezada:
This is meant to consider a set of (default) template parameters within ModelFactory.h to speed up the compilation. The default parameters create P1, Q1, P2, Q2 FE-spaces with a single quad rule. 

See Issue #688

### DIFF
--- a/proteus/ModelFactory.h
+++ b/proteus/ModelFactory.h
@@ -2,6 +2,7 @@
 #define MODELFACTORY_H
 #include <iostream>
 
+//#define FULL_BUILD 1
 #define NO_INSTANCE std::cout<<"Constructing model object from template class:"<<std::endl \
   <<"return static_cast<Model_Base*>(new ModelTemplate<CompKernelTemplate<" \
   <<nSpaceIn<<","							\
@@ -14,8 +15,9 @@
   <<nDOF_trial_elementIn<<"," \
   <<nDOF_test_elementIn<<"," \
   <<nQuadraturePoints_elementBoundaryIn<<">());" \
-  <<std::endl<<std::flush;
+  <<std::endl<<std::flush
 
+#ifdef FULL_BUILD
 namespace proteus
 {
   template<class Model_Base, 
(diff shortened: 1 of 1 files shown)


manuel-quezada:
@cekees and @wacyyang. For the moment I consider by default P1, P2, Q1 and Q2 spaces with just one quad rule. I noticed an speed up in the compilation of about ~2. The first thing I want to see is which tests fail.

cekees:
Thanks, this should  be interesting.

manuel-quezada:
There are 3 tests that failed. These are: 

`__________________________ TestRichards.test_richards __________________________
/home/travis/build/erdc/proteus/linux2/lib/python2.7/site-packages/proteus-1.4.2-py2.7-linux-x86_64.egg/proteus/tests/griffiths_lane_6/test_griffiths_lane6.py:49: running the test CRASHED with signal 0
_______________________ TestRichards.test_elastoplastic ________________________
/home/travis/build/erdc/proteus/linux2/lib/python2.7/site-packages/proteus-1.4.2-py2.7-linux-x86_64.egg/proteus/tests/griffiths_lane_6/test_griffiths_lane6.py:75: running the test CRASHED with signal 0
_________________________ test_bochev_pressure_cavity __________________________
/home/travis/build/erdc/proteus/linux2/lib/python2.7/site-packages/proteus-1.4.2-py2.7-linux-x86_64.egg/proteus/tests/solver_tests_mprans/test_bochev_pressure_stabilization.py:72: running the test CRASHED with signal 0`

wacyyang:
I just compared those two settings and get this result. 

----------------------default case
real		7m40.444s
user	        7m27.140s
sys		0m7.350s
----------------------full build
real		17m40.644s
user	        17m22.440s
sys		0m14.580s

cekees:
We may want to add one additional p1 quadrature combination, which is lobatto (nodal)

manuel-quezada:
Hi @alistairbntl. As you might be aware, we are trying to optimize the build. I am currently working on reducing the set of template parameters that are considered within ModelFactory.h. This will basically reduce the number of compiled files within Proteus. Anyway, as a first start, I decided to compile by default just P1, P2, Q1 and Q2 spaces each with a given quadrature rule. I noticed that the test 

_tests/solver_tests_mprans/test_bochev_pressure_stabilization.py_

is not passing with this set of FE-spaces. I checked the reason for this and realized that you are using P1 spaces with 6 quad points (instead of 4). Is there any reason you  are using this many quad points? If not, can you change the test to consider 4 quad points? By doing that we won't need to pre-compile the set of template parameters for such test.

alistairbntl:
@manuel-quezada - I think 4 quadrature points should be fine.  Should I make the changes directly to this branch?

manuel-quezada:
@alistairbntl never mind. The compilation time in Travis got increased by just ~14 seconds due to that specific set of template parameters so it is not worth doing the change.

manuel-quezada:
@cekees. After doing: 
`
make distclean
make develop FULL_BUILD=1
`
I ran an example that uses a set of template parameters just within the **FULL_BUILD** section and the code didn't work; i.e, the compilation was done via the fast version. I am not sure how to pass the flag to the compiler. If I define **FULL_BUILD=1** inside ModelFactory.h, the compilation works as expected.

zhang-alvin:
@manuel-quezada Not sure if you found a way around the FULL_BUILD macro issue, but you should be able to pass in an environmental variable into the Proteus build environment through `proteus/config/default.py` or directly in `setup.py` to define a compiler flag to which you can supply to each relevant Extension like in https://github.com/erdc/proteus/blob/master/setup.py#L243-L257.  

For example in `config/default.py ` you can append to an existing compiler argument string
```
if (os.getenv('USE_FULL_BUILD')):
    PROTEUS_EXTRA_COMPILE_ARGS = PROTEUS_EXTRA_COMPILE_ARGS + ['-DFULL_BUILD']
```

PR #692, opened 2018-02-13
Mquezada/elliptic redist 3 p

manuel-quezada:
PR to incorporate and test elliptic re-distancing on the 3-phase code.


manuel-quezada:
@cekees this is the PR we talked about. Basically, I duplicated the code from NCLS and RDLS to NCLS3P and RDLS3P. I even copied the references for the automated tests and they all passed so they give the exact same solution.

manuel-quezada:
@cekees can you take a look at this PR? It started as a PR to extend elliptic re-distancing to RDLS3P but then I added the surface tension here as well. The surface tension is implemented in 2D and 3D for RANS3P. I feel the PR is still relatively small but, ideally, I don't want it to grow any more. Thanks.

cekees:
Sure. If I don’t get to it by Tuesday ping me again

Issue #697, opened 2018-02-24
add hotstart for p2 elements


cekees:
Hot starting just pulls the initial time and initial conditions from and existing solution archived in XDMF. The finite element space has to provide a read function for itself, which is called here:

https://github.com/erdc/proteus/blob/master/proteus/NumericalSolution.py#L1164

We need to provide this function for the higher order elements.

adimako:
I think this has been addressed

PR #699, opened 2018-02-26
Add hotStart for P2 finite element (Issue #697)

wacyyang:
In this pull request, we will solve the problem of reading a P2 finite element function from h5+xmf file into Proteus.


wacyyang:
To check this implementation, I am running cylinder problem 2 times. The **hotStart** is used in the 2nd run. The time interval for the first run is [0,0.2], and for the 2nd run is [0,0.1]+[0.1,0.2]. The results show a big difference in _p_ and _u_ even at _t=1.01_ after one time step. The same difference is observed for the P1P1 case also.  Maybe it is related to the _Spinup_ process.

wacyyang:
I realized a possible reason for the failure of using HotStart in the above parallel case. It is because the data is saved using 1 processor. If I choose to save the data using 2 processors, and use 2 processors to apply HotStart, then there is no inconsistency in reading data. And it seems this piece of code works for all finite element spaces, and maybe we can move it into the base class.

proteus/FemTools.py
```diff
@@ -5398,6 +5398,18 @@ def writeFunctionHeaderEnsight(self,u,filename,append=False,firstVariable=True,c
                                   u.name+' '+filename+u.name+'.scl****\n')
     def writeFunctionXdmf(self,ar,u,tCount=0,init=True):
         self.XdmfWriter.writeFunctionXdmf_C0P2Lagrange(ar,u,tCount=tCount,init=init)
+    def readFunctionXdmf(self,ar,u,tCount=0):
+#         import pdb
```
cekees: remove these commented debugging lines

proteus/NumericalSolution.py
```diff
@@ -1185,8 +1185,10 @@ def calculateSolution(self,runName):
             if time >= self.tnList[-1] - 1.0e-5:
                 logEvent("Modifying time interval to be tnList[-1] + tnList since tnList hasn't been modified already")
                 ndtout = len(self.tnList)
-                dtout = (self.tnList[-1] - self.tnList[0])/float(ndtout-1)
-                self.tnList = [time + i*dtout for i in range(ndtout)]
+                #dtout = (self.tnList[-1] - self.tnList[0])/float(ndtout-1)
```
cekees: remove comments

proteus/default_so.py
```diff
@@ -30,3 +30,5 @@
 
 dt_system_fixed = None
 """A system-wide wide time step used by SplitOperator objects"""
+
+skipSpinupOnHotstart = False
```
cekees: add a doc string to this option

PR #701, opened 2018-02-27
Submodule updated with shorter tests

Pedrohrw:
At this time, I created a new submodule. Now, this submodule is pointing at the branch "pom-tests" at the air-water-vv repository. Moreover, the air-water-vv tests that will be run by Travis every time we make a pull request in proteus repository are shorter. I reduced the duration of the simulation in order to try to addapt to Travis limit time.
This change is provisional, since the submodule has to point in the master branch of the air-water-vv tests in the end. In the branch "pom-tests" in the air-water-vv repository I have removed the .travis.yaml file, since the air-water-vv tests will be run in the proteus submodule togheter with proteus.
I was discussing about this with @adimako, but it is not completely sure.

So, we have to decide what is the best way to test all these cases.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "air-water-vv"]
 	path = air-water-vv
-	url = https://github.com/Pedrohrw/air-water-vv.git
-	branch = master
+	url = https://github.com/erdc/air-water-vv.git
+	branch = pom-tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
 
 script:
 - py.test --boxed -v linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus/tests --cov=linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus
-- py.test --boxed -v air-water-vv/Tests/test_dambreak_Colagrossi.py
+- py.test --boxed -v air-water-vv/Tests/
 
 notifications:
   email:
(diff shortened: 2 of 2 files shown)


cekees:
BTW, you may want to take a look a this: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/. For some reason your commits aren't getting linked up properly.

cekees:
I'm not sure why two of the tests are failing, but it looks like an import path issue. In order to test this you may need to do `make distclean; make install` and then run the test command as it is down in .travis.yaml

Pedrohrw:
I think I know why these two tests are failing. I am going to try one more time. If not, I will try what you are suggesting. 

Thank you Chris.

cekees:
One thing we could do to speed up the builds is use the travis build matrix support. Right now we run the air-water-vv tests after the main proteus tests:

```
script:
- py.test --boxed -v linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus/tests --cov=linux2/lib/python2.7/site-packages/proteus-$(python -c "import proteus; print proteus.__version__")-py2.7-linux-x86_64.egg/proteus
- py.test --boxed -v air-water-vv/Tests/```

Instead, we could use the travis build matrix support, which allows you to define multiple "env's" that run in parallel. To be gin with we could just separate the main proteus and air-water-vv tests, but it would probably be useful to separate the basic proteus regression tests, which run quickly, from the longer tests that run full models. The full test matrix support is documented here:

https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

And specific ideas on how to use it to speed up the test suite are here:

https://docs.travis-ci.com/user/speeding-up-the-build/

**Comment by cekees** [comment]
Another thing we can do to speed up the tests is to cache the .hashdist directory. I think we could do something like this in .travis.yaml:

```
cache:
  directories:
  - $HOME/.hashdist
```

PR #703, opened 2018-02-28
Travis cache

cekees:


cekees:
Caching hashdist src and build cut the total test time to 30 minutes and allowed both the standard proteus tests and the air-water-vv tests to run.
/cc @Pedrohrw @adimako

PR #704, opened 2018-03-01
Fix calculate Residual and Jacobian in rans3p

wacyyang:
In the last merge, some new parameters are added to the RANS3P 2D case. In this PR, we add same parameters for entropy method and 3D cases in order to keep the interface of _calculateResidual_ and _calculateJacobian_ consistent.


cekees:
@arnsong and @Milad-Rakhsha this also changes to our new convention on signed distance and solid normal for level set representation of the solid (negative inside solid, normal pointing out of solid, so n=grad(phi_s)). Thanks @wacyyang .

PR #707, opened 2018-03-08
Fenton waves for WaveTools

tridelat:
Fenton waves from John Fenton's script, modified to work on Linux.
http://johndfenton.com/Steady-waves/Fourier.html
Fenton waves can be created by selecting `waveType='Fenton2'` for `MonochromaticWaves` class in `proteus.WaveTools`, by passing the same information as a `'Linear'` waveType.
Additional stack dependencies: ncurses.
@adimako @cekees


cekees:
This is nice. Did we get his permission? I know @adimako was discussing it. If we're going to cut/paste his code in we'll need to get his permission to add his copyright to the CONTRIBUTORS file under the agreement that this code using the same license. Otherwise we might want to split this off a s separate python package so he can license it however he likes.

tridelat:
@cekees We received his permission to use it but I think we would need to ask again if we can add him to the list of contributors before merging in (@adimako?) 
It is also possible to separate it as a standalone python package if it is more convenient for licencing purposes (it also could make it accessible outside of Proteus).

tridelat:
We got his permission to add the code under the MIT license and add his name to the list of contributors. I think just a small test is needed @adimako @cekees

cekees:
@tridelat if the tests pass, please feel free to squash and merge when you're ready.

tridelat:
@cekees It looks like it exceeded the allocated time for the build to finish

PR #708, opened 2018-03-09
PETTT Feature #1: Model Reconstruction

zhang-alvin:
First of a series of PRs for various features that have been implemented. The primary feature of this PR is geometric model reconstruction from Proteus mesh data structures with proper topology but not shape information.

The model reconstruction handles 3D geometries and identifies external boundaries by the material flags.
The reconstruction capability is handled in SCOREC/core and assembling the requisite inputs are the primary modifications to Proteus. 

- [ ] add in 2D model reconstruction for jupyter notebook
- [ ] update jupyter notebook to show capability

PR #709, opened 2018-03-09
PETTT Feature #2: Anisotropic grading

zhang-alvin:
Second part of series that depends on the merge of #708. 
The primary features pertain to anisotropic mesh adaptation:

- initial element size computation based on shortest height
- aspect ratio grading


proteus/MeshAdaptPUMI/SizeField.cpp
```diff
@@ -660,6 +674,46 @@ static void SmoothField(apf::Field *f)
   op.applyToDimension(0);
 }
 
+void getTargetError(apf::Mesh* m, double &target_error){
+  //Implemented for 3D and for serial case only so far
+  assert(m->getDimension()==3);
+  std::cout<<"Enter target Error\n";
```
cekees: Do these print statements get called on every processor?

PR #716, opened 2018-03-16
Fix EV by merging calculateResidual functions 

manuel-quezada:
The main reason for this PR is to re-implement the EV artificial viscosity in RANS3PF inside the calculateResidual and calculateJacobian original functions. 

- After re-implementing the EV art viscosity I delete calculateResidual_entropy_viscosity and calculateJacobian_entropy_viscosity. 
- I delete several parameters passed to calculateResidual that are not used. They were there due to some tests I did long time ago that never lead to a working method/code. 
- I make minor changes to the EV method within NCLS3P and VOF3P based on results from marin3p.
- I make automated tests to check the convergence of the projection scheme and one benchmark for a rising bubble problem.


manuel-quezada:
@cekees two of the rising bubble tests are failing here but not on my machine. Do you have any idea what might be happening?

manuel-quezada:
@cekees I checked the triangleFlag and I think there was a minor error in the documentation. I corrected it and push the changes. Did you do something to fix the problem we had in Travis? I am not sure why Travis passed at the end. My guess is that we had some issue with the way the files were being tracked and that removing and put them back fixed the problem.

PR #718, opened 2018-03-22
advective penalty flux for non-conservative form

cekees:
When we using the non-conservative form of the NSE momentum equations we need to use an upwind penalty flux in the case of flow entering the domain. We already to this in RANS3P. This PR adds them for RANS2P. This term effectively stabilizes cases where we have flow into the domain due to hydrostatic pressure Dirichlet conditions with outflow conditions, or in some cases with velocity Dirichlet conditions for inflow.

### DIFF
--- a/proteus/mprans/BoundaryConditions.py
+++ b/proteus/mprans/BoundaryConditions.py
@@ -586,6 +586,17 @@ class BC_RANS(BoundaryConditions.BC_Base):
                                             -
                                             smoothedHeaviside_integral(smoothing, phi)))
 
+        def diffusiveFlux_u(x,t):
+            g = hydrostaticPressureOutletWithDepth_p_dirichlet(x,t)
+            phi = x[vert_axis] - seaLevel
+            if phi >= smoothing:
+                H = 1.
+            elif smoothing > 0 and -smoothing < phi < smoothing:
+                H = smoothedHeaviside(smoothing, phi)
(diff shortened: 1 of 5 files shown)


cekees:
@tridelat I didn't realize this PR was touching the BoundaryConditions module or I would have added you earlier.

manuel-quezada:
Alistair explained me the idea: in outflow boundaries set u=0 if rho*(u.n)<0 to avoid re-circulation. This is done via double integration by parts of the advective non-linear term in the momentum equations. If the BCs are imposed via flux conditions then the condition is set to 0 directly; otherwise, it is set to u_bc. The code makes sense to me.

cekees:
Yeah, we were already doing this in RANS3P here: https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS3PF.h#L1472

Although, looking at that, we need to make one more change to make them consistent. Notice that the if there is NO velocity dirichlet condition and flow  is coming into the domain, then RANS3P doesn't do anything whereas RANS2P defaults to u_bc=0. This makes a difference when you're doing strong boundary conditions on velocity but have an outlfow boundary (and hence can't specify at least one velocity component). I'll add that to RANS3P.

alistairbntl:
Here is the new boundary condition function.   I'm running an example to make sure it works so please don't merge until I've approved.

PR #719, opened 2018-03-23
switched back to default optimization options

cekees:
- `PROTEUS_OPT="-O3 -DNDEBUG"` is again the default
- `export PROTEUS_OPT="-O0 -UNDEBUG"` to get fast compile times but slow proteus
- travis will use the fast build


manuel-quezada:
Nice! I like this better.

PR #721, opened 2018-03-23
Parameter classes

cekees:
This allows defining input parameters for physics, numerics, and split operator (system) information as objects instead of modules. Among other things, the idea is to allow much easier usage in the test framework and jupyter notebooks where modules don't work well because they are singletons (effectively globals).  This PR is trying to provide a migration path that doesn't break module-based input files. Check the tests I modified for demonstration of how we can get rid of all the tricky reload statements.


cekees:
Thanks. Yes, that's not hard to fix.

cekees:
@tridelat , please give it another shot:
```
>>> from proteus.defaults import Physics_base
>>> p1 = Physics_base()
>>> p2 = Physics_base()
>>> p1.dirichletConditions[0]=1
>>> p2.dirichletConditions[0]=2
>>> assert(p2.dirichletConditions[0] != p1.dirichletConditions[0])
```

PR #726, opened 2018-03-29
Fix bugs on CLSVOF during parallel runs

manuel-quezada:
Small PR to fix minor issues when running CLSVOF model in parallel. Most of the work is done for the parallel computations of some metrics on the solution.


manuel-quezada:
@cekees, in reality they are scalars that I need to compute inside the c++ code and use in the python code. I thought on two ways of implementing this. One way (which is what I am doing) is to declare them as arrays of length 1 and treat them as any other array. Alternatively, I can require the c++ function to "return them". This second option is somehow similar to what I do here (https://github.com/erdc/proteus/blob/master/proteus/mprans/CLSVOF.h#L168) to compute some metrics. This is a trick I copied from you from some of the code you did in Germany, also to compute some metrics.

proteus/mprans/CLSVOF.py
```diff
@@ -576,15 +579,25 @@ def __init__(self,
         # FOR nonlinear CLSVOF; i.e., MCorr with VOF
         self.timeStage=1 #First order time integration
         # For normal reconstruction
-        self.norm_factor = numpy.zeros(1)
+        self.min_distance = 0.
```
cekees: If we stay with length 1 arrays, I think these should be `*_distance[0] = 0`.

cekees: Oh, nevermind. I see what you're doing.

manuel-quezada:
Great! The PR is also good to go from my side so you can merge it. Thanks.

PR #727, opened 2018-03-31
SBM for 3D NSE

wacyyang:
In this pull request, we will try to finish following tasks as an extension of 2D SBM:
- [X]  implement 3D version SBM
- [X] modify 2D version SBM to make it has the same structure as the 3D version
- [ ] check drag/lift coefficients for 3D benchmark problems from the paper _Benchmark Computations of Laminar Flow Around a Cylinder_


wacyyang:
@cekees As shown in the Figure, the drag/lift coefficient is close to the result in the paper, although there is no obvious vortex shedding behind the cylinder. 

cd-cl

There is only slight oscillation behind the cylinder at the beginning stage and it is smoothed later. Maybe it is because of the numerical diffusion. If the velocity at the inflow boundary is set to be 10 times bigger, there are some clear oscillation. At this stage, I think we can close this pull request and merge the branch so that other group members can join us to test the implementation. As for this benchmark problem, 3D-2Z, maybe we can do a careful test in another pull request, by comparing different method such as turning off numerical viscosity or SUPG, using high order FEM, or running on a finer mesh.

PR #728, opened 2018-04-03
interface clsvof with rans3pf

manuel-quezada:
This PR is meant to provide the needed interface between CLSVOF and RANS3PF. I created a 2D and a 3D automated tests for this PR.In both cases I consider a drop falling in an empty rectangular container. In addition, I compute the L2 norm of the error (if the exact solution is provided) at the EOS (End Of Simulation).


manuel-quezada:
@cekees this PR is a first PR to combine the CLSVOF model with the NS models. I start with RANS3PF. Later I will work on RANS2P as well. During the tests I had to play with the tolerance. Having atol=10^-10 was to restrictive, I had to set atol=10^-8. I think the reason for this is that the tolerance on the nonlinear solver is smaller and that the codes are compiled in a different way (optimized vs non-optimized).

cekees:
@Pedrohrw you might want to look at this test problem. The idea is that it is easy to switch to the new version of clsvof method. 

/cc @manuel-quezada @adimako

PR #729, opened 2018-04-05
Pom/air water vv tests

Pedrohrw:
- Submodule updated 
- Enable air-water-vv tests


adimako:
@Pedrohrw it is failing, I think it is because the path in air-water-vv has not been updated

adimako:
No, that should be back to "True". We did this because it could not find the Fenton module, but now that's been fixed with the autoFenton PR.

adimako:
Happy to merge this, when the tests pass

PR #730, opened 2018-04-05
Putting latest updates of master to gco_rans3p

adimako:


adimako:
I will go ahead and merge without the tests. Tests will be sorted in #612

PR #738, opened 2018-04-14
Mquezada/spin up for CLSVOF and interfacing CLSVOF with RANS2P

manuel-quezada:
This PR is meant to:

- **Enable a spin up step for CLSVOF**. The spin up step is a L2 projection. When we use high-order spaces we need to use Bernstein polynomials to assure a non-singular lumped projection during the reconstruction of the target for the penalization. In this situation we need to be careful how the initial condition is represented into the space. This is true since the Bernstein polynomials are not interpolatory. To avoid loosing accuracy during this step we can do a L2 projection.
 
- **Provide the interface between CLSVOF and RANS2P**. This comes with an automated test. 

- I also put some code to do the normal reconstruction via a consistent L2 projection (instead of lumping the mass matrix). This is not a good idea unless some extra treatment is done to avoid non-physical oscillations near the peaks. The code comes with an FCT step that uses global bounds to guarantee the reconstruction is between -1 and 1. The code works in serial but for the moment this part is disabled (it can't be reached) until we do more research about it.


manuel-quezada:
@cekees I think this PR is ready. The air-water-vv tests are taking way too long. Maybe it would be good to consider lighter versions of them when testing.

PR #743, opened 2018-04-19
Random change just to try

Pedrohrw:


Pedrohrw:
The purpose of this pull request is just to check if the convergence check implemented in the air-water-vv tests works before merging it into the awvv master branch. This convergence check could be found at the branch 'pom/tests' at the air-water-vv repository. With the result I get from here, it seems to be the new changes done at the air-water-vv tests are working. 
The two tests that are failing at this pull request are:
- Colagrossi: I forced it to fail. To check if the changes had been correct, I forced one test to fail and it fails like it should be.
- Wavesloshing: Unexpectedly this test case is not converging. So because of the implementation of the convergence check it seems to be that this case is not converging when one of the last proteus versions is used to run the simulation. @cekees  @adimako @tridelat

adimako:
I recommend finding out what the issue is with the sloshing case and then merging

Pedrohrw:
I have been compiling different proteus versions and I noticed that the wavesloshing test case starts to fail after the commit 2744498cf5bc869817a10b8b9e078d3a3d629817 (Adds the Bochev pressure stablization method to RANS2P2D). @adimako @cekees @tridelat

Pedrohrw:
@alistairbntl

cekees:
@Pedrohrw try removing the pressure projection stuff from RANS2P2D.h to restore the calculation to be identical  to before  that commit. It's  really only a few lines that  need  to  be removed. If that fixes  the issue then we know there's a subtle issue. I don't see anything that should cause a problem in the code @alistairbntl added, but there  could  be a memory overwrite  if  there's a mistake in the indexing.

PR #746, opened 2018-04-24
FSI: moorings visualisation (h5) and fixes

tridelat:
- [x] override setup initial of elements for rest length
- [x] visualisation of moorings with h5/xmf files
- [x] node parallel distance search fix
- [x] switches added for external forces


tridelat:
@cekees this is only for the moorings, the same could be done for rigid bodies if necessary though (for now it is easy to visualise in paraview as long as a conforming mesh technique is used)

PR #748, opened 2018-05-01
add comment at SpatialTools

Pedrohrw:


Pedrohrw:
Indeed @cekees. Just that. It is the only porpuse of this pull request.

PR #749, opened 2018-05-03
Separate awvv tests

Pedrohrw:


Pedrohrw:
This could be a different way of testing the air-water-vv cases. Instead of testing them all togheter, we could separate them in different processes. It could be maybe a bit less efficient, but at least we are a bit more sure that it is not stopping due to the time limitation. Could it be valid? 
@cekees  @adimako

cekees:
I like it! In the long run maybe we can get the test set organized into groups that take about 30 minutes.

PR #751, opened 2018-05-03
When all particles are balls

wacyyang:
In this PR, we will try to speed up IBM for the case that all particles are balls since all computations on distance, normal direction, and velocity at any (quadrature) point in the global coordinate can be obtained directly in `.h` file.


cekees:
Looks good. @arnsong @dtoconn @tweigand you might want to be aware of this branch bypassing external sdf, normal, and velocity calculations for the special case of spherical particles. @wacyyang you haven't added the angular velocity correction yet, right? Or did I miss it?

wacyyang:
Yes, you are right. I am going to add angular velocity now.

wacyyang:
@cekees I think this PR is ready to merge. Both 2D and 3D RANS3P can use ball as particle now.  The macro  `#define USE_CYLINDER_AS_PARTICLE`in RANS3PF.h can be commented out when merging this branch. This macro is used to set the particle as cylinder and is just for debug now.

PR #752, opened 2018-05-09
Adimako/add penalty for packing

adimako:


adimako:
- [ ] Stabilise model while sediment is at rest (pressure friction module)
- [ ] Correct shear friction module
- [ ] Validate sediment settling velocity

PR #757, opened 2018-05-15
Lag eddy viscosity

cekees:
The Smagorinsky and dynamic Smagorinsky models introduce an eddy viscosity that is nonlinear in the velocity and velocity gradient. Here I add a switch to allow using the value from the previous time step (i.e. and explicit approximation of the eddy viscosity). I also fixed a  bug where the eddy viscosity  wasn't getting set on external boundaries, which is important for outflow boundaries in particular.   So far, tests show that lagging the eddy viscosity recovers  quadratic nonlinear solver performance, with weak and strong boundary conditions.


cekees:
@adimako I just needed to apply the changes to  RANS2P2D

PR #758, opened 2018-05-17
Interface between solid and fluid

wacyyang:
In this PR, we are trying to fix some small bugs and as a first step to unify the interface between solid motion and momentum equation because several ways to input solid information are available now, including: `granular_sdf_Calc`, `particle_sdf` or  `use_ball_as_particle`.

### DIFF
--- a/proteus/mprans/RANS3PF.py
+++ b/proteus/mprans/RANS3PF.py
@@ -481,6 +481,7 @@ class Coefficients(proteus.TransportCoefficients.TC_base):
         if self.use_ball_as_particle==1:
             pass
         elif self.granular_sdf_Calc is not None:
+            corresponding_point_on_boundary = np.zeros((3,),'d')
             temp_1 = np.zeros(self.model.q[('u', 0)].shape, 'd')
             temp_2 = np.zeros(self.model.q[('u', 0)].shape, 'd')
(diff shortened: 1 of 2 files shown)


proteus/mprans/RANS3PF.py
```diff
@@ -521,7 +525,7 @@ def attachModels(self, modelList):
                             self.ebq_global_grad_phi_s[ebN,kb,:]=sdNormals
                             for j in range(len(sdNormals)):
                                 corresponding_point_on_boundary[j] = self.model.ebq_global['x'][ebN,kb][j] - sdf_ebN_kb*sdNormals[j]
-                            self.ebq_particle_velocity_s[ebN,kb,:]=vel(0.0,corresponding_point_on_boundary)
+                            self.ebq_particle_velocity_s[ebN,kb,:] = vel(t=0.0, corresponding_point_on_boundary, self.particle_centroids[i])#YY:But self.particle_centroids[i] has not been updated?
```
wacyyang: @cekees Considering the first step, because `self.particle_centroids` is initialized as `0` and the information of solid has not been passed into this module, the information here is not correct. Maybe we can move this part to `preStep` and delete similar update steps in `postStep`.  What do you think?

cekees: If we're going to be providing analytical velocities then the centroids should also be provided a this point, shouldn't they?

PR #760, opened 2018-05-29
Clean up and debugging interface stresses

adimako:


adimako:
@cekees I think we should wrap up this one. Results for the settling velocity are decent (see below)

adimako:
capture

PR #765, opened 2018-06-12
Embedded LogEvent using Python/C API

zhang-alvin:
First attempt at getting some profiling functions into the C-code to finally begin to address Issue #3 

This PR introduces an interface (wrapper?) to call Profiling.logEvent() from C-code using the Python/C API. The implementation is based primarily on the example in `https://docs.python.org/2/extending/embedding.html`. 
There are 2 inputs to this interface: 

1.  a char* array for the desired string to be logged
2.  an integer for the logging level 

Example Usage:
```
          char buffer[50]; 
          sprintf(buffer,"A tree grows in vicksburg %i %f",5,1.235678);
          logEvent(buffer,2);
```
I've tested for a memory leak by calling the logEvent() function in getResidual() 10000 times:
```
        for (int i=0;i<10000;i++){
          char buffer[50]; 
          sprintf(buffer,"A tree grows in vicksburg %i %f",5,1.235678);
          logEvent(buffer,2);
        }
```
and compared the memory profile against a run without the call to logEvent. The difference in memory usage is negligible (<1e-6 MB). 

Thanks to @wacyyang for initial testing.


zhang-alvin:
@cekees Yes this current approach does not yield a memory leak. Below is a snippet of what the output looks like in the log file from a test case (see 2nd line which uses the logEvent() function in mprans/RANS2P2D.h):

```
[    6994] after ksp.rtol= 0.01 ksp.atol= 9e-09 ksp.converged= True ksp.its= 51 ksp.norm= 0.000403216738037 reason = 2
[    6994] Entered rans2p 2D calculateResidual
[    6994] Maximum CFL = 0.33

```

PR #777, opened 2018-07-06
fixed missing initialization of basis gradients

cekees:
This appears to be what was causing an intermittent problem with using `LevelModelType = Pres.LevelModel` (the optimized pressure calculation in RANS3P). 
/cc @manuel-quezada @wacyyang @adimako @Pedrohrw @arnsong


cekees:
@manuel-quezada , yes it was the uninitialized memory that was producing garbage. I'm not sure how reproducible the error would be across a lot of machines, but the the rotational form was definitely not right with `LevelModelType=Pres`

manuel-quezada:
Agree! It never worked for me indeed.

PR #781, opened 2018-07-16
Fix some bugs in SBM

wacyyang:
In this PR, we are going to fix some bugs found in the implementation of SBM for RANS3P compared to the implementation of SBM for Poisson problem. 
  
1.  - [x] initialize `isActiveDOF` outside
2.  - [x] use ebN instead of eN
3.  - [x] check surrogate edges such that at least one of its nodes is owned
4.  - [x] avoid double couting for force computation
5.  - [x] use `normal` to keep consistency 
6.  - [x] check `h_penalty<=dist`
7.  - [x] use right `ebN_i_j` for sparse pattern


proteus/mprans/RANS3PF2D.h
```diff
@@ -4621,11 +4641,12 @@ namespace proteus
                     int ebN = elementBoundariesArray[eN*nDOF_mesh_trial_element+opp_node];//only works for simplices
                     surrogate_boundaries.push_back(ebN);
                     //now find which element neighbor this element is
-                    if (eN == elementBoundaryElementsArray[eN*2+0])
-                      surrogate_boundary_elements.push_back(1);
+                    if (eN == elementBoundaryElementsArray[ebN*2+0])
```
cekees: the old code and the new code is equivalent here, right? You're just asserting what the old code was assuming?

proteus/mprans/RANS3PF2D.h
```diff
@@ -4676,11 +4697,11 @@ namespace proteus
                   }
                 else if (pos_counter == 3)
                   {
-                    element_active=1.0;
+                    element_active=1;
```
cekees: switch back to 1.0, and fix other instances

proteus/mprans/RANS3PF2D.h
```diff
@@ -5459,13 +5481,13 @@ namespace proteus
                     /* globalJacobian[csrRowIndeces_p_w[eN_i] + csrColumnOffsets_p_w[eN_i_j]] += elementJacobian_p_w[i][j]; */
 
                     /* globalJacobian[csrRowIndeces_u_p[eN_i] + csrColumnOffsets_u_p[eN_i_j]] += elementJacobian_u_p[i][j]; */
-                    globalJacobian[csrRowIndeces_u_u[eN_i] + csrColumnOffsets_u_u[eN_i_j]] += element_active*elementJacobian_u_u[i][j];
-                    globalJacobian[csrRowIndeces_u_v[eN_i] + csrColumnOffsets_u_v[eN_i_j]] += element_active*elementJacobian_u_v[i][j];
+                    globalJacobian[csrRowIndeces_u_u[eN_i] + csrColumnOffsets_u_u[eN_i_j]] += elementJacobian_u_u[i][j];
```
cekees: switch back to original way of doing it by calculating instead of if statement

proteus/mprans/RANS3PF2D.h
```diff
@@ -5622,8 +5650,10 @@ namespace proteus
                         const double zero_vec[2]={0.,0.};
                         for (int j=0;j<nDOF_trial_element;j++)
                           {
-                            register int ebN_i_j = ebN*4*nDOF_test_X_trial_element + i*nDOF_trial_element + j,
-                              ebN_local_kb_j=ebN_local_kb*nDOF_trial_element+j;
+                            register int ebN_i_j = (ebN*4
```
cekees: still not quite sure about this. I'll come back to this.

wacyyang:
I will make those changes following your suggestions. Another main change is the initialization of `isActiveDOF `. Without this change, it will depend on the order of looping over all elements.

proteus/mprans/RANS3PF2D.h
```diff
@@ -2217,6 +2217,16 @@ namespace proteus
         const int nQuadraturePoints_global(nElements_global*nQuadraturePoints_element);
         std::vector<int> surrogate_boundaries, surrogate_boundary_elements, surrogate_boundary_particle;
         //std::set<int> active_velocity_dof;
+        if(USE_SBM>0)
```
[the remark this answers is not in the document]
cekees: OK, I see the dof may be interleaved and using memset could be tricky. Fine to leave it  like it is for now.

proteus/mprans/RANS3PF2D.h
cekees: OK, I'm convinced you are correct here.

proteus/mprans/RANS3PF2D.h
```diff
@@ -3264,7 +3277,7 @@ namespace proteus
                       }
 
                     double dist = 0.0;
-                    double distance[2], P_normal[2], P_tangent[2]; // distance vector, normal and tangent of the physical boundary
+                    double distance[2], P_normal[2], P_tangent[2], normal_Omega[2]; // distance vector, normal and tangent of the physical boundary
```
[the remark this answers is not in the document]
wacyyang: `P_normal` is the unit normal vector pointing outward of the particle; `normal_Omega` is the unit normal vector pointing outward of the true fluid domain; `normal` is the unit normal vector pointing outward of the surrogate domain. `normal` is used for consistency, but for approximation `normal_Omega` looks better. Numerically, I see no big difference between those two cases. I just use it for debugging. I think it is better to delete `normal_Omega` for clarity.

cekees:
Replaced by #793

PR #782, opened 2018-07-19
Mquezada/new elliptic redistancing

manuel-quezada:
PR to incorporate some changes in the elliptic redistancing.


manuel-quezada:
@cekees I think this is ready to go. I modify the broken tests. I also checked the results of the two benchmarks for the rising bubble. These tests are important since the old redistancing (Eikonal eqn + SUPG + shock capturing) was not able to produce decent results. Since I got somehow closer to that method I was concerned about these tests but they did good.

PR #783, opened 2018-08-02
removed hide_code jupyter extension

cekees:
The hide_code extension is no longer maintained and was breaking with recent updates to other parts of the statck, so I removed it. You need to do a `make distclean` for this to take effect (e.g. `make distclean; make N=4 PROTEUS_OPT="-g -UNDEBUG" develop; make jupyter`

/cc @alistairbntl and @bchambers28


alistairbntl:
@cekees - I have no preference if this or the other PR I submitted today is used.  Either way works for me.  If hide_code will continue to be a problem in the future than perhaps this is the best path forward.

PR #784, opened 2018-08-03
update code hide version

alistairbntl:
@cekees, @bchambers28 - updating the hide code version seems to have fixed the jupyter problems I was having.


bchambers28:
Updating the hide_code version fixed the problem with jupyter for me as well.

Issue #766, opened 2018-06-20
Mass Correction: MCorr.postStep() behavior


zhang-alvin:
I'm having some trouble understanding the mass correction procedure in terms of how the degrees of freedom are manipulated. In MCorr.postStep(), there are the following lines:
(https://github.com/erdc/proteus/blob/master/proteus/mprans/MCorr.py#L169-L173)
```
            self.lsModel.u[0].dof += self.massCorrModel.u[0].dof
            self.lsModel.q[('u', 0)] += self.massCorrModel.q[('u', 0)]
            self.lsModel.ebqe[('u', 0)] += self.massCorrModel.ebqe[('u', 0)]
            self.lsModel.q[('grad(u)', 0)] += self.massCorrModel.q[('grad(u)', 0)]
            self.lsModel.ebqe[('grad(u)', 0)] += self.massCorrModel.ebqe[('grad(u)', 0)]
```
This modifies the degrees of freedom and quadrature point fields of the level-set model, but this modification isn't reflected in the modelList.uList set of unknowns. So the result is the following sequence:

1.  MCorr.postStep() updates quadrature point fields and u.dof[:] of the level-set model
2.  RANS2P uses modified quadrature point fields
3.  Solver reaches level-set model again. When computing the residual, the modelList.uList set of unknowns writes over the u.dof[:] array.

This means that RANS2P uses the mass-corrected level-set field, but the level-set model is updating the uncorrected level-set field from the previous time step. 

Is this expected behavior or am I interpreting something wrong?

Example of behavior at step 3 in NCLS.getResidual(), just before calling `self.setUnknowns()` at https://github.com/erdc/proteus/blob/master/proteus/mprans/NCLS.py#L1185:

```
(Pdb) print self.u[0].dof.mean()
0.0754403492887
(Pdb) print u.mean()
0.0754478228232
```
 and just after the call

```
(Pdb) print self.u[0].dof.mean()
0.0754478228232
(Pdb) print u.mean()
0.0754478228232
```

zhang-alvin:
There's also a difference between VOF.uList[:] and VOF.levelModelList[0].u.dof[:] as well because of the call to setMassQuadrature() in postStep():
https://github.com/zhang-alvin/proteus/blob/master/proteus/mprans/MCorr.py#L1140.

wacyyang:
@zhang-alvin Here are several remarks on your concern
[...]

PR #785, opened 2018-08-07
level-set ebqe[('grad(u)',0] was not set in NCLS.h

zhang-alvin:
Odd behavior was observed when looking at #766. The behavior was such that `self.lsModel.ebqe[('grad(u)',0)]` was all zeros in the first call to `MCorr.postStep()` for a RANS2P, NCLS simulation:

```
(Pdb) self.lsModel.ebqe[('grad(u)',0)].min()
0.0
(Pdb) self.lsModel.ebqe[('grad(u)',0)].max()
0.0
```
```
(Pdb) where
  /home/alvin/Projects/proteus/proteus/linux2/bin/parun(6)<module>()
-> exec(compile(open(__file__).read(), __file__, 'exec'))
  /home/alvin/Projects/proteus/proteus/scripts/parun(474)<module>()
-> runName + '_run_prof')
  /home/alvin/Projects/proteus/proteus/proteus/Profiling.py(190)__call__()
-> return func(*func_args, **func_kwargs)
  /home/alvin/Projects/proteus/proteus/proteus/NumericalSolution.py(1535)calculateSolution()
-> self.postStep(model)
  /home/alvin/Projects/proteus/proteus/proteus/NumericalSolution.py(1664)postStep()
-> postCopy = levelModel.coefficients.postStep(model.stepController.t_model,firstStep=self.firstStep)
> /home/alvin/Projects/proteus/proteus/proteus/mprans/MCorr.py(174)postStep()
-> self.lsModel.ebqe[('grad(u)', 0)] += self.massCorrModel.ebqe[('grad(u)', 0)]
```
Subsequent calls to `MCorr.postStep()` showed that the `self.lsModel.ebqe[('grad(u)',0)]` data structure wasn't being modified anywhere else but in `MCorr.postStep()`. 

This PR aims to correct this behavior by setting `self.lsModel.ebqe[('grad(u)',0)]` in `NCLS.h` as the normalized gradient of the level-set. The result is that in the first call to `MCorr.postStep()` the values of `self.lsModel.ebqe[('grad(u)',0)]` are no longer zero.
```
(Pdb) self.lsModel.ebqe[('grad(u)', 0)].min()
-0.017362136439284113
(Pdb) self.lsModel.ebqe[('grad(u)', 0)].max()
0.9999999917611154
```
To my knowledge, this array is really only used in `https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS2P.py#L1588`, which is used to compute the length scale for the residual calculation pertaining to exterior element boundaries:
`https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS2P.h#L3010`

I'm not sure if something similar is necessary in `NCLS3P` , but I know @manuel-quezada has included this computation in `CLSVOF`: `https://github.com/erdc/proteus/blob/master/proteus/mprans/CLSVOF.h#L857-L861`


cekees:
@manuel-quezada could you just double check that  this is consistent with what you're doing?

manuel-quezada:
This looks good to me. @zhang-alvin and I discussed about this when he was here. The only difference I found between his changes and what I do is that he explicitly normalizes. I don't do that. If we do a decent job at redistancing then it shouldn't matter. In any case it sounds safer to do what @zhang-alvin is doing.

PR #786, opened 2018-08-10
Arnsong/fix sbm

arnsong:
### Introduction
This is an attempt to standardize the way that Proteus couples to particle codes in a manner that leaves responsibility in the appropriate places, i.e., the particle code calculates and hands off information such as the signed distance function or the particle positions and Proteus only worries about calculating the fluid solution with this information.  This is a work in progress, so any suggestions are welcome. 

### Implementation
This API contains only 3 functions to facilitate information handoff between Proteus and the particle solver. We access these function through the particle solver object which is passed to the `RANS3PF.Coefficients` object.  In this document, we will refer to the particle solver object as `particles`
- `particles[i].x()` -- Return the centroid position of particle **i**
- `particles[i].sdf(position)` -- This calculates the sdf for a particle **i** at **position**.
- `particles.updateSDF(...)` -- Vectorized sdf calculation for at Proteus nodes and integration points

The first two functions should be straightforward to implement in the particle solver. The vectorized `updateSDF` function is a little more involved.  In theory, another particle code should be able to take the code that we use to expose our c++ backend to python and merely replace our headers with theirs and link to their libraries.  Theoretically... There is a little more work to be done on the C++ side to make this a reality, but very doable.

### DIFF
--- a/proteus/mprans/RANS3PF.py
+++ b/proteus/mprans/RANS3PF.py
@@ -213,7 +213,7 @@ class Coefficients(proteus.TransportCoefficients.TC_base):
                  particle_beta=1000.0,
                  particle_penalty_constant=1000.0,
                  particle_nitsche=1.0,
-                 particle_sdfList=[],
+                 particle_sdfList=None,
                  vos_function=None,
                  particle_velocityList=[],
                  granular_sdf_Calc=None,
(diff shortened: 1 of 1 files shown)


cekees:
@arnsong this looks good. The particles object also needs a `size` member function, correct?

arnsong:
@cekees Oops.  Yes, the particles object needs a `size` member function.  At the moment, the centroid update isn't as clean as I would like it to be, i.e., it's handle outside of the particle solver currently.  If you look at the argument list for `updateSDF`, you'll see that I was hoping to update the centroid locations completely on the particle side. That being said, once I've fixed that, we'll still need the `size` function for initializing the `particle_centroids` vector.

PR #788, opened 2018-08-17
MeshAdapt via Banded Interface

zhang-alvin:
PR to merge in banded interface based adapt and initial adaptation of mesh prior to solve.
This includes corrections to lagged subgrid error and shock capturing post-adapt during the "restart" phase for bdf1 time integration. The restart is exact for a static mesh and when redistancing and mass correction is not applied. This is because Proteus applies RD and MC through the levelModel.u.dof and quadrature point fields and not from the modelList.u field (see #766)

There are still some issues remaining with the bdf2 case even without adapt so proper solution transfer will be addressed later for that scenario.

Subsequent posts will show some results.

To-Do:

- [x]  `--memory` flag is needed at the moment which suggests a memory leak of some sort
- [x] Documentation
- [ ] Improve naming schemes
- [ ] CFL-based sizing of the interface band
- [x] Get tests to pass
- [x] add 2D dambreak and cylinder tests to check restart procedure


zhang-alvin:
Feature branch was rebased on top of `master` to simplify the review process.

PR #790, opened 2018-08-21
CLSVOF lambda fix, scales properly with domain size

jhcollins:
mean_distance wasn't being calculated properly which prevents proper scaling with domain size.


manuel-quezada:
It looks good to me but I am guessing that some tests will be broken. Let's wait for Travis to let us know the problems.

manuel-quezada:
@cekees I am aware of this PR and the bug it is fixing. Or are you talking about a different scaling issue? Btw, I think the PR is good to go.

jhcollins:
@cekees
Yes, Manuel has gone over scaling with me. The bug was discovered after Manuel’s troubleshoot. I would not have found it on my own. 

We also thought it would be a good moment to familiarize myself with pull requests and Travis

PR #791, opened 2018-08-22
Yy/sbm/asserterr

wacyyang:
In this PR, the following changes are made:

- simplify the computation of `middle_point_coord `
- `dist` may be negative
- replace `assert(ebN>=nElementBoundaries_owned)`
- update `.h5` files


cekees:
the tests are failing because the hdf5 file comparison files don't have the same number of quadrature points as what is being run. Run git diff ane make sure you have no changes locally.

wacyyang:
@cekees It looks weird. I am not sure what is wrong here. This is what I get on my laptop. 
screenshot 2018-08-22 21 07 07

cekees:
what changes did you make to the Makefile?

wacyyang:
It is the definition of `FC` 
screenshot 2018-08-22 21 44 31

cekees:
It appears that the mesh generated on your machine and the testing image are different.

PR #796, opened 2018-09-04
Switch from hand-coded wrappers to cython for canalyticalSolutions

cekees:
- [x] finish test module
- [x] check  documentation


cekees:
This is finished but merging into #810 because it requires some fixes from that branch.

PR #800, opened 2018-09-06
Mquezada/fixing clsvof

manuel-quezada:
PR to include pre-redistancing into the CLSVOF model. I also consider now a single stage time stepping scheme to keep the cost roughly the same.


cekees:
@tridelat  We're getting 

TestCable::testHangingCableANCF FAILED

```
       T_sol = -np.ones(3)*g*rho*(np.pi*d**2/4.*L)
>       npt.assert_almost_equal(-T, T_sol)
linux2/lib/python2.7/site-packages/proteus-1.4.2-py2.7-linux-x86_64.egg/proteus/tests/test_mbd_chrono.py:34: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
x_id = array([ True,  True,  True]), y_id = array([False, False, False])
hasval = 'nan'
    def chk_same_position(x_id, y_id, hasval='nan'):
        """Handling nan/inf: check that x and y have the nan/inf at the same
            locations."""
        try:
            assert_array_equal(x_id, y_id)
        except AssertionError:
            msg = build_err_msg([x, y],
                                err_msg + '\nx and y %s location mismatch:'
                                % (hasval), verbose=verbose, header=header,
                                names=('x', 'y'), precision=precision)
>           raise AssertionError(msg)
E           AssertionError: 
E           Arrays are not almost equal to 7 decimals
E           
E           x and y nan location mismatch:
E            x: array([nan, nan, nan])
E            y: array([-0.       , -0.       ,  0.0385238])
```

Any ideas? Did we need to update the chrono version in the stack when we did the recent update to MBD?

manuel-quezada:
@cekees and @tridelat indeed that particular test passes sometimes and fails others on my computer. This is true not just on this particular branch but also in the master (and updated) branch. I just ran it 10 times (using the master branch) and it failed 3 times.

tridelat:
@cekees @manuel-quezada yes, it looks like the problem is still there unfortunately. I thought the issue was solved with #798 but apparently not.. The stack does not need to be updated as long as chrono is on version 3.0.0 (which is the case since 1+ years on stable/proteus of hashstack). If Travis builds everything from scratch, and the nans keep appearing sometimes, we should skip that test for now

PR #801, opened 2018-09-10
Mquezada/two phase flow

manuel-quezada:
Non intrusive PR that acts as first step towards an app oriented set of files for two phase flow. 

- [x] Add support for rans2p with clsvof and have first running example.

- [x] Being able to pass arbitrary names to the different problems. 

- [x] Pass different options (e.g. -C, -D, etc) to the main script. 

- [x] Make sure it works with adaptivity. 

- [x] Add documentation file. 

- [x] Add asserts to make sure the problem is setup correctly. 

- [x] Add support for rans3p. 

- [x] Include some 2D and 3D two phase flow examples from the paper. 

- [x] Manage parallel runs. 

- [x] Create and pass PETSc options file.

- [x] Add automated tests within proteus/tests


manuel-quezada:
Hi @cekees and @tridelat. I think this PR is ready to go. I was able to make automated tests by just calling parun. I still didn't test the gauges but that is coming in the next PR by Tristan. I also added a small documentation file that explains how to set up and run a problem.

manuel-quezada:
I changed the default value of the flag `useSuperlu` and I forgot to update the files to compare. This is fixed now.

cekees:
@tridelat could you do a PR of your branch onto this one? I don't want to merge this yet and need some time to review it further. If @manuel-quezada merges your PR onto this branch with the "Rebase and merge" then all your commits will be preserved on this branch. If you are doing some tweaking to the Python usage that will save me some time on the review of this branch anyway.

manuel-quezada:
In my opinion these two PRs are non-intrusive in the sense that even though they are big most the code is in new files. Indeed, in this PR there are 50 files involved out of which 47 are new! I only touched:

- parun to include the TwoPhaseFlow flags. 
- setup.py to do trivial changes for testing and so on. 
- NumericalSolution.py to make sure we can run adaptivity by just passing the --pumi flag.

In any case I propose we close this PR and concentrate only on #806. Indeed, @tridelat branched out from my branch so most of the code is there anyway.

tridelat:
@cekees @manuel-quezada if we merge #806 in this PR, the final squash merge into master will lead to a loss of history, contributors, and blame, and I think this happened before in some previous PRs. (merging this PR to #806 and then squash merge #806 into master would lead to the same problem).
Rebase and merging would still ultimately lead to this problem because of the --squash into master.
Maybe it is not such a big deal but we should probably be a bit careful about this when several persons are working on (and branching out from) the same feature.

In any case, I just resolved conflicts on #806 so it can be merged into this one if it is more convenient, which was the original plan until #806 got bigger than expected. (some of the TwoPhaseFlow/tests might be broken after merging though due to some syntax changes)

manuel-quezada:
@tridelat if we merge this branch onto your branch then the commits and history that would be lost are from my branch. I am OK with that. I expect there would be conflicts during the merge since we do somethings slightly different. During the merge we can favor your changes if you feel you are taking better advantage of Python usage. After that we can close this PR.

tridelat:
I created a new branch called `TwoPhaseFlow` where we can squash merge this PR, and then squash merge #806, to finally non-squash merge `TwoPhaseFlow` into `master`, which would result in only 2 commits. Let me know if that works for you, I think it is cleaner and keeps proper history, blame, and contributions.

PR #802, opened 2018-09-10
checks involving arTime string and str(t) no longer match

cekees:
changed str(t) to match the way we format arTime:

"{0:e}".format(t)

This appears to be the regression that was causing the archive not to contain all the fields.


cekees:
@tridelat I turned  on skip for the ancf cable test that is failing. @manuel-quezada said it looks like we have had this problem before and had been skipping the test.

PR #805, opened 2018-09-14
Adimako/sed cleanup

adimako:
Finalising existing test cases and cleaning up


adimako:
@cekees I will merge it later on. There are some issues with the bc tests

PR #806, opened 2018-09-17
TwoPhaseFlow - adding models to utils and various changes

tridelat:
This PR is for merging into `mquezada/TwoPhaseFlow`

Main changes: 
- [x] Parameters in main class
- [x] Parameters for different models in different subclasses
- [x] Additional models: vof, ncls, rdls, movemeshmonitor, addedmass
- [x] Indexing of models directly from parameters instead of hard coded order in `_so` file
- [x] now call `parun [main_file].py --TwoPhaseFlow` instead of `parun TwoPhaseFlow_so.py --TwoPhaseFlow`, allowing for any name for the main file while still using `utils.TwoPhaseFlow_so.py` as so file
- [x] Parameters attributes are frozen (error thrown if wrong parameter name set externally), with still the option to add a non-existing attribute (`addOption`)
- [x] Add ls_consv model
- [ ] minimise arguments for initialising TwoPhaseFlowProblem class (e.g. mesh arguments could be set after initialisation)
- [x] outputStepping as a class rather than dictionary variable in TwoPhaseFlowProblem
- [x] FESpace as a class rather than dictionary
- [ ] add general options in another Parameters class


proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -39,10 +42,12 @@ def __init__(self,
                 assert type(nnz)==int, "Provide (int) nnz"
         assert domain is not None, "Provide a domain"
         assert triangleFlag in [0,1,2], "triangleFlag must be 1, 2 or 3"
-        assert type(initialConditions)==dict, "Provide dict of initial conditions"
```
manuel-quezada: If `initialConditions` is `None` I would like the code to break and let me know that I need to provide them. The same is true for the `boundaryConditions`

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -79,9 +87,10 @@ def assert_initialConditions(self,ns_model,nd,initialConditions):
         assert 'vel_v' in initialConditions, 'Provide vel_v in ICs'
         if nd==3:
             assert 'vel_w' in initialConditions, 'Provide vel_w in ICs'
-        assert 'clsvof' in initialConditions, 'Provide clsvof in ICs'
-        if ns_model==1: #rans3p
-            assert 'pressure_increment' in initialConditions, 'Provide pressure_increment in ICs'
+        if ns_model == 1:
```
manuel-quezada: Are you trying to make the use of `clsvof` to be linked with rans3p? and to use the other level set with rans2p? I am not opposed to that but I think in general it would be better to introduce a parameter called `ls_model` that is equal to 0 and 1 depending on which model you want to use (clsvof or ls+redist+vof+mcorr). If we do that then we have `ns_model` and `ls_model` so one can use either of the four possibilities. They all work with each other. After that we can decide what is set by default. What do you think?

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -79,9 +87,10 @@ def assert_initialConditions(self,ns_model,nd,initialConditions):
...
+            assert 'clsvof' in initialConditions, 'Provide clsvof in ICs'
+            if ns_model==1: #rans3p
```
manuel-quezada: I think there is an indentation error here.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -120,12 +129,14 @@ def __init__(self,
                  final_time,
                  dt_init=0.001,
                  dt_output=None,
-                 nDTout=None):
+                 nDTout=None,
+                 dt_fixed=None):
```
manuel-quezada: Can you remind me what is `dt_fixed` for? I think it has the same use to what I call `dt_output`; basically how often in real time units you want to output the solution. Is this what `dt_fixed` is for?

manuel-quezada:
Here are more comments: 

- Thanks for following my formatting style on the _p.py and _n.py files in the models you added. It doesn't have to be my style (so feel free to change it if you like something better) but I definitely prefer all of the models to have the same style, whichever it is.  

- I like the way you handle the parameters.

- I like the idea of not hard coding the indexes. It might not be necessary since the number of combinations that they can have is small and we know them a priori (depending on the value of ns_model, ls_model and if the mesh is moving or not). But I do agree it is cleaner your way. However, I don't like that they have to be provided by the user. I want the least experienced user to only have to define a domain, a set of initial conditions and a set of boundary conditions. Nothing more. If that is all they provide, I want to us to make decisions on what will be run by default (rans2p, rans3p, clsvof or not, parameters, etc). Then, if someone wants something different we can allow that to happen (and we need to document how to do it). The idea of splitting operator is already a numerical method concept/idea. I don't want to require any knowledge on numerical methods. At least for the default behavior. What do you think? 

- Ineed, following the spirit of the last comment, I like a lot the way you set the BCs (e.g. `tank.BCs[0].setFreeSlip`). Maybe we should also set a common set of initial conditions (e.g. rest, damBreak, etc). This allow the possibility of requiring less knowledge (at least for common problems).

- I like how you automatically select the spatial tools BCs and I do agree we should allow the possibility of changing them if desired. You didn't do this for rans3p, right? Maybe we should try to do it as well for consistency between the models? 

- I like that we can pass different names for the problems. That was in my check in list indeed. 

- I will do a small and short pdf file as documentation on how to use all of this once we have the first PR ready to get merged in. Then we can change it as we see fit.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: `dt_fixed` is the maximum time step to be taken, but not for archiving (which I think is what `dt_output` was here, but correct me if I am wrong) and can be used with CFL. At each time step, `dt` will be whatever is smaller between the time step respecting the CFL condition and `dt_fixed`

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: You are right, it should not be there as it can be used with any `ns_model`. I think modifying the `initialConditions` assertions here would be good, because we could use `'clsvof`', '`ncls`' or even no model for the free surface. We could check that all the models that are included in the Problem class have their respective `initialConditions`

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: Yes, I think an `ls_model` parameter is a good idea and would be useful to choose and switch between the 2 models easily

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: For `boundaryConditions`, I removed this check because if SpatialTools is used, then there is no need to provide `boundaryConditions`. I can modify the assertion to check that it is not `None` or that the domain is using SpatialTools (and therefore it is okay to be `None`)

manuel-quezada: The thing about the `boundaryConditions` is a good idea. So should we put back the assertion of `initialConditions` being not None? The way you have it now it will not break if you don't provide `initialConditions`, right?

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -73,44 +86,106 @@ def __init__(self,
...
+                self.Parameters.Models.ncls.index = ind
+                ind += 1
+                self.Parameters.Models.rdls.index = ind
+                ind += 1
+                self.Parameters.Models.ls_consrv.index = ind
+                ind += 1
+        elif self.ls_model == 1:
+            self.Parameters.Models.clsvof.index = ind
+            ind += 1
+            if self.ns_model == 0:
```
manuel-quezada: Is there an indentation error here? I think this line and its corresponding block should happen independently of the value of `self.ls_model`.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
manuel-quezada: Understood and yes, `dt_output` is for archiving.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
```diff
@@ -73,44 +86,106 @@ def __init__(self,
...
+                assert 'clsvof_DFBC' in boundaryConditions, "provide clsvof_DFBC"
+            if ns_model==1: #rans3p
+                # check dirichlet BCs
+                assert 'pressure_increment_DBC' in boundaryConditions, "Provide pressure_increment_DBC"
+                # check advective flux BCs
+                assert 'pressure_increment_AFBC' in boundaryConditions,"Provide pressure_increment_AFBC"
+                # check diffusive flux BCs
+                assert 'pressure_increment_DFBC' in boundaryConditions,"Provide pressure_increment_DFBC"
+        else:
+            assert self.domain.useSpatialTools is not None, 'Either define boundaryConditions dict or use proteus.mprans.SpatialTools to set Boundary Conditions and run function assembleDomain'
```
manuel-quezada: I think this is what you meant before, right? Basically to allow the user not to define `boundaryConditions` and assert that `SpatialTools` is being used.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
manuel-quezada: Btw, is this way of controlling the outputting convenient for you guys?

proteus/TwoPhaseFlow/models/rans2p_p.py
```diff
@@ -50,9 +57,11 @@
 # ************************************ #
 # ********** MODEL INDEXING ********** #
 # ************************************ #
-CLSVOF_model=1
-VF_model=None
-LS_model=None
+ME_model = mparams.rans2p['index']
+assert ME_model != None, 'rans2p model index was not set!'
```
manuel-quezada: Change to `is not None` instead of `!=`. We decided some time ago to use that convention. If I recall correctly.

manuel-quezada:
Ok. I looked at the changes and they look good. Here are some few more comments: 

- We definitely should share sections between models for setting up `nnx` and so on. We don't need to do that now but let's keep it in mind. 

- I saw how you handle the indexing automatically now. I think there is a bug wrt an indentation and in the examples the indexing is still done by the user. We can fix the examples later as long as we have one working example where that is done automatically. If you make sure it works for the fallingBubble then I will fix all other examples. 

- Yes, I agree that we should allow the possibility of changing the indexing if the user wants to do that. This is already possible in the way you are doing it in the last commit, right?  

- I will take care of changing rans3p and clsvof once we merge your branch into mine.

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: I think it is fine, we used `nsave` before, which is the number of save per second, but using the rate of save works too

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: Actually I forgot to put the option for `ls_model == 0`, because usually we solve vof after RANS2P, and it looks like clsvof is before in your implementation, is that right? If so, we should sometimes put free surface stuff before or after RANS

proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py
tridelat: Yes, if SpatialTools was used, then BoundaryConditions associated to the geometry boundaries will be used too. I will add a switch within TwoPhaseFlowProblem in case someone wants to use SpatialTools for the geometry but still the old way of defining BC

proteus/TwoPhaseFlow/models/rans2p_p.py
tridelat: Yes it should be `is not None`

tridelat:
- I have added a `MeshOptions` instance in `TwoPhaseFlowProblem.Parameters.mesh` but did not link it yet in the _n files

- I am not sure about this one, I forgot the NS model for `ls_model == 0` (fixed now), but is there another thing going wrong?

- Now yes

- Thanks, we could also merge this one after you merge yours into master if that one is working well

manuel-quezada:
Good! The PR looks good to me. I have no more comments to make. I checked it out and run the falling bubble test and it works fine. My branch is now working with adaptivity (thanks to Alvin's help) but I need to clean some things up. I will do that within the next days and then we can see how to incorporate both branches. I tend to prefer to merge both branches and then merge that into master but if you prefer other way that is fine as well.

tridelat:
@manuel-quezada 
Thanks for the review, there are a few small changes and cleaning up (that should not break anything) that I should probably do as well on this PR before merging. I also need to add the ALE mesh _n and _p files for the elasticity approach.

The reason I would prefer to squash merge the 2 PRs separately is that there has been more changes than first planned on this one (#806) which is derived from #801, and if you squash merge a PR you loose the history and contributions, which is usually fine when there is only 1 contributor, but is a problem with several contributors, especially for the blame (who wrote what) that also gets lost. I think this happened a couple of times in squash merged PRs before. Squash merging those 2 PRs separately would still give a clean history (only 2 commits in master). Let me know if that makes sense, whichever option is preferable is fine with me in the end.

manuel-quezada:
All right. It makes sense to me. Before trying to get my PR into master I want to be able to pass different names to the problems. The reason is because if I can do that then the automated tests are trivial (see [https://github.com/erdc/proteus/blob/504db08506629371abf1db24ed747567671d6fad/proteus/tests/SWFlows/test_SWFlows.py] where I copied your approach). If it is OK with you I can incorporate that into my PR directly and try to get it merged into master. After that you can make a PR for your branch into master.

PR #808, opened 2018-09-21
Changed velocity equations for solitary waves in WaveTools.

bchambers28:


bchambers28:
The test_wavetools equations have been updated and a reference has been added.

adimako:
@bchambers28 this has gone in. Thanks for spotting this

PR #813, opened 2018-09-27
pytables 3.4.2: syntax

tridelat:
@cekees 
Rebuilding a stack from scratch on hydra gave me errors about `File.createArray`, which should be `File.create_array` in python 3 syntax.
see https://www.pytables.org/MIGRATING_TO_3.x.html

Strangely, I could not reproduce the error on an Arch Linux installation (I did not rebuild the stack from scratch however, but it also has pytables 3.4.2)

### DIFF
--- a/proteus/Archiver.py
+++ b/proteus/Archiver.py
@@ -84,7 +84,8 @@ class AR_base(object):
                                  "r")
             else:
                 xmlFile_old=open(os.path.join(self.dataDir,
-                                              filename+str(self.rank).decode()+".xmf"),
+                                              "{filename}{rank}.xmf".format(filename=filename,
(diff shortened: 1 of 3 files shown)


cekees:
I think the issue may be that  something is wrong with h5py on that machine. The pytables code in Archiver is the old way of doing parallel writes to hdf5 and only happens if you can't import h5py. I was actually getting ready to eliminate pytables as dependency of proteus.

tridelat:
@cekees That makes sense as it would not hit that part of the code when testing with my installation on Arch Linux. I am not sure what is wrong with the h5py install on hydra (centOS) as I can use it normally it seems. Closing for now

PR #815, opened 2018-09-27
Alistairbntl/velocity block pc

alistairbntl:


alistairbntl:
@cekees - I think this is good to go in.  I could add a test to increase the diff coverage, but I don't think it would be that valuable.  This PR is really just small maintenance stuff anyway.

PR #818, opened 2018-10-12
TwoPhaseFlow: 1.0

tridelat:
@cekees @manuel-quezada new PR for the TwoPhaseFlow module with the 2 squash commits from #801 and #806 for review. Ideally this PR should not be squash merged, to keep history, contributions and blame.

This PR is mostly non-intrusive as it mostly creates new file.

EDIT: should be squashed and merged due to conflicts with diverging PRs (#921), and because #921 committed to master most changes from @manuel-quezada already

TO DO (future work / post PR):
- [ ] `OutputStepping` could be attached to `TwoPhaseFlowProblem` instead of passed
- [ ] reduce arguments for initialization of `TwoPhaseFlowProblem`
- [ ] rename `TwoPhaseFlow` to `MultiPhaseFlow`? (as it can use rans3p for example)


cekees:
@tridelat and @manuel-quezada what is it that you think squash and merge does to the history, contributions, and blame?

tridelat:
@cekees from what I gathered, and correct me if I am wrong, squash and merge erases all history and contributors of the branch to only associate it to one person, removing the blame and associating every changes to that person only in the process. I think it makes sense when there is only mostly one contributor as it allows for a cleaner history for one feature, otherwise it is nicer to keep track of who did what in case of questions/modifications in the future (using blame for example) by squash merging per contributor or per feature for example

manuel-quezada:
Here is a way to change authorship of commits:

https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit

We can merge this and change to the correct authorship for the main 2 commits.

manuel-quezada:
Personally I don't care about that issue but I understand that it is cleaner and important so I respect whatever route you guys want to follow. I honestly just want this PR to be finished. I am getting frustrated about it not being finalized even though it is in a working stage from some time ago already. And even more, even though it is absolutely non intrusive so it won't affect anybody's work.

cekees:
@manuel-quezada the authorship on the commits is correct and will remain so as long as I merge the PR via either a regular merge commit or rebase. What @tridelat is getting at is that if we use the squash and merge workflow then it would make him the author of all the commits on the new squashed commit.  The way this has been resolved is to open a 3rd branch and PR with both authors' original branches squashed and merged into the new branch. Then we just need to be careful not to use the squash and merge approach with the 3rd branch in order to preserve the individual authorship.

cekees:
@tridelat @manuel-quezada and @jhcollins there are a couple of test failures on the two-phase tests after merging in master in preparation for  finalizing this.  

[gw0] [ 47%] FAILED proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_fallingBubble 
[gw0] [ 47%] FAILED proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_marin 
[gw0] [ 48%] FAILED proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_quiescentTank 
[gw0] [ 48%] FAILED proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_risingBubble 
[gw0] [ 48%] FAILED proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_TwoDimBucklingFlow

manuel-quezada:
I checked the five tests on my laptop using a more refined mesh and larger time and the results look fine to me. I am not surprised at all that these tests are failing since there have been slight modifications in the methods since the PR was originally created. I updated the h5 files based on the new code in master. However, it seems the tests are still failing. I don't understand why that is!

cekees:
OK @manuel-quezada it may be a difference in  your  stack, particularly the PETSc update, which may also change the version of  superlu. I will run the tests on my machine, and send you  some results before  pushing reference files  from my machine, which should be very close to the travis environment. Glad to know the the proteus branch itself doesn't seem to be the issue.

manuel-quezada:
Sounds good! btw, this branch allows the user to pass the flag --pumi to run the simulations with mesh adaptivity. Most of this is happening in NumericalSolution.py and a little bit in parun. When I originally created the PR such option was working. I think it is worth asking Alvin to take a look at it and make sure it still works. The intention (and it was working that way) is that the user only has to pass the flag --pumi to turn it on.

manuel-quezada:
Another comment is about the choice of the default parameters. Since I barely use(d) rans2p I won't say much about it other than the fact that the current default value for `timeDiscretization` is `None`, which looks weird to me. Wrt to rans3pf, I would suggest that we change `USE_SUPG=False` and `ARTIFICIAL_VISCOSITY=3` which are the values we use in the paper with Haydel. Also for rans3pf, we should include `INT_BY_PARTS_PRESSURE=1`, which I think should be used for the surface tension implementation to make sense.

tridelat:
@manuel-quezada, when `timeDiscretization = None`, it applies backward Euler, but I agree that it would be better to have something that indicates this more explicitly

cekees:
@manuel-quezada while it's not the issue causing the test failures, there is an issue in a lot of the new classes that is causing problems, which is the use of large local arrays allocated on the stack instead of on the heap, as in here: https://github.com/erdc/proteus/blob/TwoPhaseFlow/proteus/mprans/CLSVOF.h#L2274
I think we should change these to std::valarray<double>, probably as class data and use the resize() member function of valarray to allow for the possibility that NNZ or NUMDOF changes.Variables allocated on the stack should be scalars or small arrays that fit in cache, whereas these are the size of the mesh.

cekees:
@manuel-quezada I pushed a change that turns up the log level and uses superlu for the linear solver and no linesearches in Newton. This is in order to figure out what is going with the tests. The reason the tests aren't reproducible is that the clsvof method is not converging. Switching to superlu and no linesearches yields convergence for some of the tests, but several are still stagnating in the nonlinear solver around 1e-4. I think generally we shouldn't use petsc linear solvers in tests without detailed petsc option settings, because the default behavior solver behavior can change when petsc changes.  I'm not sure what is happening with the nonlinear solver stagnation, however. Can you check if the solvers are converging with these new settings?

manuel-quezada:
@cekees, in my opinion it is better to have `useSuperlu=False` as default and change it in the particular problems. I changed that in TwoPhaseFlowProblem.py and pass `useSuperlu=True` for the setup of all the tests. Indeed, even after your change some tests were using petsc since `useSuperlu=False` was passed for some of them. Let's better be clear in the setup of all the problems. Now all tests are indeed using superlu.

All the tests get to the final time for me. Moreover, for the five tests, I changed the final time to `T=1.0` and all five simulations finished. I looked at the solutions in paraview and they look fine to me. Maybe you had problems since you were still using petsc and didn't pass the petsc options. I have to admit that I am a bit surprised that the risingBubble problem worked with useSuperlu=True (recall that for this problem the null space of the pressure increment equation is not empty). 

I commented `maxLineSearches=0`; i.e., I allow line searches. I think this is more robust for some problems. What is your reasoning behind this change?

In the passing, I updated all .h5 files with `useSuperlu=True` for all tests. These tests are now passing in Travis but there is an extra one that fails. It is for something not related though.

manuel-quezada:
There is an unrelated test failing. I am not sure why that is.

cekees:
@manuel-quezada  I think you may not have git-lfs installed properly so your `git add` is polluting the repository with full h5 files. Can you try installing git-lfs and remove your last commit from the repository? See e.g. https://stackoverflow.com/questions/448919/how-can-i-remove-a-commit-on-github . I think @zhang-alvin and @tridelat may have some experience doing this as well. Since it's just your last commit and I'm the only person likely to be working this branch, it shouldn't be a problem.

Fast-forward
 proteus/TwoPhaseFlow/TwoPhaseFlowProblem.py                       |   2 +-
 proteus/TwoPhaseFlow/models/clsvof_n.py                           |   2 +-
 proteus/tests/TwoPhaseFlow/TwoDimBucklingFlow.py                  |   3 ++-
 proteus/tests/TwoPhaseFlow/comparison_files/TwoDimBucklingFlow.h5 | Bin 131 -> 528563 bytes
 proteus/tests/TwoPhaseFlow/comparison_files/fallingBubble.h5      | Bin 131 -> 133654 bytes
 proteus/tests/TwoPhaseFlow/comparison_files/marin.h5              | Bin 131 -> 115350 bytes
 proteus/tests/TwoPhaseFlow/comparison_files/quiescentTank.h5      | Bin 131 -> 369719 bytes
 proteus/tests/TwoPhaseFlow/comparison_files/risingBubble.h5       | Bin 131 -> 859083 bytes
 proteus/tests/TwoPhaseFlow/fallingBubble.py                       |   3 ++-
 proteus/tests/TwoPhaseFlow/marin.py                               |   3 ++-
 proteus/tests/TwoPhaseFlow/quiescentTank.py                       |   3 ++-
 proteus/tests/TwoPhaseFlow/risingBubble.py                        |   5 +----
 12 files changed, 11 insertions(+), 10 deletions(-)
Encountered 5 file(s) that should have been pointers, but weren't:
	proteus/tests/TwoPhaseFlow/comparison_files/TwoDimBucklingFlow.h5
	proteus/tests/TwoPhaseFlow/comparison_files/fallingBubble.h5
	proteus/tests/TwoPhaseFlow/comparison_files/marin.h5
	proteus/tests/TwoPhaseFlow/comparison_files/quiescentTank.h5
	proteus/tests/TwoPhaseFlow/comparison_files/risingBubble.h5

cekees:
On the linesearch question, it's fine with me if you leave it on if you know of cases where it actually leads to convergence of an otherwise divergent newton iteration. I'm not sure I can point to any redistancing cases where that is true. We generally want to get things configured where we have reliable quadratic or superlinear convergence. I turned it off in these tests because they were all failing anyway, apparently due to the lack of good petsc options since switching to superlu seemed to fix the divergence.

cekees:
On superlu versus petsc in the tests. My point is that if you are going to use petsc in a test or as defaults, you need to explicitly nail down all the options and not rely on petsc defaults. See @alistairbntl 's test https://github.com/erdc/proteus/blob/master/proteus/tests/solver_tests_mprans/test_bochev_pressure_stabilization.py and others that show how to explicitly set the solver options. For problems with a constant pressure in the null space it seems safer to use a properly configured petsc solver, though it's possible that superlu has put in a simple fix for the special case for a one-dimensional  constant null space.

manuel-quezada:
@cekees: 

- I removed the last commit as requested. I am going to let you handle the update of the h5 files.
- About the line search, I remember having more robust behavior with line searches when I was re-distancing something from a discontinuous profile but I can't remember the specifics of the problem. 
- I agree on your point about needing to specify the petsc options for testing. I also agree it is cleaner to use superlu especially considering the tight tolerances in the tests.

cekees:
OK @manuel-quezada . Note that this issue also still needs to be addressed: https://github.com/erdc/proteus/pull/818#issuecomment-470776771

manuel-quezada:
@cekees, about #818, the only place where that happens is inside the FCTStep function which we don't use. It is there to deal with the section about high-order convergence for the CLSVOF paper. I suggest to leave it there for a potential improvement in the future. Namely, if we want to consider higher-order spaces then we need to do a better job (than a lumped L2 projection) to compute the vector field *q*. Doing that would require us to invert the mass matrix which comes with its problems. Off course one way to deal with those problems is to use FCT. For sake of finishing this PR (and submitting the paper) I suggest we deal with that when (and if ever) we decide to use high-order spaces.

cekees:
@manuel-quezada this is about ready, but we can't figure out why the moses test isn't passing. I checked a couple of things including the initial time step and the SystemStepExact parameter, but I'm still not about to get the tests to pass at full resolution. Can you take a look?

PR #821, opened 2018-10-15
Disc ICs for CLSVOF and cell average EV in RANS3PF

manuel-quezada:
This PR: 

- allows the CLSVOF method to start with a discontinuous initial condition by doing a spin up step to re-distance the disc. initial condition. By default this is not run unless `disc_ICs=True`. 

- change the EV implementation to be cell based (instead of pointwise). This is to increase the robustness.


manuel-quezada:
The details about the re-distancing are reported in the paper in the corresponding repository in `multiphase_via_EV/paperV2`. In the paper there is also an example and a convergence table. Here is an example. 

disc_ics_v2_phihat
disc_ics_v2_phi0

Issue #819, opened 2018-10-13
clean up macro usage in EV implementations


cekees:
We're using macros in C++ header files in a lot of places (SW2DCF.h, VOF.h,... ) where we should be using inline functions. Since those files get included in other files (e.g. the cython files), they can and do get expanded in unintended locations. This bit me on some recent compiles, so I added one fix here that can be used as an example: https://github.com/erdc/proteus/pull/810/files#diff-3f17cf93e98ac1999a572f29010247e9L29
That's not to say they all have to be replaced, but I think as  rule it's better to use inline functions for reasonably complicated functions as well as use unique, capitalized names for macros. It would also be good to document the functions and macros.

manuel-quezada:
I don't have automated tests for the SWEs so I don't want to make that change yet. I will change all the other files and make sure I didn't break any test. After that I will think about how to handle safely the SW2DCV.h file.

adimako:
@cekees @manuel-quezada does #822 solve this one

cekees:
Yes!

PR #822, opened 2018-10-15
mquezada/Change macros to inline functions 

manuel-quezada:
In this PR I will handle issue #819.


cekees:
I don't think you need (or generally would want) the "using" statements exposing the new function definitions in the header file. Since you're using the functions within the proteus namespace anyway they should already be available where they are needed.  This is my bad, since I pointed you to an example where I put  "using proteus::f"  in the header, but that was only because there were other macros that depended on the definition of "f".  Now that you've got them all defined as inline functions, I believe you can get rid of those "using" statements, which will make the code safer, as intended. Basically, "using proteus::func" in the header file is going to expose func to other things that include the header, which defeats the purpose of enclosing in a namespace.

manuel-quezada:
I see. So the whole idea is to make those functions not visible outside of their respective .h file, right? by doing `using proteus` I was making them visible again.

manuel-quezada:
I included tests for the SWEs before changing from the macros to the inline functions. Once they pass I will change SW2DCV.h as well. This implementation of SW2DCV is the most recent I have (i.e., with convex limiting). I am off course not including anything about sharing _p and _n files.

manuel-quezada:
@cekees I added the SWEs tests and changed most of the macros to inline functions within SW2DCV.h. However, there are two more I need to change but for some reason when I change those the tests don't pass. It seems the macro and the inline functions are not giving the same thing. Can you take a look to see if you spot something? I will keep investigating as well.

PR #827, opened 2018-10-16
SW2DCV.h: variable name conflict (Arch Linux build)

tridelat:
Mini PR for a conflict with a variable name when building in Arch Linux. I don't think it breaks anything


cekees:
@manuel-quezada may have mention this already but #822 will fix the name conflict. I'll leave this open until that goes in, but we'll probably just close this PR instead of merging unless for some reason you still have a name conflict.

cekees:
@tridelat see if the latest master fixes this issue

tridelat:
@cekees yes, #822 solved this, closing

PR #832, opened 2018-10-26
Cekees/fix soliton

cekees:
@bchambers28 and @adimako I reran the solitary wave problem with the original form of the soliton velocity profile, and I think it is correct. The wave speed is slightly slower and there is some oscillation and decay in the maximum free surface elevation. Anyway, putting this as a PR for discussion.
soliton_decay

soliton0001
soliton0002
soliton0062


adimako:
@once soliton tests are updated, ready to go in

PR #833, opened 2018-10-30
Add a new limiter for the VOS equation

cekees:
Adds a conservative global limiter for volume fractions that have a physical min/max. Also fixes up and refactors the VOF/VOS code. 
/cc @manuel-quezada


proteus/mprans/CLSVOF.h
```diff
@@ -878,8 +881,9 @@ namespace proteus
 		    // SAVE SOLUTION // for other models
 		    ///////////////////
 		    q_u[eN_k] = u;
-		    q_m[eN_k] = u; //porosity*u;
+		    q_m[eN_k] = u;
 		    q_H[eN_k] = Hnp1;
+		    q_mH[eN_k] = porosity*Hnp1; //porosity*H(\phi)=(1-q_vos)*H(\phi)
```
adimako: Update CLSVOF to include porosity effects (also below)

proteus/mprans/RANS3PF.h
```diff
@@ -1050,28 +1058,28 @@ namespace proteus
...
         double new_beta =   closure.betaCoeff(1.0-phi_s,
                                           rho,
                                          fluid_velocity,
                                            solid_velocity,
                                             viscosity);
         //new_beta/=rho;
-        mom_u_source +=  (1.0 - phi_s)*new_beta*( (u - u_s) - nu_t*gradC_x/closure.sigmaC_ );
-        mom_v_source +=  (1.0 - phi_s)*new_beta*( (v - v_s)- nu_t*gradC_y/closure.sigmaC_ );
-        mom_w_source +=  (1.0 - phi_s)*new_beta*( (w - w_s)- nu_t*gradC_z/closure.sigmaC_ );;
+        mom_u_source +=  (1.0 - phi_s)*new_beta*( (u - uStar_s) - nu_t*gradC_x/closure.sigmaC_ );
```
adimako: Does this mean that the calculation is updated? Or just the naming

proteus/mprans/RANS3PF2D.h
```diff
@@ -1036,23 +1058,28 @@ namespace proteus
...
 					    solid_velocity,
-					    viscosity);
+					    viscosity)*DRAG_FAC;
+        //new_beta = 254800.0;//hack fall velocity of 0.1 with no pressure gradient
+        double beta2 = 156976.4;//hack, fall velocity of 0.1 with hydrostatic water
+        
+        mom_u_source += (1.0 - phi_s) * new_beta * (u - u_s) - TURB_FORCE_FAC*new_beta*nu_t*gradC_x/closure.sigmaC_  +
+          (1.0 - phi_s)*(1.0-DRAG_FAC)*beta2*(u-u_s);
+	mom_v_source += (1.0 - phi_s) * new_beta * (v - v_s) - TURB_FORCE_FAC*new_beta*nu_t*gradC_y/closure.sigmaC_ +
+          (1.0 - phi_s)*(1.0-DRAG_FAC)*beta2*(v-v_s);
```
adimako: This is a hack, should be removed?

proteus/mprans/RANS3PSed2D.h
```diff
@@ -1,12 +1,12 @@
...
 #ifndef RANS3PSed2D_H
 #define RANS3PSed2D_H
 #include <cmath>
 #include <iostream>
 #include "CompKernel.h"
 #include "ModelFactory.h"
 #include "SedClosure.h"
-
+#define DRAG_FAC 1.0
+#define TURB_FORCE_FAC 0.0
```
adimako: Are these hacks?

proteus/mprans/RANS3PSed2D.h
```diff
@@ -1951,22 +1965,23 @@ namespace proteus
...
-					dmom_u_source,
-					dmom_v_source,
-					dmom_w_source);
 
-				               updateFrictionalPressure(vos,
-                        grad_vos,
-						mom_u_source,
-						mom_v_source,
-						mom_w_source);      
+		/* /\* updatePenaltyForPacking(vos, *\/ */
```
adimako: Agreed, remove this, not necessary if the limiter is around

proteus/mprans/CLSVOF.h
```diff
@@ -878,8 +881,9 @@ namespace proteus
 		    // SAVE SOLUTION // for other models
 		    ///////////////////
 		    q_u[eN_k] = u;
-		    q_m[eN_k] = u; //porosity*u;
+		    q_m[eN_k] = u;
```
manuel-quezada: Shouldn't this be `q_m[eN_k] = porosity*u'`?

manuel-quezada:
I think we should add the implosion tests for the VOS model.

proteus/mprans/CLSVOF.h
cekees: @manuel-quezada you checked the way you updated the porosity effects in CLSVOF, right?

proteus/mprans/RANS3PF.h
cekees: It's changed slightly to use the extrapolated/lagged u_s.  After looking at the way we were handling the momentum exchange, I changed a few things to make each sub-step linear and ensure conservation of momentum in the exchange. There are several ways to keep it conservative.  We may need to  return to this issue.

proteus/mprans/RANS3PF2D.h
cekees: I was going to leave these in for now. They are  useful in debugging changes to the formulation and numerics.

proteus/mprans/RANS3PSed2D.h
cekees: They are for debugging. The first turns off the physical drag and replaces  it with a simple global constant drag. The second turns off the turbulent force closure  relation.  I will turn the off before merging but probably leave them in the .h file for future testing.

proteus/mprans/RANS3PSed2D.h
cekees: OK, will do.

cekees:
@manuel-quezada sounds good on adding the implosion tests. See @adimako 's question about the old VOF/LS/VOS. It was my understanding that the old approach will still work and continue to be supported.

proteus/mprans/CLSVOF.h
cekees: I would think so @manuel-quezada. Based on the history, you removed the comment that suggests that, however. I think the main thing is that we be clear and consistent about which q fields store the vof and the sdf.

cekees: Looking at the whole file and how this variable is used, I actually think it should be porosity*Hnp1. This should be the conserved quantity, which is actually porosity*H(u).  I don't think we're using this variable in the actual CLSVOF method because of how the time discretization is actually implemented.

cekees: I switched this to porosity*u and tried to make the additional changes required by that.

PR #843, opened 2018-11-07
Cekees/update sed penalty

cekees:
@manuel-quezada I added an initial version of the cut cell refinement and the interpolation operators we discussed for 2D triangles.


manuel-quezada:
Good. I will look into that. We should talk with @wacyyang about it so that he and I can implement it in our corresponding problems of interest.

PR #845, opened 2018-11-20
MBD: replace cython headers by links to SWIG Chrono pyEngine API

tridelat:
This PR is for replacing (where possible) cython headers that allowed using Chrono through python with links to the Chrono pyEngine API directly, as this one became more full fledged.

This will allow users to use Chrono pyEngine directly while linking it with proteus. `ProtChBody` and `ProtChSystem` now have accessible Chrono `ChBody` and `ChSystemSMC` instances that can be used directly with pyEngine.

This only works with one of the latest dev versions of Chrono as 3.0.0 did not include `ChSystemSMC` as an accessible class through the pyEngine, and this one is necessary for FEM contacts (e.g. cables).
These changes are code breaking if the stack is not updated.

#### Changes affecting simulation setup:
- All functionalities such as `setPosition`, `setRotation`, etc for `ProtChBody` have been removed and should be set directly with pychrono syntax: e.g. `ProtChBody.ChBody.SetPos(...)`
- `ProtChSystem` does not have `gravity` passed on initialisation, this must be set with `ProtChSystem.ChSystem.Set_G_acc(...)` along with all other ChSystem options.
- `setSolverDiagonalPreconditioning` added to `ProtChSystem` as this does not seem to be feasible with the Chrono python API (so it sets it up on the C++ side)
- pyChronoCore removed -> use pychrono directly to create vectors, quaternions, etc
- `Mesh` renamed to `ProtChMesh` for consistency
- Contact materials must also be created with pychrono before being passed to `ProtChMoorings`

#### To do:
- name conflict for `ProtChMesh.ChMesh` (currently renamed `ProtChMesh.ChMeshh`) with Chrono's `ChMesh`, not sure why as this is not a problem for `ProtChSystem.ChSystem` or `ProtChBody.ChBody`


cekees:
@tridelat this is ready to go in when the tests pass.

PR #863, opened 2018-12-04
Support variably bathymetry in mprains.SpatialTools.Tank3D

cekees:
@bchambers28 @adimako  @tridelat here's a start on what we discussed on the proteus slack channel. In any case I think this provides a way forward with your work on the soliton @bchambers28 . It seem to me that for general basin geometries it's better to define the geometry directly with the PiecewiseLinearComplexDomain. 

Example script:
```
from proteus import Domain
from proteus.mprans import SpatialTools as st
import numpy as np

domain = Domain.PiecewiseLinearComplexDomain()
tank = st.Tank3D(domain, dim=[4., 4., 1.],bathyFunc=lambda x,y : x/5.)
tank.setSponge(x_p=1.0,x_n=1.0)#,y_p=1.0,y_n=1.0)
tank.setAbsorptionZones(dragAlpha=10,x_n=True, x_p=True)#, y_n=True, y_p=True)
st.assembleDomain(domain)
```

Generate test mesh with:
`tetgen -Kp mesh.poly ; pv mesh.1.vtk`


tridelat:
@cekees it works well if only `x` is defined but it looks like the absorption zone gets deformed if `y` is used in the `bathyFunc`.

@cekees @bchambers28 it is also possible to use `CustomShape` instead of `PiecewiseLinearComplexDomain` directly, to still have access to some of the functionalities of SpatialTools such as the BC structure and setting relaxation zones by region index/flag. (`CustomShape` is basically using the same arguments that a `Domain` instance would use, but can be combined with other shapes as well)

tridelat:
@cekees Just realised that my comment above does not really make sense, as the user could want to have this slope in the `y` direction of your example even if it affects one of the relaxation zones.

cekees:
@tridelat Is this worth merging once the issues are fixed?

cekees:
@bchambers28 are you using this functionality?

bchambers28:
I am not using that branch anymore.

Thank You,
Bradley Chambers

Issue #854, opened 2018-11-27
convert cnumericalFluxModule.c to cnumericalFlux.pyx


cekees:
- [x] create cnumericalFluxTest.pyx module and add forward declarations for functions in numericalFlux.h  (see https://github.com/erdc/proteus/blob/master/proteus/canalyticalSolutions.pyx#L3)
- [x] add cnumericalFluxTest extension build to setup.py (see line for canalyticalSolutions)
- [ ] add test_numericalFlux.py to proteus/tests (see test_analyticalSolutions.py)
- [x] add python wrapper functions  (see https://github.com/erdc/proteus/blob/master/proteus/canalyticalSolutions.pyx#L49)
- [x] ensure wrappers replace everything exposed from numericalFlux.h and cnumericalFluxModule.c and pass tests
- [x] rename cnumericalFluxTest.pyx to cnumericalFlux.pyx and remove cnumericalFluxModule.c and related setup.py lines

See http://docs.cython.org/en/latest/src/tutorial/external.html

PR #864, opened 2018-12-10
Get cnumericalFlux.pyx and solve #854

wacyyang:
Try to solve #854


wacyyang:
@cekees I think it is ready to merge this branch. For the test of this file, I am still debugging it, because there is some memory corruption. Maybe we can merge this branch first, and I will create a PR later for the test.

PR #866, opened 2018-12-13
Alistairbntl/convert lapack wrappers

alistairbntl:


alistairbntl:
A few comments / issues to address before merge
- [ ] *** lapack operators on the transpose of incoming matrix ***
- [ ] finalize documentation
- [ ] need to make sure `lapackWrappers` namespace still works

cekees:
@alistairbntl something seems to be preventing travis from completing the build. Don't know if it's just our project or more widespread. Can you verify that this updated branch passes the tests locally?

cekees:
Notice that you'll need to pull from github before testing, since I fast forwarded the branch for you.

alistairbntl:
@cekees - It's not clear to me just yet why this is happening.  Should be something small but I haven't found it yet.

cekees:
@alistairbntl it looks like the new shockCapturing module was failing to build because it was missing the include_dirs argument in setup.py. Not sure how that happened, but I think 9edfd40 will fix it.

Issue #851, opened 2018-11-27
convert flcbdfWrappersModule.cpp to Cython


cekees:

cekees:
Added #860

Issue #849, opened 2018-11-27
Convert cmeshToolsModule.cpp to cmeshTools.pyx


cekees:
- [x] create cmeshToolsTest.pyx module and add forward declarations for functions from cmeshTools.h, mesh.h and meshio.h  (see https://github.com/erdc/proteus/blob/master/proteus/canalyticalSolutions.pyx#L3)
- [x] add cmeshToolsTest extension build to setup.py (see line for canalyticalSolutions)
- [ ] add test_cmeshTools.py to proteus/tests (see test_analyticalSolutions.py)
- [x] add python wrapper functions  (see https://github.com/erdc/proteus/blob/master/proteus/canalyticalSolutions.pyx#L49)
- [x] ensure wrappers replace everything exposed in cmeshToolsModule.cpp and pass tests
- [x] rename cmeshToolsTest.pyx to cmeshTools.pyx and remove cmeshToolsModule.cpp and related setup.py lines

See http://docs.cython.org/en/latest/src/tutorial/external.html

PR #874, opened 2018-12-22
Refactor cmeshTools and flcbdfWrappers

cekees:
Addresses #849 and #851


proteus/MeshTools.py
```diff
@@ -1626,8 +1642,8 @@ def convertFromPUMI(self, PUMIMesh, faceList,regList, parallel=False, dim=3):
            self.elementBoundaryNumbering_subdomain2global,
            self.edgeOffsets_subdomain_owned,
            self.edgeNumbering_subdomain2global) = (
-              flcbdfWrappers.convertPUMIPartitionToPython(self.cmesh,
-                  self.subdomainMesh.cmesh))
+               cpartitioning.convertPUMIPartitionToPython(self.cmesh,
```
zhang-alvin: `convertPUMIPartitionToPython` is currently not defined in `cpartitioning` or anywhere else. It used to be in `flcbdfWrappersModule.h`. I can look into reimplementing the function which assigns python object handles that Proteus uses to the cmesh data structures created from an apf/PUMI mesh. Based on syntax, it doesn't look like `cpartitioning` is the right file to put this function anymore. I'll likely move it to `MeshAdaptPUMI.pyx`.

proteus/mprans/MoveMesh.py
```diff
@@ -493,7 +493,8 @@ def __init__(self,
         compKernelFlag = 0
         if self.nSpace_global == 2:
             import copy
-            self.u[2] = copy.deepcopy(self.u[1])
+            self.u[2] = self.u[1].copy()
```
zhang-alvin: Some of the issues I encountered with the MeshAdapt workflow in this PR stemmed from the switch to shallow copies from deep copies. In my case, the deep copies were needed to maintain a history of values, but a deepcopy wasn't doable with the changes from this PR because of some error like `self.mesh object could not be pickled`. Is that the reasoning for this line of code? And are we sure that there is a test that accounts for this?

proteus/mprans/RANS2P.py
```diff
@@ -1423,7 +1272,8 @@ def __init__(self,
             self.elementDiameter = self.mesh.elementDiametersArray
         if self.nSpace_global == 2:
             import copy
-            self.u[3] = copy.deepcopy(self.u[2])
+            self.u[3] = self.u[2].copy()
```
zhang-alvin: Same question as previous comment.

proteus/mprans/RANS2P_IB.py
```diff
@@ -1024,7 +1022,8 @@ def __init__(self,
             self.elementDiameter = self.mesh.elementDiametersArray
         if self.nSpace_global == 2:
             import copy
-            self.u[3] = copy.deepcopy(self.u[2])
+            self.u[3] = self.u[2].copy()
```
zhang-alvin: Same issue as first comment

proteus/mprans/RANS3PF.py
```diff
@@ -2160,7 +2007,8 @@ def __init__(self,
             self.elementDiameter = self.mesh.elementDiametersArray
         if self.nSpace_global == 2:
             import copy
-            self.u[2] = copy.deepcopy(self.u[1])
+            self.u[2] = self.u[1].copy()
```
zhang-alvin: again....

proteus/mprans/RANS3PSed.py
```diff
@@ -1699,7 +1535,8 @@ def __init__(self,
             self.elementDiameter = self.mesh.elementDiametersArray
         if self.nSpace_global == 2:
             import copy
-            self.u[2] = copy.deepcopy(self.u[1])
+            self.u[2] = self.u[1].copy()
```
zhang-alvin: same

proteus/mprans/SW2D.py
```diff
@@ -245,7 +245,8 @@ def __init__(self,
         self.coefficients = coefficients
         # cek hack? give coefficients a bathymetriy array
         import copy
-        self.coefficients.b = copy.deepcopy(self.u[0])
+        self.coefficients.b = self.u[0].copy()
```
zhang-alvin: same...

proteus/mprans/SW2DCV.py
```diff
@@ -418,7 +418,8 @@ def __init__(self,
         self.coefficients = coefficients
         # cek hack? give coefficients a bathymetriy array
         import copy
-        self.coefficients.b = copy.deepcopy(self.u[0])
+        self.coefficients.b = self.u[0].copy()
```
zhang-alvin: same....

proteus/mprans/MoveMesh.py
zhang-alvin: I see. Thanks for the clarification. The custom copy should be a safe switch since it's based off the `numpy` copy routine, so I no longer think it's necessary to confirm that all lines are covered by tests.
Apart from the `deepcopy` issues that I mentioned in the adaptivity workflow (which I fixed with several commits), I'm not seeing any additional errors. 

I think this is good to go.

PR #880, opened 2019-01-02
Use sdf and vdf to provide solid information for IBM in RAN2P

wacyyang:
Some extensions are added into RANS2P:
- [x] Use sdf and vdf to provide solid information
- [x] Add pseudopenalty method for 2D IBM
- `use_pseudo_penalty>0` pseudopenalty method: use solid velocity in the discretization of the time derivative 
   - `==1` treat nonlinear term implicitly ONLY at points inside the solid domain
   - `==2` treat nonlinear term explicitly
   - `==3` turn off the nonlinear term
```
         [1] 10.1016/j.apnum.2007.04.011
         [2] 10.1017/jfm.2018.582
```
- `use_pseudo_penalty<=0` IBM
   - `==-1` turn off time derivative inside the solid
   - `==-2` add a term due to the global momentum conservation
- [x] Add acceleration penalty for solid
- [ ] Test them before the merge


wacyyang:
@Milad-Rakhsha Yes, you are right that `self.ebq_particle_velocity_s[ebN,kb,:]=vel(0.0,corresponding_point_on_boundary)` should not have `0.0`. This array is used in SBM and I am going to fix it maybe in this PR. As for location of updating `phi_s`, if we put it in `postStep` then we have to add those lines in `attachModel` also. I am trying to simplify this part. What do you think?

PR #881, opened 2019-01-03
Alistairbntl/update superlu wrappers

alistairbntl:


proteus/csuperluWrappersModule.pyx
```diff
@@ -0,0 +1,343 @@
...
+
+# ARB - the cython directives above help improve the code speed.  it is worth noting
+# however that these directives can lead to segfaults if the code is not implemented
+# correctly.  perhaps they should only be enabled when not in debug mode?
+
+import numpy as np
+import cython
+cimport numpy as np
+
+# cdef extern from "proteus_superlu.h"  -- ARB: need to use this or some other non-hardcoded approach ...
```
cekees: Does  using proteus_superlu.h work? Or can you just  use slu_ddefs.h and make sure the include path is on the cython extension statement for this  in setup.py?

alistairbntl: @cekees - I think using `slu_ddefs.h` will be best.  At the moment I'm working on getting the SparseFactor functionalities up and running in Cython.  It has been a little bit tricky because of all the nested data structures.  Once I've got that wrapped up I'll make sure the dependency paths are set correctly.

alistairbntl:
@cekees - hold off on merging this for a little bit.  I've got the prototypes for the `smoother` wrappers started.  Should have it wrapped up (haha) before tomorrow.

cekees:
Sounds good. I  hope to have the postprocessing finished by tomorrow too. Then we're done!

Issue #850, opened 2018-11-27
remove triangleWrappersModule.c


cekees:
I propose that  we remove this entirely instead of developing new Cython wrappers. Instead, replace it with tools that simply call the triangle executable, as we do with tetgen in 3D.  I believe we only need to reproduce the behavior of the `TriangleBaseMesh` constructor from TriangleTools, unless I'm missing other uses of TriangleTools or TriangleUtils in other proteus modules.

```MeshTools.py:    from proteus import TriangleTools
MeshTools.py:        tmesh = TriangleTools.TriangleBaseMesh(baseFlags="Yp",
MeshTools.py:        from . import TriangleTools
MeshTools.py:        tmesh = TriangleTools.TriangleBaseMesh(baseFlags=self.triangleOptions,
NumericalSolution.py:from . import TriangleTools
NumericalSolution.py:                        tmesh = TriangleTools.TriangleBaseMesh(baseFlags=n.triangleOptions,
```

tridelat:
Closing as #882 was merged

PR #882, opened 2019-01-14
removed triangleWrappersModule, TriangleTools, and TriangleUtils

tridelat:
PR for issue #850


cekees:
Looks like this almost does it. Two of @alistairbntl 's nice matrix tests are failing, but I suspect we may just need to enforce a baseline set of flags. It seems to not be producing a required edge file:  

  x86_64.egg/proteus/NumericalSolution.py:294: in __init__
      check_call("mv {0:s}.1.edge {0:s}.edge".format(p.domain.polyfile), shell=True)
  linux2/lib/python2.7/subprocess.py:190: in check_call
      raise CalledProcessError(retcode, cmd)
  E   CalledProcessError: Command 'mv reference_triangle_2d.1.edge reference_triangle_2d.edge' 
  returned non-zero exit status 1

tests/matrix_constructor/test_mass_matrix_reference_simplex.py

tridelat:
@cekees what would be preferable between enforcing the creation of the .edge file with a flag, or removing the requirement for such file (as the test case should work without it)?

tridelat:
@cekees some tests don't pass because the results are not exactly the same (but very close), and I think this has to do with decimal points describing the mesh nodes. They seem to be limited to 5 in `master`

PR #883, opened 2019-01-14
Cythonize cpostprocessing

cekees:


proteus/cpostprocessing.pyx
```diff
@@ -0,0 +1,1298 @@
...
+                                                  <int*>(skipflag_elementBoundaries.data),
+                                                  <double*>(elementResidual.data),
+                                                  <double*>(dx.data),
+                                                  <double*>(n.data),
+                                                  <double*>(conservationResidual.data),
+                                                  <double*>(vConservative.data))
+
+def calculateConservationJacobianPWL_opt(int nNodes_owned,
+                                         np.ndarray interiorElementBoundaries,
+
```
alistairbntl: @cekees - If the Cython directives `#cython: boundscheck=False, wraparound=False, nonecheck=False, initializedcheck=False` were inserted at the beginning of the file that would speed up these loops.

cekees:
Thanks @alistairbntl .  Agree on making things consistent, optimizing cython options, eliminating warnings, and further refactoring. I thought we could make those issues that folks could take  up independently once the dust settles on this project, and we've had a chance  to compare notes.

proteus/cpostprocessing.pyx
cekees: I'm going to hold off on these for  now. In most  of the legacy stuff the looping is  done in C anyway.

PR #887, opened 2019-01-23
SpatialTools: fix for Tank2D relaxation zones when translated

tridelat:
Small PR for a fix that made relaxation zones to have the wrong center when tank was translated.
The functions `setAbsorptionZones` and `setGenerationZones` should still be called after translating as it is now.


tridelat:
@cekees, I fixed the SpatialTools tests but I am not sure why the MCorr test fails

cekees:
Thanks, I'll take a look. I found some solver issues when I was updating stack. Not major problems but could produce perturbations with respect to the archived solutions on the order  of 1e-8.  I'll check  if that is the case here.

PR #893, opened 2019-01-28
Adimako/ramp in paddle

adimako:
Added ramp times for moving paddle.


adimako:
@cekees @tridelat this one fails because of time connection errors for the package patchIf. It seems like that it cannot connect to our server and get the package.

PR #894, opened 2019-01-30
Adding ramp time in paddles and test for paddles

adimako:
This is superseding #893


adimako:
@cekees, I will wait till you fix the access to stack. Just ping it in slack once ready

PR #897, opened 2019-01-31
fix option -H hotstart dataDir issue

burgreen:
One line change is not tested. I am waiting for all Python3 conversions before updating the MSU build process of the latest Proteus version.


cekees:
Thanks @burgreen

PR #899, opened 2019-02-06
Alistairbntl/update to petsc 3.10.2

alistairbntl:


cekees:
These relevant changes are all in master now, so I'm closing this. The .hashdist_default file is no longer needed and not part of the source tree.

PR #900, opened 2019-02-12
Adimako/combined waves

adimako:


adimako:
@cekees @tridelat This one should work nicely. The new class takes multiple waveTools classes and sums them up

PR #905, opened 2019-02-23
Refactoring ibm/sbm/unfitted methods

cekees:


cekees:
@Milad-Rakhsha and @wacyyang @arnsong I think the indexing change that I just merged from @wacyyang 's PR is actually incorrect. The index into the particle velocity should be eN_k_3d. I thought we had all agreed that the external DEM codes would always supply 3D sptial information (locations, velocities, normals, etc regardless of whether proteus is solving in 2D or 3D). Can you confirm? I think it's best if you guys branch of of this branch `cleanup_ib` and do PR's back onto this branch, not onto master until we have finished this cleanup and ensure that all of @Milad-Rakhsha's tests are running properly again.

wacyyang:
I just realized the bug in my last PR as pointed out by @cekees  and we should use `eN_k_3d ` and declare all variables as `eNxKx3` or change some lines of `RANS3PF2D.h`. Personally, I prefer to unifity them and use 3D information even for 2D particles. I am going to go ahead and fix this bug. Actually that is the way we deal with ball particles. What do you think, @arnsong @Milad-Rakhsha ?

cekees:
@wacyyang my point was partly that I already went with the en_k_3d convention on this branch.

arnsong:
@cekees I'm good with that convention, i.e. `eN_k_3d`.  Makes sense to me and will keep things simpler.

Issue #908, opened 2019-02-25
Small bugs to compute 3D Hessians inside the CompKernel


manuel-quezada:
This is to report what I believe are two bugs in the computational Kernel related to 3D Hessians. 

**Bug 1**
In this line: https://github.com/erdc/proteus/blob/master/proteus/CompKernel.h#L1077, instead of `[I*2+J] = 0.0;`, I think it should be `[I*NSPACE+J] = 0.0;`. Note that this is not a specialized function (to 2D problems) but one general for NSPACE. 

**Bug 2**
I think the function in this line: https://github.com/erdc/proteus/blob/master/proteus/CompKernel.h#L1111 is wrong. The function should init hess_trial to zero before doing certain computations. This is partially done; i.e., some entries are not zero out. Currently, the first loop is: 

```
    for (int j=0;j<NDOF_TRIAL_ELEMENT;j++)
      for(int I=0;I<NSPACE;I++)
	hess_trial[j*NSPACE+I] = 0.0;
```

I think it should be: 

```
    for (int j=0;j<NDOF_TRIAL_ELEMENT;j++)
      for(int I=0;I<NSPACE;I++)
        for(int J=0;J<NSPACE;J++)
          hess_trial[j*NSPACE2+I*NSPACE+J] = 0.0;
```

manuel-quezada:
Closed since this was solved within #921

PR #909, opened 2019-02-25
Fix bug in CompKernel

manuel-quezada:
This small PR is to handle issue #908


manuel-quezada:
@cekees I updated the h5 file for the test that is failing but I had the same problem as with the PR for the TwoPhaseFlow. I am not sure what is happening. Maybe I am pushing the h5 files wrongly.

cekees:
Is your local stack submodule at 50b47cb?

manuel-quezada:
@cekees sorry I missed the comment about the stack. It seems that when I changed my email in github, the configurations about the notifications got changed so I was not getting emails about them (this is also true for the #818 comment). I changed them back so I should be notified again.

PR #912, opened 2019-03-06
PR to measure speed of the code

manuel-quezada:
Small PR to measure the speed of the code. This is useful for checking how the code scales in HPCs and maybe to start getting a sense of what a good speed is and how our different methods and implementations perform. To activate this the user needs to add `measureSpeedOfCode=True` in the so file. 

The code starts measuring the speed after the first output and stops after 100 time steps or before the second output (whatever happens first). Then it writes a txt file with the number of time steps, number of processors and times. It also reports the time to solve 1 DOF for one time step multiplied by the number of processors; i.e., how much it took one processor to solve 1 DOF on one time step. If I recall correctly, based on Jean-Luc's standards back when I was in the PhD, 1E-5 was a good time. I did some experiments in 2D with the TwoPhaseFlow code using rans3p + clsvof and I got ~1.5E-5 using one processor. The time increased fast as I used more processors but I think that was (maybe) because of expected strong scaling issues.

### DIFF
--- a/proteus/NumericalSolution.py
+++ b/proteus/NumericalSolution.py
@@ -1770,7 +1770,7 @@ class NS_base(object):  # (HasTraits):
                     NDOFs=0
                     for i,mod in enumerate(self.modelList):
                         if (i in self.so.modelSpinUpList) == False: #To remove spin up models
-                            NDOFs += len(mod.uList[0])
+                            NDOFs += mod.par_uList[0].size
                     #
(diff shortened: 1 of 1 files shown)


adimako:
@manuel-quezada I confirm that 1e-5 is a good time, and it is typical for most FV solvers.

manuel-quezada:
@cekees done!

cekees:
closing, since this is addressed in #921

Issue #913, opened 2019-03-07
Pressure issues with sediment solver


adimako:
May relate to #907 

Failure of the solver to maintain hydrostatic fluid pressure within the sediment phase. This seems to be a regression as we could do this with a previous version. In the image below, you can see the pressure field with the latest version of Proteus - with Entropy viscosity schemes for the sediment phase (left panel) against an earlier simulation with the ShockCapturing scheme and an earlier version of the code (174420ee70cc965b9b1828dc4982ac4e3fd9e21e)
![image](https://user-images.githubusercontent.com/8258298/53961514-aebd3c00-40e0-11e9-8d26-48611bb34b31.png)

adimako:
@cekees @manuel-quezada this is what I get with the shock capturing scheme - running with the master version. I do not think the problem is the EV schemes but rather lies in the VOS / Pressure model. 
![image](https://user-images.githubusercontent.com/8258298/53966515-39a33400-40eb-11e9-97af-f49292a5a823.png)

adimako:
@cekees @manuel-quezada After further investigation, it seems like it is the pressure friction terms causing these oscillations. Setting the limiter below the max theoretical limits improves things. Further investigating.

adimako:
By setting the limiting fraction far enough from the theoretical packing density, the solution is much more well behaved. Note that inside the sediment, the hydrostatic pressure gradient is proportional to the weighted density of the fluid and the sediment, not the fluid phase. This is something that we need to look at, although I feel like it is more a convention, rather than a real problem
![image](https://user-images.githubusercontent.com/8258298/54121442-547aee80-43f2-11e9-86b5-8e380f6a7dde.png)

adimako:
After looking at 3PSed2D, I am suspicious of the conversion of the sed equation in nonconservative form. 

The old version (conservative, terms multiplied with vos) is here https://github.com/erdc/proteus/blob/bd0233c171bb9b122b9c27ddcd13539bacad743f/proteus/mprans/RANS3PSed2D.h

@cekees I have some questions:

- Does the hamiltonian correspond to the form of the pressure that we have in the conservative equation? E.g. if the pressure term is multiplied by vos, should this one be too?
- Same goes for the hamiltonian gradients
- Whilst the drag terms have been divided by vos to be consistent with the nc form, the pressure friction terms do not appear to be

cekees:
[...]

PR #916, opened 2019-03-12
Updated RANS3PSed module with correct vos scaling

adimako:


cekees:
@adimako thanks for catching this. Your change  looks correct  to me. Does it address #913 ?

adimako:
Nope still looking at it

PR #921, opened 2019-03-15
mquezada/TwoPhaseFlowForPaper2

manuel-quezada:
This PR is to incorporate my original TwoPhaseFlow framework into master. The PR is meant to reproduce the results in the two phase flow paper.


manuel-quezada:
@cekees, I fixed a test that was not passing. This test is for the projection scheme with edge based EV. It makes sense that it doesn't pass since I fixed the way the Hessians are computed, which are needed for such method. 

I created 6 tests for the TwoPhaseFlow work. One for each file we use for the paper. There are two that are a bit more expensive so I skipped them. If you want to check how long they take and you think they are not too long, then we can include them as well.

cekees:
Thanks for putting this up @manuel-quezada.

manuel-quezada:
@cekees as you can see this test `proteus/tests/LS_with_edgeBased_EV/MCorr/test_mcorr.py::TestMCorr::test_edge_based_EV` is also failing. This might also be related with the change in the computational kernels. Indeed, this test uses the same methods than the previous test I fixed. 
However, the test passes in my computer. I was thinking that maybe the discrepancy is very small and some difference in the stack (from Travis and my computer) is creating this discrepancy. If that is the case we should update the h5 file and maybe decrease the tolerance a little bit. What do you think?

PR #923, opened 2019-03-18
TwoPhaseFlow conflict update

tridelat:
PR for updating TwoPhaseFlow conflicts


cekees:
Looks good.  You may just want to merge into TwoPhaseFlow and then @manuel-quezada and I can review the changes relative to master. Because we're still mostly testing relative to reference data, the moses test failure  may just be a solver tolerance and roundoff perturbation.

manuel-quezada:
@tridelat that is right! I forgot that I am using a different stack than you guys. In particular, I think that my PETSc is different. Apparently, this created problems with some tests since the tolerances are very tight. Maybe you can reduce the tolerance to see if that solves your problem. The test you are having troubles with (Moses) is different from the rest since it uses an external force that must be correctly passed from the TwoPhaseFlow interface all the way to the c++ code. There is also a flag that goes with external forces to tell rans3pf that the force must be multiply by the density. This flag is needed to consider two cases with forces: convergence tests and something like the Moses test.

tridelat:
Okay I am merging in. I couldn't get the test to pass though, even when reducing the tolerance.
@manuel-quezada I moved the `forceTerms` option to a rans3p option instead of a global TwoPhaseFlowProblem option (and linked it accordingly in rans3p_p and moses) as it seemed to affect only the rans3p model, let me know if it should instead remain a global problem option. I think we should have minimal arguments to pass to TwoPhaseFlowProblem when instancing.

tridelat:
Looks like squash and merge of this branch still gives conflicts for `TwoPhaseFlow` with `master`.. Is it okay to do a simple merge of this branch and then squash and merge the `TwoPhaseFlow` branch into `master`? @manuel-quezada were most of your changes included in the previous PR?

PR #926, opened 2019-03-20
initial attempt at fixing cfemIntegrals, cnumericalFlux

cekees:
@manuel-quezada and @wacyyang it looks like some fraction of the new cnumericalFlux and cfemIntegrals wrappers are not quite right but not caught by any tests. See the diffs on this PR for examples. I found this while trying one of the old DG tests: 
`proteus/tests/poisson_2d$ parun poisson_het_2d_p.py poisson_het_2d_dgpk_n.py -l 5 -v`

It's still broken, unfortunately, so I may not have applied exactly the right fix.  We should be able to work out the corrections by looking at the old wrappers in the last release: 

https://github.com/erdc/proteus/blob/268084b0dc415bfc725e7b5f3e9bc1428bd3a139/proteus/cnumericalFluxModule.c#L1898


cekees:
The p1 IIPG method seems to be working (run with above command). That suggestions something isn't right in the interior DiffusionAdjoint terms in cfemIntegrals.pyx. The exterior terms appear to be OK. This is the term that provides "adjoint consistency" in the SIPG form.

cekees:
I added additional DG input files in `tests/dg` to do P^0, P^1, and P^2 for the rotating Gaussian hill with SSPRK and DG. I only fixed enough of the problems with the cnumericalFlux, cfemIntegrals, and ctimeIntegration wrappers to get the 2D advection and 3D advection-diffusion tests running. It would be good for @manuel-quezada @wacyyang and @ejtovar to check cfemIntegrals.pyx, cnumericalFlux.pyx, and ctimeIntegration.pyx, for the times of signature mismatch that I fixed in this PR before we merge this in, just to get this minor headache behind us.  I just went back and looked at the corresponding hand-wrapped code from 1.5.1 to find out what the argument list was: https://github.com/erdc/proteus/tree/fa592f1be42d5c2cf3c1c5053d023eb3271842a5

proteus/femIntegrals.h
```diff
@@ -738,9 +738,9 @@ extern void updateGlobalJacobianFromInteriorElementBoundaryFluxJacobian_dense(
   int *elementBoundaryElements,
   int *elementBoundaryLocalElementBoundaries,
   int *nFreeDOF_element_r,
-  int *nFreeDOF_element_u,
   int *freeLocal_r,
   int *freeGlobal_r,
+  int *nFreeDOF_element_u,
```
cekees: Note @manuel-quezada I also changed the order of these arguments to be consistent across python and C. In Python we are always using (all r stuff),(all u stuff) (e.g. nFreeDOF_element_r, freeLocal_r, freeGlobal_r, nFreeDOF_element_u,...), but the order was different in C, so the wrappers would have to catch that re-rordering. With this change there is not re-ordering going on.

proteus/ctimeIntegration.pyx
```diff
@@ -214,36 +214,30 @@ def applyDGlimitingP1Lagrange1d_withVacuumTol(int enforcePositivity,
...
-                              np.ndarray elementBoundariesArray,
+def computeElementNeighborShapeGradients(np.ndarray elementBoundariesArray,
                               np.ndarray elementNeighborsArray,
                               np.ndarray elementBarycentersArray,
                               np.ndarray elementBoundaryBarycentersArray,
                               np.ndarray elementNeighborShapeGradients):
-    ccomputeElementNeighborShapeGradients(nElements_global,
-                  					  nElementBoundaries_element,
-                  					  nSpace,
+    ccomputeElementNeighborShapeGradients(elementNeighborsArray.shape[0],
```
cekees: @ejtovar here's an example of where I had to fix the wrappers. Note I just get the integer lengths directly from the required vectors, which simplifies the Python interface and makes it less error prone.

proteus/cnumericalFlux.pyx
```diff
@@ -3310,10 +3286,10 @@ def calculateDiffusionMatrixSplittings_LDG_sd(int aSplit,
                            np.ndarray aHat,
                            np.ndarray aTilde):
      ccalculateDiffusionMatrixSplittings_LDG_sd( aSplit,
-                            nElements_global,
-                            nElementBoundaries_element,
-                            nQuadraturePoints_element,
-                            nQuadraturePoints_elementBoundary,
+					         q_a.shape[0],
```
cekees: @wacyyang here are the kind of problems that need checking for on cnumericalFlux.pyx

proteus/cfemIntegrals.pyx
```diff
@@ -3995,7 +3997,9 @@ def updateGlobalJacobianFromElementJacobian_eb_CSR(np.ndarray elementNeighbors,
                                                    np.ndarray csrRowIndeces_ru,
                                                    np.ndarray csrColumnOffsets_eb_ru,
                                                    np.ndarray elementJacobian_eb,
-                                                   np.ndarray globalJacobian):
+                                                   globalJacobian):
```
cekees: @manuel-quezada in addition to making sure the integer inputs to C are properly queried from the size of the input arrays, I also found some instances where the CSR Jacobian was declared as an numpy array. I think I caught most of these.

proteus/cnumericalFlux.pyx
wacyyang: @cekees I have compared those change to https://github.com/erdc/proteus/tree/fa592f1be42d5c2cf3c1c5053d023eb3271842a5. And the interface of those 6 functions are correct now.

PR #928, opened 2019-03-22
TwoPhaseFlow: use classes instead of _n and _p files

tridelat:
This PR is for converting the _p and _n files of each model into classes of `Physics_base` and `Numerics_base`

Parameters can be accessed this way:
```
myTpFlowProblem = TwoPhaseFlowProblem()
m = myTpFlowProblem.Parameters.Models
m.rans2p.n.nnx = 10
m.rans2p.p.movingDomain = False
m.rans2p.p.CoefficientsOptions.useVF = 1
...
```

- [x] All physics and numerics options are now printed in the log
- [x] convert _so to class as well
- [x] tests passing
- [x] new syntax for all models in Parameters.py


cekees:
Nice! We can get rid of `proteus/TwoPhaseFlow/models` now right?

tridelat:
@cekees yes, they have been replaced by classes and are not imported anymore, but some models (AddedMass, moveMesh) still need to be implemented

cekees:
I think the build is just failing because setup.py still refers to TwoPhaseFlow.models.

PR #934, opened 2019-03-27
Fix parallel runs for SW2DCV

manuel-quezada:
This PR does the following:

- Small change to NonlinearSolver.py to make sure the SW2DCV code runs in parallel. 
- Clean up the getResidual function within SW2DCV.py
- Add capability to pass a set of DOFs to constraint via a function provided by the user. This is useful for some applications (I am using it for the hydraulic jump). It can also be used to set more specialized BCs.


manuel-quezada:
I added few lines to measure the speed of the code after every output (without considering the output process). The idea is to monitor the code constantly to make sure that the speed and scaling is always what we expect and that nothing slowed down the code.

cekees:
@manuel-quezada this is ready to go in if you're finished with it.

manuel-quezada:
This PR is complete and finished w.r.t. its original intention (making sure that the code runs in parallel plus other minor things). However, just today I found out a bug in the computation of the max wave speed, which is used to compute the dij matrix. I raised issue #941. We can fix it here but it will likely break all the tests.

cekees:
Let's do another PR for the new issue.

PR #936, opened 2019-03-28
Adimako/pressure bug

adimako:
Latest branch for sediment solver and turbulent wall functions


proteus/NumericalSolution.py
```diff
@@ -1761,7 +1761,8 @@ def calculateSolution(self,runName):
                             self.systemStepController.sequenceStepTaken(model)
                     #end model split operator step
                     if stepFailed:
-                        systemStepFailed = not self.systemStepController.retrySequence_modelStepFailure()
+#                        systemStepFailed = not self.systemStepController.retrySequence_modelStepFailure()
```
cekees: I think you should be able to get this behavior with  `maxSolverFailures=1` in your _n file.

PR #937, opened 2019-03-28
AddedMass tests: tpf syntax; compare 6x6 matrix; only one phase

tridelat:
PR for modified added mass test using only one phase and comparing full 6x6 matrix.
Results are not compared to analytical solution, but to simulation results as of this commit (which are relatively close to analytical solution)


cekees:
This ready to merge but @bchambers28 @wacyyang @adimako @BranocRichards @manuel-quezada @alistairbntl @zhang-alvin it's a good example of how to clean up a two-phase flow test by collapsing all the p/n/so files into single python script.

PR #938, opened 2019-04-01
TwoPhaseFlow: hotfixes

tridelat:
Small fixes for the parameters of some models in the TwoPhaseFlow module.

- [ ] add simple test using: RANS2P, VOF, NCLS, RDLS, MCorr, and moveMesh (to be added in air-water-vv)

### DIFF
--- a/proteus/Domain.py
+++ b/proteus/Domain.py
@@ -743,7 +743,7 @@ class PlanarStraightLineGraphDomain(D_base):
         """
         Write the PSLG using the poly format.
         """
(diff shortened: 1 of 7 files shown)


tridelat:
@cekees this last commit forces the domain to write a polyfile if `writePoly` is called, not sure why this was not the case if `polyfile` was different from `None`. If a user wants to use an external polyfile, `writePoly` should not be called at all. This caused issues in some cases where the `writePoly` was called before setting `he` and called again later on with the right `he`.

PR #947, opened 2019-04-11
disabled speed check on TwoPhaseFlow app and edge based mcorr check

cekees:
We found a bug in the new speed check in the two-phase flow app wherein it gives a divide by zero for certain combinations of time step decisions. I didn't have time to fix it so I'm just disabling it by default for now. I also went ahead and disabled the test for the edge-based viscosity for the mcorr check because we can't reproduce the test failures, and I haven't been able to find any uninitialized variables or tolerance issues that might cause fluctuations in the results.

### DIFF
--- a/proteus/BoundaryConditions.py
+++ b/proteus/BoundaryConditions.py
@@ -79,11 +79,37 @@ class BoundaryCondition:
 
         Parameters
         ----------
-        value :
+        value : float
+                Constant value
 
         """
         self.uOfXT = lambda x, t: value
 
 
-    def setLinearBC(self, a0, a1, i):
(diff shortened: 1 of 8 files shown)


manuel-quezada:
I can't only remember 3 divisions: by the number of time steps, by the number of processors or by the number of DOFs. The code stops the measurement of the speed if an output is about to happen. The idea is to skip such process during the computation of the speed. Maybe if for some reason the NumericalSolution wants to output the solution without having done any step, then the number of time steps will be zero. Or maybe I am not counting the first time step after an output and you are requesting the output at every time step. That would explain the problem. But I am just guessing. Anyway, it's good you can disable it easily for the TwoPhaseFlow apps without affecting it for other things.

PR #952, opened 2019-04-12
DEBUG: playing around with travis to try fixing

zhang-alvin:
This is a dummy PR to look at how Travis performs with different environments to help identify what's going on with MeshAdaptPUMI module


cekees:
Doesn't look like this was successful, so I'm closing it for now but will leave the branch.

PR #953, opened 2019-04-15
MeshAdaptPUMI: predictive resolution for interface and combined approach

zhang-alvin:
This PR contains a number of developments for the MeshAdaptPUMI module including:

- predictive resolution for interface based adapt
- interface based adaptive trigger
- combined approach of interface-based adapt coupled with error estimator using size-field intersection
- corrected solution transfer for the system of rans2p, vof, ls, rd, mcorr using Backward Euler
- crude way of getting restart/hotstart with adaptivity and almost-exact reconstruction of the data structures.

This PR is actually continuation of #788 (around August 2018) but has duplicate commits (likely from an unnecessary `rebase` when forming the previous PR).


zhang-alvin:
Yes it was a memory leak detected by

zhang-alvin:
I've added a 2D dambreak Colagrossi case that can use the MeshAdapt workflow which uses the SpatialTools setup in `https://github.com/erdc/air-water-vv/tree/scorec/meshAdaptAniso_examples/MeshAdapt_PredictiveControl/dambreak_Colagrossi_SpatialTools`

PR #954, opened 2019-04-17
TwoPhaseFlow: PETSc Options

tridelat:
PR for adding PETSc options directly into TwoPhaseFlow app so that it becomes possible not to use a separated petsc.options file and default options are centralised.
@manuel-quezada @cekees @adimako 

- [x] add option to still only use external file
- [ ] add petsc options to:
  - [x] RANS2P
  - [x] VOF
  - [x] NCLS
  - [x] RDLS
  - [x] MCorr
  - [x] MoveMeshElastic
  - [x] MoveMeshMonitor
  - [x] AddedMass
  - [ ] RANS3PF
  - [ ] CLSVOF
  - [ ] Pressure
  - [ ] PressureInitial
  - [ ] PressureIncrement
  - [ ] Kappa
  - [ ] Dissipation


tridelat:
@cekees currently someone can either use the default PETSc options or, if they provide an option file, use that instead. Do you mean that:
1) each option should be checked and overwritten only if the exact same option is in the file (e.g. `rans2p_ksp_type`)
or
2) if options for one model are in the file (e.g. `rans2p_` prefix is found in an option), then only the options from that file should be used on that model (none of the default options would be applied), while other models that have no option in the file would still use the default options

cekees:
I think option 2 makes the most sense to me. I think if one is going to override the default options for a solver for a given model then it should be complete solver settings, but I like the idea of being able to override the solver options for just a single prefix. For example we might want to explore options for the added mass solver without having to copy paste options for every other solver as well.

alistairbntl:
@cekees @tridelat - Sorry, I'm a bit late to the party, so I apologize if my questions are out of place.  Anyway, I'm trying to see the best way to smoothly integrate the PCD options I have within the TwoPhaseFlowProblem setting.  Can you point me towards an example that uses the TwoPhaseFlowProblem class so that I can verify my approach works in the intended way?
  
Also, I'm interested in how you envisage implementing the PETSc options for multiple solvers.  With the approach you're using here Tristan, I think a conditional statement that sets the options based on the user's selected preconditioning within the petsc initialization function you've added makes the most sense (e.g if asm is selected using these settings, if PCD is selected use these etc).

Related to this, for the PCD options, I have some settings that can be used as default, but there is a good chance the user will want to change these defaults.  Do we intend to keep the petsc option files as the primary location for changing the petsc solver settings, or do we intend to allow the user to change the defaults with the numerics class?  I suppose the advantage to allowing the petsc defaults to be be changed withing the numerics class is that it will reduce the need to maintain multiple files (e.g. a simulation file and a petsc file).

alistairbntl:
@tridelat @cekees - I've pushed up some default PETSc settings for the PCD preconditioner.  As you can see there is a nonsense conditional statement on line 520 of the Parameters.py file, but I would imagine we could use this idea to cover several different preconditioning options.
I also added the `n.linearSmoother` parameter as this is where the preconditioner type could be specified by the user.  The default `SimpleNavierStokes3D` case would be consistent with the ASM options currently specified.  
Are these additions consistent with the approach being used in `Parameters.py`?

tridelat:
@cekees option 2 has been pushed

@alistairbntl I think a conditional statement is a best as well, one that lets the user choose between asm, superlu, pcd, etc
Here is an example of a test using the TwoPhaseFlow syntax:
https://github.com/erdc/proteus/blob/master/proteus/tests/AddedMass/addedmass2D.py

As it is now, the user can change the options even in the main file I think by accessing the petsc options object when calling `PETSc.Options()` (this object seems to be shared, it does not create a new instance when calling `Options()`). But if modifying the options for a model, the user would need to set all options of that model as none of the defaults would be taken otherwise.
A workaround would be to call `[my_model_parameters].initializePETScOptions()` manually in the main file to populate the options with some defaults, and then add some more or overwrite options manually. (as any subsequent call to `initializePETScOptions()` will look if the model prefix is already present in the options, it won't reset to the defaults later).

Does the `linearSmoother` `SimpleNavierStokes3D` work in 2D as well? (looks like it does as tests pass) If it should be use as default with RANS2P, we should keep it there. By the way, any options that are in the default_n/default_p files are accessible from any of the model parameters class, even if they do not appear explicitly in Parameters.py

tridelat:
@alistairbntl also, if you want to make a few additions/changes, the best might be to make another PR from this one, and merge them separately into master to keep history/gitblame (because we usually squash merge PRs).
This one could be merged into master now as it does not break anything, and that would allow you to work directly from master to add PCD and other options, let me know

alistairbntl:
@tridelat - sounds good to me, merge away.  I'm happy to work on a different branch.

PR #956, opened 2019-04-19
Release 1.6.0

tridelat:
PR for tagging release 1.6.0.
(to be merged only when fully ready to release)

### DIFF
--- a/doc/source/_templates/layout.html
+++ b/doc/source/_templates/layout.html
@@ -104,11 +104,12 @@
      style="height:1in;"> A Python toolkit for computational methods and simulation</p>
     <div class="btn-group" role="group">
       <a class="btn btn-success"
-          href="https://github.com/erdc-cm/proteus/archive/1.5.1.tar.gz">
+          href="https://github.com/erdc-cm/proteus/archive/1.6.0.tar.gz">
             <span class="glyphicon glyphicon-download"></span> Download
       </a>
       <button type="button" class="btn btn-success dropdown-toggle" data-toggle="dropdown"><span class="caret"></span></button>
       <ul class="dropdown-menu">
+        <li><a https://github.com/erdc-cm/proteus/archive/1.6.0.tar.gz">Version 1.6.0</a></li>
         <li><a https://github.com/erdc-cm/proteus/archive/1.5.1.tar.gz">Version 1.5.1</a></li>
         <li><a https://github.com/erdc-cm/proteus/archive/1.5.0.tar.gz">Version 1.5.0</a></li>
         <li><a https://github.com/erdc-cm/proteus/archive/1.4.2.tar.gz">Version 1.4.2</a></li>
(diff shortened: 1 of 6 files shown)


zhang-alvin:
I know that I was working with the TwoPhaseFlow app with the sphere in box case, which doesn't really use most of the h-adaptivity tools. Is there an existing TwoPhaseFlow version of the 2D/3D dambreak problem (or something similar) that I can test out to make sure there aren't any issues with h-adaptivity?

manuel-quezada:
@zhang-alvin you can find the 2D dam break problem with the setup by Colagrossi and the Marin problem here https://github.com/erdc/proteus/tree/master/proteus/tests/TwoPhaseFlow. Both tests should run with rans2p or rans3p and with either the clsvof model or the non-monolithic level set method (VOF, NCLS, etc). I would be interested to learn how you run those tests and the results you get.

tridelat:
@zhang-alvin are you planning to add things to TwoPhaseFlow for h-adaptivity before 1.6.0 is released?

zhang-alvin:
@tridelat @manuel-quezada Sorry for the delay. It looks like h-adapt works fine as is for now as long as the rans2p + vof+ls+rd+mcorr setup is used. I've pushed a case to the `scorec/meshAdaptAniso_examples` branch of `air-water-vv`: `https://github.com/erdc/air-water-vv/tree/scorec/meshAdaptAniso_examples/MeshAdapt_PredictiveControl/dambreak_Colagrossi_TwoPhaseFlow`

which shows how to use h-adapt with the Two-Phase-Flow app. I have below a sample initial mesh and final mesh with the VOF field.

initialMesh

finalMesh

manuel-quezada:
@zhang-alvin two comments: 
- I have made h-adaptivity work with rans2p+clsvof within an older version of TwoPhaseFlow. I am sure that either it works with your current setup or it should be trivial to make it work as well. Maybe it is worth testing that out for this release. 
- I suggest to have the example/test of h-adaptivity not only on air-water-vv but also in Proteus.

zhang-alvin:
@manuel-quezada 

1)I took a look at rans2p+clsvof and there were some changes that needed to be made, particularly in the post-adapt solution transfer/proteus-reinitialization step, but guaranteeing that the solution transfer is doing the right thing will take some time.

I get the following results:
Initial
initialMesh_clsvof
Final
finalMesh_clsvof

In the final step, the vof field towards the left shows some pronounced perturbations/waviness. Is this expected behavior from clsvof? 

If things look alright, I'll go ahead and push the changes into `release_1.6.0` branch. Or maybe a PR if that's cleaner (@tridelat ).

2) I agree that these examples should probably be in the proteus repo; ideally, I would put these examples as part of a notebook, but I think that would hold up this release for quite a while.

manuel-quezada:
@zhang-alvin 
1) No, the clsvof shouldn't behave like that and it doesn't for this problem with non adaptive meshes. Maybe it is better if we don't delay this release for this to be working. If you push these changes to a different branch, I can take a look at them whenever I get a chance. 
2) Understood.

tridelat:
@zhanga @cekees I am about to merge this (when the tests pass), but without squashing as there is #958 that was merged here instead of master, in order to keep proper history

PR #958, opened 2019-04-26
Small Fixes and Cleanup

zhang-alvin:
This small PR is to correct a merge error from a previous PR where a function declaration is omitted and some variables needed to be defined in the solution transfer process.. 

Also, there seems to be an unnecessary warning regarding chrono in RANS2P.py.


tridelat:
@cekees I don't know what this chrono_model stuff is for, looks like it was added from changes coming from an IBM related PR

PR #959, opened 2019-05-02
AM: fix for residual when sum is non-zero

tridelat:
Fix for consistency issue with the AddedMass model with certain meshes in parallel


tridelat:
@cekees , as this is done per processor, do you mean that the addition to the residual should be first gathered and then logged? For storing, it can be in a variable per processor.

cekees:
I think it actually needs to summed and added only to the zeroth node on the zeroth processor.

tridelat:
@cekees that's not what it is doing right now, but it can be modified to gather with MPI and then substract on master processor if that's preferable? (and then we can log and/or store)

PR #980, opened 2019-06-24
MBD: boundaryFlags set as empty list

tridelat:
Mini PR for setting boundaryFlags as empty list instead of `None` (for passing `for` loops without checking if it is `None`)


tridelat:
@cekees merged this as it is only changing one line and is needed before the 1.6.1 release

PR #988, opened 2019-07-02
AddedMass: matrix tansformation from global to local

tridelat:
Transformation of added-mass matrix from global to local frame

- AM global to local
- Sum of moments from Proteus fixed

### DIFF
--- a/proteus/NonlinearSolvers.py
+++ b/proteus/NonlinearSolvers.py
@@ -689,6 +689,7 @@ class AddedMassNewton(Newton):
     def solve(self,u,r=None,b=None,par_u=None,par_r=None):
         if self.F.timeIntegration.t >= self.F.coefficients.next_solve:
             self.F.coefficients.next_solve += self.F.coefficients.solve_rate
+            self.F.coefficients.updated_global = True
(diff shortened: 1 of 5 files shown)


tridelat:
@cekees it seems to be performing well, I think a test would be useful such as solving the added-mass problem several times for a rectangle that keeps the same dimensions but that is rotated in different directions in a single fluid phase (so that the should the local added-mass matrix should be the same if the domain walls are infinitely far).

adimako:
@tridelat leave the branch open to merge it to 1.6.X. Or can you confirm that this thing with M_t is the only bug addressed in this one?

tridelat:
@adimako this PR ended up addressing 2 things: the global to local added mass matrix, and the moments bug, both should go into 1.6.2. Let me know when/if you are planning to release

adimako:
@tridelat. If the global to local added mass matrix is not a bug. I'd rather merge only the bug. I believe that we should move to 2.0.0 because the switch to python 3 is a significant one.  We should not maintain 1.6.X properly, but only fix bugs to allow for people to use it

PR #995, opened 2019-07-31
1.6.x

adimako:


adimako:
New wave theory. @tridelat this is very similar to the focused wave one

PR #1000, opened 2019-08-23
AM: restrained DOF contributions removed after rotation

tridelat:
- remove restrained DOF contributions after rotation
- global 2 local switch (default is False)


tridelat:
@cekees @jhcollins There is also some work to do for the added-mass to apply it exactly when solved globally with the AddedMass model because: moments are applied locally, but forces are applied globally in chrono. (so we cannot just rotate the whole added-mass matrix to local coordinates..)

For now it is okay to just keep it in global form because the added-mass contributions are usually larger for translational components than rotational components in the simulations we run

cekees:
Need to make the 1.7.0 release this morning. Are you trying to get this in before the release?

PR #1005, opened 2019-09-10
Added capability for reading multi-block stls

adimako:


adimako:
@cekees @tridelat @BranocRichards @jhcollins you may find this interesting. I modified the STLShape code to read multi-block STLs. This means that you could either:
-Concatenate a series of stl files to a single stl file and read it to assign vertices, facets (and associated boundary flags) 
-Save a multiblock file in paraview (e.g. an OpenFOAM case with boundaries or a vtk file perhaps) as an stl, and load it in Proteus
Note that in an stl file, an stl block is defined from a line that starts with solid, to a line that starts with endsolid. The new reader assign boundary names based on the stl naming e.g. if the stl block starts with `solid Bed0`, then this boundary will be named `Bed0`. 
This development came after looking (with Antonino) how to load the Fargo Case from OpenFOAM to Proteus, and we kind of believe that this should be convenient. 
I 'd like some feedback especially for @tridelat who has written the original code

tridelat:
@adimako looks good, maybe a new test would be nice too. Does it concatenate well when vertices are shared between the blocks? (I think that was one issue you mentioned originally)

adimako:
It did create the mesh of an STL from OpenFOAM. Antonino will set up a simple test case with 6-7 stls and we will see if we can get the flow

cekees:
Sounds like a good plan. I'm glad this is working.

adimako:
It seems like it is working. This is a test case prepared by Antonino following the Ansys - OpenFOAM workflow. All boundaries are STLs, we did not any of the tank3D classes etc.

image

adimako:
We will put some flow inside and if it works, we will replace with the Fargo case

adimako:
... including some changes that are in 1.6 as well

proteus/SpatialTools.py
```diff
@@ -1071,15 +1071,19 @@ def __init__(self, domain, filename):
...
-        self.BC_list = [self.BC['stl']]
+        #        self.boundaryTags = {'stl': 1}
+        self.BC={}
+        self.BC_list=[]
+        for key,value in self.boundaryTags.items():
+            self.BC[key] = self.BC_class(shape=self, name=key)
+            self.BC_list += [self.BC[key]]
         # self.BC = BCContainer(self.BC_dict)
-
+
```
tridelat: trailing whitespace

tridelat:
It would be nice to have a small section in the documentation explaining how to use multi-block STLs as well, for example in https://erdc.github.io/proteus-doc/tools/spatial_tools.html (source: https://github.com/erdc/proteus-doc/blob/master/source/tools/spatial_tools.rst )

adimako:
I would not remove the boundary condition issues, as these were done in the course of the development work for the STLShape.

WaveTools changes are mostly cosmetic - I could revert the master files and do these separately.

adimako:
@tridelat see https://github.com/erdc/proteus/pull/1015. Once we merge this one, the waveTools changes will disappear

adimako:
@tridelat see changes I did (no WaveTools, remove white space) and merge. Changes in Boundary conditions had to be done to make the STLShape workflow work.

Note that I am happy the way it is working now for a simple inlet / outlet domain. Perhaps we should try to set-up a flume with porous regions and relaxation zones

tridelat:
@adimako @cekees I merged this, but the STL file for the test is very large (1.1M), so I think it would be better to use simpler files for the test (such as a simpler STL geometry) or to use git lfs if it is necessary to use such a large file. The proteus repository is getting quite heavy (385+MB)

PR #1011, opened 2019-09-17
Adimako/1.6.x fix stl

adimako:


adimako:
waiting for tests to run

PR #1012, opened 2019-09-27
Update centos.py

adimako:
working for hydra


tridelat:
Codecov does not pass because the edited lines are machine-specific (cannot really be tested with Travis), merging

PR #1017, opened 2019-10-14
Adimako/boundary normals

adimako:
- [x] Adding place holders for boundary function to include normal vectors at each face

- [x] Modifying wall boundary conditions to accommodate local unit normals

- [x] Retrieving unit normals from the k / dissipation model


adimako:
@tridelat and @cekees any ideas how to retrieve normals at boundaries? 
My understanding is that the boundary points are the ebqe['x'] structure. But I cannot find any normals. 

We could either read them or calculate them? We should also make it generic for mesh motion

proteus/NumericalFlux.py
```diff
@@ -1798,11 +1798,25 @@ def __init__(self,vt,getPointwiseBoundaryConditions,
         self.hasInterior=False
     def setDirichletValues(self,ebqe):
         for ci in range(self.nc):
-            self.ebqe[('u',ci)].flat[:] = ebqe[('u',ci)].flat[:]
-            for (ebNE,k),g,x in zip(list(self.DOFBoundaryConditionsDictList[ci].keys()),
-                                    list(self.DOFBoundaryConditionsDictList[ci].values()),
-                                    list(self.DOFBoundaryPointDictList[ci].values())):
-                self.ebqe[('u',ci)][ebNE,k]=g(x,self.vt.timeIntegration.t)
+            try:
```
cekees: @adimako here is an example of how to add a try/except statement to allow for the g(x,t,n) function interface for Dirichlet conditions.  I tested this on a simple duct problem. For Kappa, etc., you may need to make a change like this to the setDirichletValues member of some other NumericalFlux classes. https://github.com/erdc/proteus/blob/master/proteus/mprans/Kappa.py#L69

adimako: I only need the normals for the Dirichlet conditions, so I won't change advective / diffusive. Placeholders will be there in case someone wants to use it for other models

proteus/mprans/BoundaryConditions.py
```diff
@@ -1422,37 +1419,39 @@ def getFluidVelocityLocalCoords(self, xi, element, rank):
...
 
-    def setYplusNormalDirection(self, x, t, relax=1.0):
+    def setYplusNormalDirection(self, x, t,n, relax=1.0):
         """
         Return the point at y+ distance in normal
         direction from the boundary.
         """
         # near wall point
-        nP = (relax * self.Y * (self.nV)) + x
+        nP = (relax * self.Y * n) + x
```
cekees: Since n is the outward pointing normal on the fluid domain boundary, shouldn't  it be (-n)?

proteus/mprans/BoundaryConditions.py
```diff
@@ -1422,37 +1419,39 @@ def getFluidVelocityLocalCoords(self, xi, element, rank):
...
-                coords_relax = self.setYplusNormalDirection(x, t, relax)
+                coords_relax = self.setYplusNormalDirection(x, t, n ,relax)
                 xi, element, rank = self.findElementContainingCoords(coords_relax)
+                print (rank)
                 relax *= 0.5
             # just use the element containing the boundary quadrature point to interpolate to the y+ point
             u, v, w = self.getFluidVelocityLocalCoords(self.femSpace_velocity.elementMaps.getInverseValue(element, coords),
                                                        element,
                                                        rank)
+        print("END")
```
cekees: Should comment out or use log statements instead of print.

PR #1025, opened 2019-11-12
Eric/merge convex limiting fix

ejtovar:
PR for merging fixes for the modified Green-Naghdi equations (or dispersive shallow water). 

Things that were done:
- Fixed the convex limiting step for mGN equations 
- Turned on the entropy viscosity for mGN equations using the SW entropy/flux functions (this is also done in TAMU code) 
- Updated some of the readme file for the SWFlows app
- Change some little things on some of the tests 
- Added the analytical solution capability for tests 
- Create tests for pytest 
- Add relaxation of the local bounds for the convex limiting step

Things that need to be done:

- Do the refactoring of the SWFlows app to be like the TwoPhaseFlow stuff


cekees:
@tridelat and @zhang-alvin I think we should post-pone the refactorization of the SWFlows app until the new TwoPhaseFlow refactor has stabilized. What do you think? In that case @ejtovar can finish this up with some tests and the relaxation of local bounds in the convex limiting.

cekees:
Looks like some of the new reference solution data (h5) is not getting installed (see setup.py) or is not quite where the test thinks it is.

ejtovar:
Yes, I'm currently updating the comparison files.

cekees:
I'm guessing the problem is that the tests can't be run from an arbitrary location the way that the file paths are done in the tests. Are they assuming the test is being run from the same directory as the test file? If so you need to grab the path for the test file and then prefix that to the reference file. Take a look at some of the other tests.

ejtovar:
I used the TwoPhaseFlow tests file as a reference, but I will take a look at it later today and see what's going on.

cekees:
@ejtovar and @tridelat to debug this path error you can run on your local machine with
`PATH=./linux/bin:$PATH py.test -n 1 --dist=loadfile --forked -v proteus/tests/SWFlow`
which is just a short version of what .travis.yaml is doing. I don't think the tests that these were based off of were ever designed to actually run from travis. The basic issue is that you need to be able to run the test from anywhere, but in particularly the proteus root diretory, which where travis is going to run th test.

ejtovar:
@cekees It seems like the issue is related to the actual Green-Naghdi solver files. @zhang-alvin is currently helping me debug this. My working branch, where all the tests are currently passing, is still on version 1.6 with an older version of PETSc as well. Currently trying to hunt down the bug.

zhang-alvin:
@cekees, @ejtovar, @tridelat 

To provide a bit more detail, I can duplicate the errors seen on Travis-CI with the command:
`PATH=./linux/bin:$PATH py.test -n 1 --dist=loadfile --forked -v proteus/tests/SWFlow`

If I look at any of the log files produced, however, I see that there isn't much happening. For example, in the seawall case with log level 5, the entire log file is: 

```
[       0] Initializing Proteus
[       0] Initializing MPI
[       0] Adding . to path for loading modules
[       0] Loading problem-specific modules
[       0] Importing input modules
[       0] /home/zhanga/barn/proteus/proteus_conda/proteus/SWFlow/models/
[       0] /home/zhanga/barn/proteus/proteus_conda/proteus/SWFlow/utils/
[       0] Loading module = SWEs_so.py; with case file = seawall.py
[       0] SWEs_so
[       0] /home/zhanga/barn/proteus/proteus_conda/proteus/SWFlow/utils/
```

The log file is generated in the current directory (which in this case is the proteus home directory). The test only fails at the .h5 file comparison stage which isn't even being generated because the solve is failing. So there's a separate issue here where the tests think they are passing even when the solve terminates earlier. 

Running the case directly gives slightly more informative output: 

`parun --SWEs --path . -l11 -v seawall.py -C 'refinement=4 final_time=0.1 dt_output=0.1'`

```
Constructing GN_SW2DCV<CompKernelTemplate<2,4,3,3,3,3>());
[       1] Allocating residual and solution vectors
[       1] Allocating Jacobian
[       1] Building sparse matrix structure
[       1] Allocating parallel storage
[       1] Building Mesh Transfers
[       1] Setting GN_sw_p stepController to <class 'proteus.StepControl.Min_dt_controller'>
[       1] Setting up MultilevelLinearSolver forGN_sw_p
[       1] multilevelLinearSolverChooser type= <class 'proteus.LinearSolvers.LU'>
[       1] Setting up MultilevelNonLinearSolver for GN_sw_p
[       1] Finished setting up models and solvers
[       1] Setting up SimTools for GN_sw_p
[       1] Using tnList from so = seawall
[       1] Time sequence[0.0, 0.001, 0.1]
[       1] NAHeader Num Time Steps 2
[       1] Setting seawall systemStepController to object of type <class 'proteus.SplitOperator.Sequential_MinModelStep'>
[       1] Finished NumericalSolution initialization
[       1] Setting initial conditions
[       1] Setting initial conditions for GN_sw_p
[       1] Setting initial conditions on model GN_sw_p
[       1] Attaching models and running spin-up step if requested
[       1] Attaching models to model GN_sw_p
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see http://www.mcs.anl.gov/petsc/documentation/faq.html#valgrind
[0]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple Mac OS X to find memory corruption errors
[0]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run 
[0]PETSC ERROR: to get more information on the crash.
application called MPI_Abort(MPI_COMM_WORLD, 59) - process 0
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=59
:
system msg for write_line failure : Bad file descriptor
```
And the failure occurs at `https://github.com/erdc/proteus/blob/master/proteus/mprans/GN_SW2DCV.py#L1144`

cekees:
Another issue is that the `--path` option is not really implemented for SWFlows. I needs this bit of code. 

https://github.com/erdc/proteus/blob/master/scripts/parun#L372

cekees:
With the fix for the --path option, I can reproduce the errors you're getting @zhang-alvin. I didn't realize this branch wasn't being tested  on the 1.7 stack outside of travis. @ejtovar you should probably do that first.

tridelat:
@ejtovar has there been new development regarding this? The path issue should be easy to fix, but the failure that @zhang-alvin reported looks like the bigger thing to fix. Just wondering in order to merge this for the next release.

ejtovar:
@tridelat Sorry for the delay, I've been caught up with stuff at A&M, but I will be working on this and hopefully get back to you soon.

cekees:
Yes, thanks for jumping into this @zhang-alvin .

PR #1033, opened 2019-12-12
Removed warnings from mprans

JohanMabille:
The only remaining warning is about Numpy deprecated API.


zhang-alvin:
Were the source of these warnings from variables that were declared but not used?

JohanMabille:
Yes, some of them were even initialized after some computation.

PR #1034, opened 2019-12-12
Adimako/parameters1.6

adimako:


cekees:
Since the parameter changes already appear to be in master, I think we can just close this. The only changes here are to the reference output files, but I think they were generated by the 1.6.X release. \cc @tridelat

Issue #922, opened 2019-03-18
Refactor TwoPhaseFlow App


cekees:
- [ ] rewrite parun as pro using argparse
- [ ] replace app option as subcommand and move over app-specific options to subcommand
- [ ] split off some phases as subcommands (e.g. mesh)
- [ ] move _p,_n,_so modules into single file using p/n/so object-based approach
- [ ] move petsc options files into same single python file and set through petsc4py
- [ ] create subdirectory for apps
- [ ] refactor or remove fastarchive option if not needed
- [ ] move static docs into doc directory
- [ ] move examples into tests and enable travis on them--don't want any untested code in the repo

manuel-quezada:
Sorry for the stuff about fast archiving. I was having some troubles when outputting big files here at Kaust so I introduced that to speed up the process but I was not intending to get that into master. And, by the way, it didn't help much.

cekees:
No problem. Good to know the story behind it. The archiving needs a good cleaning soon, so I'd be interested it any performance information. We could probably add some options to selectively disable archiving for some fields.

cekees:
@tridelat and @manuel-quezada  here is a simple example of object-based physics and numerics in a single-file problem description: https://github.com/erdc/proteus/blob/master/proteus/tests/periodic/duct.py

The basic idea is to just walk through the needed models setting up a default object and then overriding the options, then set up the global system parameters. e.g. 
```
from proteus.defaults import (Physics_base, Numerics_base)
# Navier-Stokes physics
nse_p = Physics_base()
nse_p.LevelModel = RANS3PF.LevelModel.
...
# Navier-Stokes numerics
nse_n = Numerics_base()
nse_n =  LinearSolver.KSP_Petsc4py
...
# CLSVOF physics
clsvof_p = Physics_base()
...
# CLSVOF_numerics
clsvof_n = Numerics_base()
...
# global system
from proteus.default_so import *
pnList=[(nse_p, nse_n),
             (clsvof_p, clsvof_n)]
...
```
This avoids the necessity of doing `from proteus.default_p import *`, etc. and the dangers that involves due to the fact that Python modules are singleton. I would like to get rid of that last `from proteus.default_so import *`, but I have thought through the best way to do that yet. We could just have parun check for the existence of a single `so` object that could be initialized in the same manner.

zhang-alvin:
[...]

PR #1035, opened 2019-12-12
TPF: Simplifying TPF initialization 

zhang-alvin:
Starting this PR to get some feedback on proposed changes to `TwoPhaseFlow` app case setups (see #922).

The primary changes are:

1. `ParameterModelBase` objects can be declared without an association with a `TwoPhaseFlowProblem` object. 
2. Users need to declare a list of model objects derived from `ParameterModelBase` (e.g. [`ParameterModelBaseRANS2P()`, `ParameterModelBaseVOF()`]) and pass that to the instantiation of a `TwoPhaseFlowProblem` object.
3. Removal of indexing logic and instead derive indices from the supplied model list. This requires that the model object names be hard-coded and immutable, which is reasonable (i.e. the user should not be allowed to set the `ParameterModelRANS2P.name` attribute).

Remaining changes to be made: 

- [ ] Rework boundary conditions and initial conditions to be associated with the models directly
- [ ] Associate various attributes to either models, the domain, or another object to simplify the instantiation of `TwoPhaseFlowProblem`
- [ ] Rework existing test problems to function with new interface and compare results


zhang-alvin:
At the moment, tests are failing because of shift in notation. We can handle that after all of the refactoring is done.

PR #1038, opened 2019-12-13
TwoPhaseFlow Refactoring Development Branch 

zhang-alvin:
This will be the master PR for the TwoPhaseFlow refactoring (#922). Please use this PR to discuss possible directions so we can converge to a final interface.

### Development Guidelines:

- New developments should be branched from `develop/TPF` and PRs should be made against `develop/TPF`. 
- Completed PRs should be squash merged.

The final merge against master will be a normal merge to preserve commit authorship. 

### Desired Procedure:
(based on discussions with @tridelat)

User has some knowledge of engineering and programming, but not necessarily deep knowledge of fluids modelling or the underlying mathematics.

1. User has a problem of interest to solve and starts building a case-setup file
2. User defines a domain
3. User specifies mesh settings or supplies a mesh
4. User declares the `TwoPhaseFlowProblem` object
5. User adds necessary models for problem with `TwoPhaseFlowProblem.addModel()` 
6. User specifies initial conditions and boundary conditions for each model
7. User runs the case with a CLI

### Example case.py

```
import TwoPhaseFlow as TpFlow

#domain and mesh definition

myTpFlowProblem = TpFlow.TwoPhaseFlowProblem(domain=domain)

myTpFlowProblem.addModel(TpFlow.Parameters.ParametersModelRANS2P(),'flow')
myTpFlowProblem.addModel(TpFlow.Parameters.ParametersModelVOF(),'vof ')

myTpFlowProblem.modelDict['flow'].p.initialConditions.pressure.uOfXt = 
myTpFlowProblem.modelDict['flow'].p.initialConditions.u.uOfXt = 
myTpFlowProblem.modelDict['flow'].p.initialConditions.v.uOfXt = 

myTpFlowProblem.modelDict['vof'].p.initialConditions.vof.uOfXt=

# let 'wall_x+' be some boundary in the domain

myTpFlowProblem.modelDict['flow'].p.pressure.setBC('wall_x+', BC_function)  
myTpFlowProblem.modelDict['flow'].p.vel_u.setBC('wall_x+', BC_function) 

# or for more advanced BCs like wave generation
BC.setWaveGen(myTpFlowProblem,'wall_x+', waveParameters) 
```


zhang-alvin:
@cekees , @ejtovar , @tridelat , @adimako

Have a look. Feel free to add comments on the desired interface and I will update the procedure and example as necessary.

tridelat:
@zhang-alvin Thanks for the PR!

I think that for BC, we want this kind of syntax:
```
myproblem.models['rans2p'].p.BC['x-'].p.uOfXT = ...
myproblem.models['rans2p'].p.BC['x-'].u.uOfXT = ...
myproblem.models['rans2p'].p.BC['x+'].v.uOfXT = ...
...
```
So: problem -> model -> BC -> variable
Similar for initial conditions:
```
myproblem.models['rans2p'].p.IC.p.uOfXT = ...
myproblem.models['rans2p'].p.IC.u.uOfXT = ...
myproblem.models['rans2p'].p.IC.v.uOfXT = ...
...
```
And then for bigger boundary conditions (e.g. wave inlet), it could be something like this:
```
myproblem.BC['x-'].setUnsteadyVelocityInlet(*args, **kwargs)
```
And this would set optimized BCs for several models at once (with assertions making sure the models exist). The `BC` object/class placed there probably just needs to be initialized with the problem class instance as an attribute. The same would apply to initial conditions that are sometimes set for several models at one.
Anyway, I started modifying the BoundaryConditions and writing InitialConditions files, and will have a think on how to make this work nicely. Let me know what you think.

tridelat:
Btw, we need to find a better way than to have a context variable that must be absolutely named `myTpFlowProblem` for `parun` to access it

zhang-alvin:
> Btw, we need to find a better way than to have a context variable that must be absolutely named `myTpFlowProblem` for `parun` to access it

I think that this will be an easy fix. We can loop over the context objects and look for an object that is an instance of the `TwoPhaseProblem` class:

```
    prob = None
    for contextObject in ct:
        if isinstance(contextObject,proteus.TwoPhaseFlow.TwoPhaseFlowProblem.TwoPhaseFlowProblem):
            prob = contextObject
    assert prob is not None, "No TwoPhaseFlowProblem object found"
```

tridelat:
@zhang-alvin this is nicer, but can be problematic if more than one `TwoPhaseFlowProblem` instances are created in the context. Unlikely, but if this happens, it can become a confusing, kind of "hidden" bug. We can catch that though and throw an error if more than one `TwoPhaseFlow` instances are present in the context. 
On a side note, wasn't this PR designed only to merge other PRs? Just so we can edit this kind of things and finish them off before merging.

zhang-alvin:
> @zhang-alvin this is nicer, but can be problematic if more than one `TwoPhaseFlowProblem` instances are created in the co@zhang-alvin this is nicer, but can be problematic if more than one `TwoPhaseFlowProblem` instances are created in the context. Unlikely, but if this happens, it can become a confusing, kind of "hidden" bug. We can catch that though and throw an error if more than one `TwoPhaseFlow` instances are present in the context.
On a side note, wasn't this PR designed only to merge other PRs? Just so we can edit this kind of things and finish them off before merging

@tridelat Good call. Regarding the non-PR commit ... I didn't want to be overzealous with creating PRs for minor edits to the code, but you're right I should stick with the design of this PR.

tridelat:
For the record: forced pushed to fix typo in last commit

tridelat:
Just giving an update for the record on the PR: it was decided with @zhang-alvin that BoundaryConditions do not need to be changed for now (not a priority), and InitialConditions have been drafted there: https://github.com/erdc/proteus/pull/1071

Additionally, changes already on master allow Coefficients classes to be directly accessible (instead of using a "redundant" CoefficientsOptions class). We should probably merge this change here.

@zhang-alvin 

On extra things to do, we should rename class attributes that start with capital letters and that are not a class object (but are a class object instance), as this is good practice in Python

cekees:
@ejtovar it looks like there are indeed some changes to SW algorithms and data files here. Not sure that happened, but I may have done it when I tried to merge master into this branch. We should just be able to checkout out the versions of these 5-6 files that are SW-related and disjoint from TwoPhaseFlow, right?

zhang-alvin:
> @ejtovar it looks like there are indeed some changes to SW algorithms and data files here. Not sure that happened, but I may have done it when I tried to merge master into this branch. We should just be able to checkout out the versions of these 5-6 files that are SW-related and disjoint from TwoPhaseFlow, right?

@cekees I ran into this issue with my previous PR. You can always do a `git diff master -- filename` to see if there are any meaningful differences. My guess is that the master version should be the correct one.

proteus/NonlinearSolvers.py
```diff
@@ -755,6 +755,8 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
         self.F.secondCallCalculateResidual = 0
         self.computeResidual(u,r,b)
         u[:] = r
+        if par_u is not None:
```
cekees, line 758: @ejtovar this is really the only line I have a question about. This is the special SW nonlinear solver, but I don't know why these lines would be in there if you didn't put them in a commit at some point.

zhang-alvin:
@cekees I think this PR is ready to go. I'm going to open a separate PR for migrating the remainder of the meshing functions to MeshTools. After this goes in, I'll also add PRs for the proteus-tutorial and air-water-vv to reflect the updated tests for TPF.

cekees:
> @cekees I think this PR is ready to go. I'm going to open a separate PR for migrating the remainder of the meshing functions to MeshTools. After this goes in, I'll also add PRs for the proteus-tutorial and air-water-vv to reflect the updated tests for TPF.

Great. I'll review it later today.

cekees (approved):
Thanks, @zhang-alvin  Glad to so this finished!

PR #1039, opened 2019-12-14
Conda

cekees:


cekees:
@davidbrochart it looks like the conda environment still needs some tweaking. The mpi option I added to h5py fixed some tests, but next it looks like superlu and superlu_dist aren't getting configured for petsc and that there may also be some issues with gmsh and tetgen versions. You ought to be able to run these tests locally, either with `make test` or the relevant py.test commands in the travis yaml file.

zhang-alvin:
I'd like to help with getting some of these tests passing, but I'm unsure of how to begin with the conda building and testing workflow. My guess is to checkout the `conda` branch and use `conda-build` on some recipe, but I'm not sure which recipe file I should be supplying.

zhang-alvin:
Completely missed the README file contents. I think I got the workflow going now.

zhang-alvin:
My observations regarding some of the error messages: 

1) gmsh
```
        with open('gmsh_mesh_test_2D.node', 'r') as nodefile:
>           npt.assert_equal(nodefile.readline(), '3433 2 0 1\n')
E           AssertionError: 
E           Items are not equal:
E            ACTUAL: '3432 2 0 1\n'
E            DESIRED: '3433 2 0 1\n'
```
This comes from a very strict comparison between the number of entities in a saved gmsh-generated mesh and a new mesh generated from the conda-installed gmsh. The conda-forge gmsh is version 4.4.1. The hashdist version is 4.0. The proper fix is probably to allow some leeway in the test. 

2) tetgen
```
[       0] Calling tetgen on rank 0 with command tetgen -VApq1.25q12feena2.083333e-02 mesh.poly
Opening mesh.poly.
Opening mesh.edge.
Error:  Edge 1 has an invalid vertex index.
Error:  Out of memory.
```
I am almost sure this comes from a bug in tetgen 1.5 where if an existing mesh.edge file exists in the test directory, tetgen loads the mesh.edge file and fails. The command used to generate a mesh from `mesh.poly` in these tests are: `tetgen -VApq1.25q12feena2.083333e-02 mesh.poly`

Removing the `-e` option should avoid the production of the mesh.edge file: 
`tetgen -VApq1.25q12fna2.083333e-02 mesh.poly`

I haven't fully tested this idea yet. 

3) petsc/superlu
```
E   petsc4py.PETSc.Error: error code 92
E   [0] KSPSetUp() line 391 in /home/conda/feedstock_root/build_artifacts/petsc_1571651265346/work/src/ksp/ksp/interface/itfunc.c
E   [0] PCSetUp() line 932 in /home/conda/feedstock_root/build_artifacts/petsc_1571651265346/work/src/ksp/pc/interface/precon.c
E   [0] PCSetUp_LU() line 97 in /home/conda/feedstock_root/build_artifacts/petsc_1571651265346/work/src/ksp/pc/impls/factor/lu/lu.c
E   [0] MatGetFactor() line 4526 in /home/conda/feedstock_root/build_artifacts/petsc_1571651265346/work/src/mat/interface/matrix.c
E   [0] See http://www.mcs.anl.gov/petsc/documentation/linearsolvertable.html for possible LU and Cholesky solvers
E   [0] Could not locate solver package superlu. Perhaps you must ./configure with --download-superlu
```
No comment. But putting here for reference.

4) numpy
```
    def within_tol(x, y, atol, rtol):
        with errstate(invalid='ignore'):
>           return less_equal(abs(x-y), atol + rtol * abs(y))
E           ValueError: operands could not be broadcast together with shapes (599,) (711,)
```
No comment. But putting here for reference.

cekees:
On 3). I'm guessing that petsc is not actually being linked to superlu and superlu_dist. That would have to be done in the petsc conda package. We may be able to provide a variant of the conda recipe that contains a collection of external solver packages that we use through petsc.

cekees:
Actually, @martinRenou is already working on the petsc varianet: https://github.com/conda-forge/petsc-feedstock/pull/76

martinRenou:
Yes, there seems to be an issue on OSX... I am not really sure what it is, unfortunately. I will look again later.

cekees:
Agreed on 1. about gmsh. I would just relax the test.  On the tetgen issue, should the edge file be available at the time that tetgen runs and triggers the bug? Just wondering if we can clean up before the call to get around this or if the edge file causing the issue is possibly leftover from a prior test.

zhang-alvin:
> Just wondering if we can clean up before the call to get around this or if the edge file causing the issue is possibly leftover from a prior test.

The teardown method in some of the tests might be sufficient. I'll look into it.

zhang-alvin:
So the teardown method seems to work, although I haven't applied it to all the right tests yet. Looking specifically the `TwoPhaseFlow` `marin` test, there's a new error that pops up:

```
[31m[1m_________________________ TestTwoPhaseFlow.test_marin __________________________[0m
[gw0] linux -- Python 3.7.3 /home/travis/miniconda/envs/proteus-dev/bin/python
self = <proteus.tests.TwoPhaseFlow.test_TwoPhaseFlow.TestTwoPhaseFlow object at 0x7f1055785278>

    def test_marin(self):
        os.system("parun --TwoPhaseFlow --path " + self.path + " "
                  "marin.py -l5 -v -C 'final_time=0.1 dt_output=0.1 he=0.5'")
>       self.compare_vs_saved_files("marin")

proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py:80: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py:43: in compare_vs_saved_files
    assert np.allclose(expected.root.phi_t2,actual.root.phi_t2,atol=1e-8)
../../../miniconda/envs/proteus-dev/lib/python3.7/site-packages/numpy/core/numeric.py:2256: in allclose
    res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
../../../miniconda/envs/proteus-dev/lib/python3.7/site-packages/numpy/core/numeric.py:2335: in isclose
    return within_tol(x, y, atol, rtol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = array([-0.69489879,  1.89073621,  1.91054656, -0.8606013 ,  0.28859435,
        2.18782269,  2.2032087 ,  0.09772154, ... -0.44181861, -0.67134604,
        1.0349554 , -0.25031443,  0.34459293,  0.09810234,  1.19789997,
        1.20209344])
y = array([-4.91459012e-01,  1.98094816e+00,  1.95884112e+00, -5.20500891e-01,
        4.89034645e-01,  2.26546290e+00,  2...1,  1.68492685e+00,
        7.56677176e-02,  1.03563264e+00,  1.33969515e+00,  8.40721412e-01,
        6.74103474e-01])
atol = 1e-08, rtol = 1e-05

    def within_tol(x, y, atol, rtol):
        with errstate(invalid='ignore'):
>           return less_equal(abs(x-y), atol + rtol * abs(y))
E           ValueError: operands could not be broadcast together with shapes (171,) (193,)
```

This is happening because again, the test is too strict. The original comparison file had a mesh with 171 vertices whereas the test with the new tetgen/gmsh versions has a mesh with 193 vertices. This actually ties in with the repo-shrinking effort in which we probably convert these h5 file comparison tests to something a lot smaller, such as comparing the L2-norm of some relevant fields in the domain or setting up gauges and comparing those. 

Alternatively, we can save the mesh files which are much smaller than the h5 files (4 KB vs 428 KB) and have the numerics run through the same exact mesh. This is a more precise test as it removes mesh generation from the test.

@tridelat, @cekees I'd like your thoughts on this.

manuel-quezada:
@zhang-alvin I wrote this test. In my opinion computing some metric of the solution is reasonable for a test. I think some of the tests I wrote are doing that indeed. You can even choose different metrics for each test and they can even change from test to test depending on what piece of code you are considering. 

BTW, I saw you guys are trying to clean up the repository. Do you want me to check all my branches and delete them? I might want to keep one or two open, if that is OK.

martinRenou:
I started working on building `petsc` with `superlu_dist` and `superlu` during the workshop. `petsc` now builds correctly with `superlu` in my PR. I now need to fix the `superlu_dist` feedstock, I'm on it.

cekees:
> @zhang-alvin I wrote this test. In my opinion computing some metric of the solution is reasonable for a test. I think some of the tests I wrote are doing that indeed. You can even choose different metrics for each test and they can even change from test to test depending on what piece of code you are considering.
> 
> BTW, I saw you guys are trying to clean up the repository. Do you want me to check all my branches and delete them? I might want to keep one or two open, if that is OK.

@manuel-quezada I'll let @zhang-alvin give you the full story on what to do with your branches and things, but, yes, it would be good to go ahead and delete any that you are reasonably certain don't contain important code.

zhang-alvin:
> > @zhang-alvin I wrote this test. In my opinion computing some metric of the solution is reasonable for a test. I think some of the tests I wrote are doing that indeed. You can even choose different metrics for each test and they can even change from test to test depending on what piece of code you are considering.
> > BTW, I saw you guys are trying to clean up the repository. Do you want me to check all my branches and delete them? I might want to keep one or two open, if that is OK.
> 
> @manuel-quezada I'll let @zhang-alvin give you the full story on what to do with your branches and things, but, yes, it would be good to go ahead and delete any that you are reasonably certain don't contain important code.

@manuel-quezada Yes remove any branches that you can for now. The goal is to have each developer make a fork of proteus after the cleanup and the only branches that should exist on the main repo are the master branch, releases, and large development branches requiring coordination from multiple developers.

cekees:
@zhang-alvin and @davidbrochart this looks like goo progress with only 9 test failures. Is there anything I can help with?

davidbrochart:
All the [failing tests](https://github.com/erdc/proteus/pull/1045#issuecomment-569904264) (except the first one) can be fixed by reworking the checks that are either too strict or out of date (some files have to be re-created...).

zhang-alvin:
> @zhang-alvin and @davidbrochart this looks like goo progress with only 9 test failures. Is there anything I can help with?

@cekees I have tried out a way of resolving the remaining 9 tests in #1046 and discussed in #1044. The tests are failing because they're making a nodal field value comparison against .h5 files with different mesh topology. The approach to testing I tried was: 

1. First generate a coarse mesh and save it.
2. The test then loads the coarse mesh and runs through the numerics.
3. The generated (but not stored) .h5 file is read and an L2-norm of a scalar field is computed and compared against a reference value.

Advantages:
1. Tests are more well defined in that the mesh never changes, so just the numerics/methods are tested. 
2. Size of repo will be reduced during cleanup as .h5 files tend to be larger than the mesh files.
3. Tests might execute faster given the reduced size of the mesh.

Disadvantages:
1. L2-norms are not unique, so two fields on a mesh with different values on can yield the same L2-norm. It seems very unlikely for such a "collision" to occur though given a fixed problem definition.
2. A lot of tests need to be modified.

tridelat:
@zhang-alvin the L2 norm feels a little bit hacky, because a test that should have failed can have the same L2-norm as the stored L2-norm, although it is unlikely to happen.
Something else that could be done is using Gauges to compare specific points within the domain, and give a small tolerance as those would not change massively with a different mesh

davidbrochart:
@zhang-alvin you managed to make all the tests pass? I'm surprised even `test_AddedMass_3D` passed, I thought it was due to a conda-forge issue with `openblas` and `openmp`.

zhang-alvin:
> @zhang-alvin you managed to make all the tests pass? I'm surprised even `test_AddedMass_3D` passed, I thought it was due to a conda-forge issue with `openblas` and `openmp`.

I remember seeing that warning initially. Once I migrated the `AddedMass_3D` test to a new procedure, that warning disappeared. I'll double check what I did just in case.

zhang-alvin:
@davidbrochart I double-checked the setup of the test and all that I changed was the mesh file being used. If I used the previous mesh setup, then I get the `OpenBLAS` warning:

```
[      88]   NumericalAnalytics NewtonIteration: 0, NewtonNorm:  1.31037e+01
[      88]    Newton it 0 norm(r) =  1.31037e+01  		 norm(r)/(rtol*norm(r0)+atol) = 52414.7 test=r
OpenBLAS Warning : Detect OpenMP Loop and this application may hang. Please rebuild the library with USE_OPENMP=1 option.
```
If you know how to tackle this issue, that would be great.

davidbrochart:
@zhang-alvin Unfortunately this is not an easy issue, it involves either patching `gcc` or swapping the implementation of GNU's `openmp` with LLVM's one. It did try the former, but the folks at conda-forge don't want to include  a patch that could potentially break libraries that are used by a lot of people. Fot the latter, there are two PRs that will hopefully make it possible soon:
https://github.com/conda-forge/openmp-feedstock/pull/29
https://github.com/conda-forge/ctng-compilers-feedstock/pull/7

cekees:
Just to make sure I'm following: The reason we can't just compile openblas with openmp to make the warning go away is that gnu's openmp is not fork-safe? We're not using openmp or threads anywhere in proteus and generally haven't enabled it in any of the linear algebra kernels we use either, so it may be fine just to move on and be aware of the warning. We typically use all the cores on a processor via MPI-based parallelism anyway.  I think the real performance payoff will be on optimizing our FEM assembly using CUDA.

cekees:
The basic tests now run in exactly the same amount of time for the hashdist-based stack and the conda-based stack, so that's good. The conda environment install is way faster because it's completely binary, whereas there are few hashdist packages that have to be built from source right now.  So that's all good news. (Test time is 430s)

cekees:
@davidbrochart and @zhang-alvin the actual proteus build time could probably be halved by calling calling the build_ext stage explicitly, which can take advantage of the two threads available on travis.  Not a big deal.  Just making sure you know that the proteus build itself can be run in parallel, which helps a lot since our build is dominated by C++ compiles of template classes.

cekees:
I don't think it's worth trying to enable the air-water-vv tests right now. It's not hard but will take some time--I started on it here: https://github.com/erdc/air-water-vv/pull/232.

davidbrochart:
> Just to make sure I'm following: The reason we can't just compile openblas with openmp to make the warning go away is that gnu's openmp is not fork-safe? We're not using openmp or threads anywhere in proteus and generally haven't enabled it in any of the linear algebra kernels we use either, so it may be fine just to move on and be aware of the warning. We typically use all the cores on a processor via MPI-based parallelism anyway. I think the real performance payoff will be on optimizing our FEM assembly using CUDA.

Yes, that is the reason: GNU's `openmp` is  not fork-safe, so we can't build `openblas` with `USE_OPENMP=1`. The problem is that the warning is real, it *may* hang, and indeed it did in my tests. Now it seems to have disappeared, but it can reappear.

> @davidbrochart and @zhang-alvin the actual proteus build time could probably be halved by calling calling the build_ext stage explicitly, which can take advantage of the two threads available on travis. Not a big deal. Just making sure you know that the proteus build itself can be run in parallel, which helps a lot since our build is dominated by C++ compiles of template classes.

I tried to call `build_ext` but all the workers died in the CI when I did.

zhang-alvin:
@cekees , @davidbrochart I gave the `python setup.py build_ext -i` command a shot in #1049 , but I don't see any speedup in the build: https://travis-ci.com/erdc/proteus/builds/143464373. 

The workers seem to be executing just fine.

cekees:
@zhang-alvin could you put in a print statement around here to see if we're actually getting access to the 2 threads that are supposed to be available? https://github.com/erdc/proteus/blob/master/setup.py#L944

zhang-alvin:
> @zhang-alvin could you put in a print statement around here to see if we're actually getting access to the 2 threads that are supposed to be available? https://github.com/erdc/proteus/blob/master/setup.py#L944

I put several statements in to see if two workers are being initialized and distributed work appropriately. From the looks of it, they are:

https://travis-ci.com/erdc/proteus/builds/143486045 

For example, you can see towards the beginning of the log where two workers are initialized where they print out a unique integer. The process ids for each worker are then printed followed by the extension they are to build.
```
Hello Worker 0
[INFO/ForkPoolWorker-2] child process calling self.run()
[INFO/ForkPoolWorker-2] child process calling self.run()
Hello Worker 1
worker id 9012  [<Cython.Distutils.extension.Extension('mprans.cPres') at 0x7f108d4e3dd0>]
worker id 9013  [<Cython.Distutils.extension.Extension('mprans.cPresInit') at 0x7f108d4e3f10>]
```

cekees:
@zhang-alvin it looks like it is correctly farming out the build tasks to two threads. Not sure why there isn't any speed up. It's possible that `pip install...` is triggering the parallel build, but I don't see how. Maybe there really aren't two physical cores on the travis instance. Anyway, the builds are significantly faster already, so maybe we don't worry about it.  We could also turn off optimizations in the C++ compiler if speed were a huge issue. 10 min for the proteus build is really not too bad.

zhang-alvin:
> @zhang-alvin it looks like it is correctly farming out the build tasks to two threads. Not sure why there isn't any speed up. It's possible that `pip install...` is triggering the parallel build, but I don't see how. Maybe there really aren't two physical cores on the travis instance. Anyway, the builds are significantly faster already, so maybe we don't worry about it. We could also turn off optimizations in the C++ compiler if speed were a huge issue. 10 min for the proteus build is really not too bad.

Just for completeness, I took a look a local call to `pip install -v -e .` with the same instrumentation in #1049 doesn't yield multiple workers/threads:

```
...
Running command /home/zhanga/miniconda3/envs/proteus-dev/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/zhanga/barn/proteus/proteus_alvin/setup.py'"'"'; __file__='"'"'/home/zhanga/barn/proteus/proteus_alvin/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
    worker id 16528  [<Cython.Distutils.extension.Extension('mprans.cPres') at 0x7f0773302e90>, <Cython.Distutils.extension.Extension('mprans.cPresInit') at 0x7f0773375bd0>, <Cython.Distutils.extension.Extension('mprans.cPresInc') at 0x7f07711180d0>, <Cython.Distutils.extension.Extension('mprans.cAddedMass') at 0x7f075d51a690>, <Cython.Distutils.extension.Extension('mprans.SedClosure') at 0x7f0773302e50>, <Cython.Distutils.extension.Extension('mprans.cVOF3P') at 0x7f075d2d5f10>, <Cython.Distutils.extension.Extension('mprans.cVOS3P') at 0x7f075d2d5f90>, <Cython.Distutils.extension.Extension('mprans.cNCLS3P') at 0x7f075d2d5fd0>, <Cython.Distutils.extension.Extension('mprans.cMCorr3P') at 0x7f075d2d3550>, <Cython.Distutils.extension.Extension('richards.cRichards') at 0x7f075d2d3750>, <Cython.Distutils.extension.Extension('elastoplastic.cElastoPlastic') at 0x7f075d2d3590>, <Cython.Distutils.extension.Extension('mprans.cRANS3PF') at 0x7f075d2d37d0>, <Cython.Distutils.extension.Extension('mprans.cRANS3PSed') at 0x7f077137f910>, <Cython.Distutils.extension.Extension('Isosurface') at 0x7f0771bda050>, <Cython.Distutils.extension.Extension('BoundaryConditions') at 0x7f075d2d3650>, <Cython.Distutils.extension.Extension('mprans.BoundaryConditions') at 0x7f075d2d7090>, <Cython.Distutils.extension.Extension('mprans.MeshSmoothing') at 0x7f075d2d70d0>, <Cython.Distutils.extension.Extension('mprans.cMoveMeshMonitor') at 0x7f075d2d7110>, <Cython.Distutils.extension.Extension('mbd.CouplingFSI') at 0x7f075d2d7150>, <Cython.Distutils.extension.Extension('WaveTools') at 0x7f075d2d7190>, <Cython.Distutils.extension.Extension('fenton.Fenton') at 0x7f075d2d7210>, <Cython.Distutils.extension.Extension('ADR') at 0x7f075d2d7790>, <Cython.Distutils.extension.Extension('subsurfaceTransportFunctions') at 0x7f075d2d7390>, <Cython.Distutils.extension.Extension('equivalent_polynomials') at 0x7f075d2d72d0>, <Cython.Distutils.extension.Extension('cfemIntegrals') at 0x7f075d2d71d0>, <Cython.Distutils.extension.Extension('csparsity') at 0x7f075d2d7290>, <Cython.Distutils.extension.Extension('cmeshTools') at 0x7f075d2d7b50>, <Cython.Distutils.extension.Extension('ctransportCoefficients') at 0x7f075d2d7c90>, <Cython.Distutils.extension.Extension('csubgridError') at 0x7f075d2d7bd0>, <Cython.Distutils.extension.Extension('cshockCapturing') at 0x7f075d2d8110>, <Cython.Distutils.extension.Extension('superluWrappers') at 0x7f075d2d8190>, <Cython.Distutils.extension.Extension('csmoothers') at 0x7f075d2d8210>, <Cython.Distutils.extension.Extension('canalyticalSolutions') at 0x7f075d2d82d0>, <Cython.Distutils.extension.Extension('clapack') at 0x7f075d2d8890>, <Cython.Distutils.extension.Extension('cpostprocessing') at 0x7f075d2d8950>, <Cython.Distutils.extension.Extension('cnumericalFlux') at 0x7f075d2d8990>, <Cython.Distutils.extension.Extension('ctimeIntegration') at 0x7f075d2d89d0>, <Cython.Distutils.extension.Extension('cTwophaseDarcyCoefficients') at 0x7f075d2d8ad0>, <Cython.Distutils.extension.Extension('cSubsurfaceTransportCoefficients') at 0x7f075d2d8c10>, <Cython.Distutils.extension.Extension('cpskRelations') at 0x7f075d2d8d50>, <Cython.Distutils.extension.Extension('cpartitioning') at 0x7f075d2d8ed0>, <Cython.Distutils.extension.Extension('flcbdfWrappers') at 0x7f075d2d8fd0>, <Cython.Distutils.extension.Extension('mprans.cCLSVOF') at 0x7f075d2ee0d0>, <Cython.Distutils.extension.Extension('mprans.cNCLS') at 0x7f075d2ee110>, <Cython.Distutils.extension.Extension('mprans.cMCorr') at 0x7f075d2ee150>, <Cython.Distutils.extension.Extension('mprans.cRANS2P') at 0x7f075d2ee1d0>, <Cython.Distutils.extension.Extension('mprans.cRANS2P_IB') at 0x7f075d2ee210>, <Cython.Distutils.extension.Extension('mprans.cRANS2P2D') at 0x7f075d2ee250>, <Cython.Distutils.extension.Extension('mprans.cRDLS') at 0x7f075d2ee290>, <Cython.Distutils.extension.Extension('mprans.cVOF') at 0x7f075d2ee2d0>, <Cython.Distutils.extension.Extension('mprans.cMoveMesh') at 0x7f075d2ee310>, <Cython.Distutils.extension.Extension('mprans.cMoveMesh2D') at 0x7f075d2ee350>, <Cython.Distutils.extension.Extension('mprans.cSW2D') at 0x7f075d2ee390>, <Cython.Distutils.extension.Extension('mprans.cSW2DCV') at 0x7f075d2ee410>, <Cython.Distutils.extension.Extension('mprans.cGN_SW2DCV') at 0x7f075d2ee490>, <Cython.Distutils.extension.Extension('mprans.cKappa') at 0x7f075d2ee510>, <Cython.Distutils.extension.Extension('mprans.cKappa2D') at 0x7f075d2ee550>, <Cython.Distutils.extension.Extension('mprans.cDissipation') at 0x7f075d2ee590>, <Cython.Distutils.extension.Extension('mprans.cDissipation2D') at 0x7f075d2ee5d0>]
    running develop
    running egg_info
    writing proteus.egg-info/PKG-INFO
    writing dependency_links to proteus.egg-info/dependency_links.txt
    writing top-level names to proteus.egg-info/top_level.txt
    reading manifest file 'proteus.egg-info/SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    warning: no previously-included files matching '.gitmodules' found anywhere in distribution
    warning: no previously-included files matching '.travis.yml' found anywhere in distribution
    warning: no previously-included files matching 'Dockerfile' found anywhere in distribution
    writing manifest file 'proteus.egg-info/SOURCES.txt'
    running build_ext
    skipping 'proteus/mprans/cPres.cpp' Cython extension (up-to-date)
...
```
I'm going to go ahead and close #1049 since it didn't lead to any speedup.

JohanMabille:
@cekees @zhang-alvin @tridelat is there anything (except the conflict in Makefile that should be solved with #1053) preventing this PR from being merged?

zhang-alvin:
@JohanMabille The main thing that we were waiting for was for the OpenBLAS/OpenMP issue to be resolved:

>@zhang-alvin Unfortunately this is not an easy issue, it involves either patching gcc or swapping the implementation of GNU's openmp with LLVM's one. It did try the former, but the folks at conda-forge don't want to include a patch that could potentially break libraries that are used by a lot of people. Fot the latter, there are two PRs that will hopefully make it possible soon:
conda-forge/openmp-feedstock#29
conda-forge/ctng-compilers-feedstock#7

The PRs that @davidbrochart listed have merged in. I don't know if we need to modify the conda environment in any way to accomodate those changes.

davidbrochart:
@zhang-alvin I think I need to change the recipe of openblas, I will ping you when it's ready.

cekees:
I'm fine with this going in even before the openblas/openmp issue (warning) is resolved on the conda side.

proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py
```diff
@@ -20,17 +20,38 @@ def setup_method(self,method):
...
-        expected_path = 'comparison_files/' + name + '.h5'
-        expected = tables.open_file(os.path.join(self._scriptdir,expected_path))
         actual = tables.open_file(name+'.h5','r')
-        assert np.allclose(expected.root.phi_t2,actual.root.phi_t2,atol=1e-8)
-        assert np.allclose(expected.root.velocity_t2,actual.root.velocity_t2,atol=1e-8)
-        expected.close()
+
+        expected_path = 'comparison_files/' + 'comparison_' + name + '_phi_t2.csv'
+        #write comparison file
+        np.array(actual.root.phi_t2).tofile(os.path.join(self._scriptdir, expected_path),sep=",")
```
zhang-alvin: @zhang-alvin needs to comment this out

proteus/tests/TwoPhaseFlow/test_TwoPhaseFlow.py
```diff
@@ -20,17 +20,38 @@ def setup_method(self,method):
...
+        np.testing.assert_almost_equal(np.fromfile(os.path.join(self._scriptdir, expected_path),sep=","),np.array(actual.root.phi_t2).flatten(),decimal=10)
+
+        expected_path = 'comparison_files/' + 'comparison_' + name + '_velocity_t2.csv'
+        #write comparison file
+        np.array(actual.root.velocity_t2).tofile(os.path.join(self._scriptdir, expected_path),sep=",")
```
zhang-alvin: @zhang-alvin needs to comment this out

proteus/tests/cylinder2D/ibm_rans2p_3D/test_cylinder3D_ibm_rans2p.py
```diff
@@ -38,6 +45,7 @@ def teardown_method(self, method):
     def test_ex2(self):
         self.compare_name = "T1_ibm_3D_rans2p"
         self.example_setting("T=0.01 onlySaveFinalSolution=True")
+        self.teardown_method(self)
```
zhang-alvin: @zhang-alvin this call can be removed without losing functionality.

proteus/tests/cylinder2D/sbm_3Dmesh/test_cylinder2D_on_3D_mesh_sbm.py
```diff
@@ -29,12 +28,21 @@ def setup_method(self, method):
...
+                os.remove(file)
+            else:
+                pass
+
 
 
     def test_ex1(self):
         self.compare_name = "T001_P1_sbm_3Dmesh"
         self.example_setting("T=0.01 spaceOrder=1 onlySaveFinalSolution=True")
+        self.teardown_method(self)
```
zhang-alvin: @zhang-alvin this can be removed

JohanMabille:
In this case, I think it could be worth providing a second build in travis.yaml tha builds with hashdist so that both build approaches are tested.

This also means that we need to figure out how to integrate `xtensor-python` and `pybind11` in the hashdist build if we want to migrate the code of mprans to `xtensor`.

cekees:
We have the buildbots here for longer and more fine-grained testing: http://levant.hrwallingford.com:8080
I'm not opposed to adding a second build on travis, but I think we may want to use additional builds to increase the level of test coverage and let the buildbots handle the testing of source-based builds. The travis hashdist build has always been a hack because the hashdist binary capabilities are so basic.
@zhang-alvin and @tridelat  and @davidbrochart should we go ahead and merge this?

zhang-alvin:
> We have the buildbots here for longer and more fine-grained testing: http://levant.hrwallingford.com:8080
> I'm not opposed to adding a second build on travis, but I think we may want to use additional builds to increase the level of test coverage and let the buildbots handle the testing of source-based builds. The travis hashdist build has always been a hack because the hashdist binary capabilities are so basic.
> @zhang-alvin and @tridelat and @davidbrochart should we go ahead and merge this?

I'm okay with using the buildbots to cover the hashdist build since the majority of the developers/users will be switching over to the conda approach anyway. We can punt this buildbot effort/discussion over to issue #1056. 

I tried running this branch on a 2D dambreak problem and no issues regarding OpenMP/OpenBLAS arose. Once @tridelat gives this a shot as well I think we can go ahead and merge.

tridelat:
@zhang-alvin no problem on my side as well, after installing Proteus through conda, I ran a floating body case (so using Chrono as well) in series and in parallel: all went well
@cekees @davidbrochart

Issue #1040, opened 2019-12-18
Automatic documentation build


tridelat:
So right now, the "new" documentation is on a separate repository (proteus-doc), and the process to push documentation is not automatic at all. There are 3 ways to do it:
- Manual: build the intended version of proteus; build doc locally to specific folder; add changed files, new files, and remove unused files; commit; push to gh-pages branch.
- readthedocs: can automatically document Proteus, fully externally. Ideal, but would require an install of Proteus for it to work properly, and readthedocs build time limit is around 900s, and I don't think there is a way to point at an existing build (not 100% sure about that).
- Travis: can automatically build doc and push it to `gh-pages`, but only when someone commits to specific branch (such as master).

The last 2 require that the documentation is ported back to the main repository, and there is a way to do this without loosing commit authorship. I think the Travis solution might be the best one where a full installation of Proteus can be used for sphinx. This will make it very easy for anyone to add a page to the documentation and have it built as soon as it is committed to master.
The only issue I can see with that technique is that Travis could potentially time out when committing to master while it didn't on a PR because now it would also need to build the documentation.
Should we go that route?
@cekees

cekees:
I think once we get the kinks worked out of the conda package (https://github.com/erdc/proteus/pull/1037), then readthedocs may work:
https://docs.readthedocs.io/en/stable/guides/conda.html#activating-conda
Agree that travis is probably the most flexible approach.

PR #1041, opened 2019-12-18
Documentation: proteus-doc -> proteus

tridelat:
This PR brings the changes to documentation from proteus-doc to proteus, keeping commit authorship (but could not keep the date/time of changes due to some conflicts).

The reason for bringing back the documentation here is that it seems easier to have an automatic way to build it if it is centralised on this repo, see #1040 

This should **not** be squashed and merged (regular merge)


tridelat:
@cekees this is ready to be reviewed/merged (not squashed and merged so that people keep authorship of the documentation they modified).
It will not change the documentation website, just update the files so that the next time we update the docs, we get something like this: https://tridelat.github.io/proteus/
(I pushed the website generated from this branch on my fork)

tridelat:
@cekees as a side note, clicking on "Python API" leads you to this page: https://tridelat.github.io/proteus/api/modules.html, where you need to click on "proteus package" to actually access the API doc. This can be bypassed to access the proteus doc directly, although having it this way is "safer" if one day an extra package is installed side-by-side with proteus, as it would appear on that intermediate page.

cekees:
Found a few more issues: 1) the readthedocs template is  not scaling the inheritance graphs properly, 2)  the C++ docs are not showing ineritance graphs at all, and 3) the C++ docs aren't including the mathematical formulas that have been formatted for doxygen.

tridelat:
@cekees for 1) I will need to dig a little bit,
for 2) and 3) it looks like the only way is to switch back to  `doxygen` only, and not use `breathe`.
I think it might be best to just use doxygen, the only drawback is that it would be a "separate" API documentation, with a different template after clicking on "C/C++ API".
Breathe is supposed to link doxygen and sphinx nicely but it looks like it cannot reproduce the hierarchy graphs produced from doxygen.

tridelat:
@cekees for 3) the mathematical formulas actually seem to render, for example here: https://tridelat.github.io/proteus/capi/classDensityRelation.html#exhale-class-classdensityrelation
For the hierarchy/dot graphs, I don't think it is possible with breathe (yet at least), see https://github.com/michaeljones/breathe/issues/45
We can point directly to the doxygen html pages instead of using breathe if we want to inlcude the C++ graphs. It would look similar to the current https://proteustoolkit.org/capi/html/hierarchy.html (with no link back to https://proteustoolkit.org/)

PR #1042, opened 2019-12-19
Added basic chrono timestep modeling in text and h5 formats

dloney:
Adds chrono timestep logging for bodies and springs in text and h5 formats.

### DIFF
--- a/proteus/mbd/CouplingFSI.pxd
+++ b/proteus/mbd/CouplingFSI.pxd
@@ -297,6 +297,9 @@ cdef class ProtChSystem:
         bool initialised
         bool update_substeps
         int nBodiesIBM
+        object log_chrono_bodies
+        object log_chrono_springs
+        string log_chrono_format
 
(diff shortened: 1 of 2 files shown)


tridelat:
@dloney there is a function for looging of bodies (`ProtChBody`) here: https://github.com/erdc/proteus/blob/master/proteus/mbd/CouplingFSI.pyx#L1135-L1234
There isn't one for springs however

dloney:
The existing code records at the proteus time step. The new functionality logs on the chrono time step. I'd like to have the more fine resolution to keep track of the internal forces/velocities to better monitor for the systems going outside of their design limits since that can peak and drop back down by the next proteus time step.

tridelat:
@dloney I see, that sounds good, I think that it can go in like this, but we might want to separate some things from the `calculate` function into separate `prestep` and `poststep` functions if we need to add more things in the future

Issue #1044, opened 2019-12-26
Possible Rework of .h5-based Tests (because of PR #1039)


zhang-alvin:
Some issues are popping up in PR #1039, where tests are failing because comparisons are made against solution files computed from different meshes (due to version changes in the mesh generator). 

This leads to errors of the following type:
```    
def within_tol(x, y, atol, rtol):
        with errstate(invalid='ignore'):
>           return less_equal(abs(x-y), atol + rtol * abs(y))
E           ValueError: operands could not be broadcast together with shapes (599,) (711,)
```

One approach is to move towards metric-based tests. @manuel-quezada, you mentioned that some tests are already using a metric-based approach. Could you point me to some of those if you recall? 

I've looked into computing the L2 norms of scalars in the griffiths lane set of tests. It seems to be reasonable for the richards test, but I believe I'm seeing some compounding errors with the elastoplasticity test, where the test depends on the saved solution file for the richards problem. The result is that the L2 norms of each component in the displacement do not compare very well against the saved solution file (`elastoplastic_expected.h5`) when the mesh is changed. 

See https://github.com/erdc/proteus/blob/conda/proteus/tests/griffiths_lane_6/sm_gl_6_3d_p.py#L114

zhang-alvin:
The following is the current approach to generating comparison files for tests. 

```
        expected_path = 'comparison_files/' + 'comparison_' + name + '_phi_t2.csv'
        #write comparison file
        #np.array(actual.root.phi_t2).tofile(os.path.join(self._scriptdir, expected_path),sep=",")
        np.testing.assert_almost_equal(np.fromfile(os.path.join(self._scriptdir, expected_path),sep=","),np.array(actual.root.phi_t2).flatten(),decimal=10)
```

The following are the remaining .h5 files in the tests directory, which represent the tests that still need to be modified:

```
/AddedMass/addedMass2D.h5
./SWEs/dam_over_bumps/comparison_files/SWEs_dam_over_bumps.h5
./SWEs/test_reflecting_BCs/comparison_files/SWEs_test_reflecting_BCs.h5
./SWEs/oneD_dambreak_flat_bottom/comparison_files/SWEs_oneD_dambreak_flat_bottom.h5
./SWEs/test_gauges/comparison_files/SWEs_test_gauges.h5
./SWEs/paraboloid_with_friction/oneD/comparison_files/SWEs_oneD_paraboloid_with_friction.h5
[...]

PR #1045, opened 2019-12-27
Unpin dependencies

davidbrochart:


davidbrochart:
@cekees this is ready to be reviewed / merged.

davidbrochart:
@cekees the failing tests are the following. Any thoughts?
- `AddedMass/test_addedmass3D.py::TestAddedMass3D::test_AddedMass_3D`: seems to be due to `openblas` not being compiled with `USE_OPENMP=1` option.
- `CLSVOF/with_RANS3PF/test_clsvof_with_rans3pf.py::TestCLSVOF_with_RANS3PF::test_3D_falling_bubble`: it looks like `expected.root.phi_t2` and `actual.root.phi_t2` don't have the same shape (200 vs 299).
- `LS_with_edgeBased_EV/VOF/test_vof.py::TestVOF::test_EV2`: `expected.root.u_t2` and `actual.root.u_t2` are different.
- `TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_marin`: `expected.root.phi_t2` and `actual.root.phi` don't have the same shape (171 vs 193).
- `TwoPhaseFlow/test_TwoPhaseFlow.py::TestTwoPhaseFlow::test_moses`: `expected.root.phi_t2` and `actual.root.phi` don't have the same shape (237 vs 209).
- 'cylinder2D/ibm_rans2p_3D/test_cylinder3D_ibm_rans2p.py::Test_ibm::test_ex2`: `expected.root.u_t2` and `actual.root.u_t2` don't have the same shape (2573 vs 1676).
- `cylinder2D/sbm_3Dmesh/test_cylinder2D_on_3D_mesh_sbm.py::Test_sbm_cylinder2D_on_mesh3D::test_ex1`: `expected.root.u_t2` and `actual.root.u_t2` don't have the same shape (1938 vs 988).
- `griffiths_lane_6/test_griffiths_lane6.py::TestRichards::test_richards`: `expected.root.pressure_head_t1` and `actual.root.pressure_head_t1` don't have the same shape (2181 vs 1158).
- `griffiths_lane_6/test_griffiths_lane6.py::TestRichards::test_elastoplastic`: files `gl_6_3d.poly` and `gl_6_3d.smesh` cannot be accessed by `tetgen`.
- `levelset/vortex/test_vortex.py::TestVortex3D::test_vortex3D`: `tetgen` error while opening `tank3d.edge`: `Edge 1 has an invalid vertex index`.

zhang-alvin:
@davidbrochart The set of tests are generally failing because the tests make a comparison of a generated solution file against a saved solution file with a different mesh.  A version change in gmsh and/or tetgen is enough to cause this issue. I've described it with some more detail in #1039 and am looking at an alternative way to do the tests with #1044.

davidbrochart:
Thanks @zhang-alvin. This PR solves a number of issues. The number of failing tests dropped from 19 to 9. I guess it can be merged to the `conda` branch.

zhang-alvin:
> Thanks @zhang-alvin. This PR solves a number of issues. The number of failing tests dropped from 19 to 9. I guess it can be merged to the `conda` branch.

@davidbrochart Yes, I think a merge would be fine. Out of curiousity, could you elucidate the switch in the `Makefile` from `pip3` back to `pip`?

davidbrochart:
`pip` is installed through `conda`, and defaults to the matching Python version installed, which is Python 3. There is no need to specify `pip3`, and actually there is no `pip3` installed.

zhang-alvin:
@davidbrochart Another question: it looks like with this branch, the `hashdist` approach to compiling proteus will be entirely removed from the `Makefile`?

davidbrochart:
Yes, that is the point of switching to a conda-based installation, right?
Or would you like to keep it as an alternative installation?

zhang-alvin:
> Yes, that is the point of switching to a conda-based installation, right?
> Or would you like to keep it as an alternative installation?

My primary concern is along the lines of ... 
1) will the conda approach compile on HPC machines with minimal modifications? 
2) in the event of a new library or new version of an existing dependency being needed, is there a quick way to test it out locally, or does a developer have to put the package onto conda-forge?

davidbrochart:
You are right, we should not get rid of `hashdist`, mostly because of 1.
1. the conda approach will probably work on HPC machines (as long as they run under e.g. Linux and an Intel architecture), but it will not be optimal in terms of performance.
2. new libraries/versions should also be packaged through conda, but it can be done locally before pushing to `conda-forge`.

I will put back `hashdist` in the `Makefile`.

davidbrochart:
@zhang-alvin I restored `hashdist`, and created 2 new targets in the `Makefile`: `develop-conda` and `test-conda`.

zhang-alvin:
> @zhang-alvin I restored `hashdist`, and created 2 new targets in the `Makefile`: `develop-conda` and `test-conda`.

Thanks @davidbrochart . I think it looks good. I'll go ahead and squash merge this PR into conda.

PR #1046, opened 2019-12-27
[WIP] Switch select tests to use premade mesh

zhang-alvin:
Attempt at tackling issue #1044 and related to resolving tests on PR #1039. 

The idea is to store a very small mesh (~4KB for each file describing the mesh) and pass that as a fixed input to the case, which allows us to better test if the numerical machinery has changed. The resulting solutions are compared with an L2-norm comparison of a relevant scalar field, thereby removing the need to store larger .h5 files.

Initial attempt at implementation was done for the Griffiths lane set of tests.


tridelat:
@zhang-alvin nice to fix the mesh for tests, that's much more robust.
And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.

zhang-alvin:
> @zhang-alvin nice to fix the mesh for tests, that's much more robust.
> And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
> Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
> It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.

I think the use of gauges is not bulletproof either. Gauges are more specific than a global L2-norm, but the value at a given gauge is not uniquely defined by the nodal values of the enclosing element. I think if there are options to only store the nodal values for the .h5 file at a desired time step, then we can store .h5 files. Otherwise, there's too much unnecessary information stored in the .h5 files.

tridelat:
> > @zhang-alvin nice to fix the mesh for tests, that's much more robust.
> > And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
> > Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
> > It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.
> 
> I think the use of gauges is not bulletproof either. Gauges are more specific than a global L2-norm, but the value at a given gauge is not uniquely defined by the nodal values of the enclosing element. I think if there are options to only store the nodal values for the .h5 file at a desired time step, then we can store .h5 files. Otherwise, there's too much unnecessary information stored in the .h5 files.

Gauges are not bulletproof, but definitely more robust.
I am not sure what you mean by "[the value is] not uniquely defined by the nodal values of the enclosing element", as value at a gauge is actually interpolated from nodal values of enclosing element.
Maybe I am nitpicking, but it is easy to imagine test cases that would pass with L2-norm but that would fail with Gauges.

For example, a quiescent `[1.,1.]` rectangular tank with horizontal free surface in the middle: if you want to check the VOF field, and you use a smooth heaviside for VOF around the free surface, if you use a width of say `0.1` to smooth it on one case, and then `0.2` to smooth it on another case, then you would get the same L2-norm for this domain.
Gauges around the free surface would have caught that the VOF field changed due to numerics changing (here the heaviside smoothing length).

zhang-alvin:
> > > @zhang-alvin nice to fix the mesh for tests, that's much more robust.
> > > And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
> > > Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
> > > It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.
> > 
> > 
> > I think the use of gauges is not bulletproof either. Gauges are more specific than a global L2-norm, but the value at a given gauge is not uniquely defined by the nodal values of the enclosing element. I think if there are options to only store the nodal values for the .h5 file at a desired time step, then we can store .h5 files. Otherwise, there's too much unnecessary information stored in the .h5 files.
> 
> Gauges are not bulletproof, but definitely more robust.
> I am not sure what you mean by "[the value is] not uniquely defined by the nodal values of the enclosing element", as value at a gauge is actually interpolated from nodal values of enclosing element.
> Maybe I am nitpicking, but it is easy to imagine test cases that would pass with L2-norm but that would fail with Gauges.
> 
> For example, a quiescent `[1.,1.]` rectangular tank with horizontal free surface in the middle: if you want to check the VOF field, and you use a smooth heaviside for VOF around the free surface, if you use a width of say `0.1` to smooth it on one case, and then `0.2` to smooth it on another case, then you would get the same L2-norm for this domain.
> Gauges around the free surface would have caught that the VOF field changed due to numerics changing (here the heaviside smoothing length).

I can definitely see the L2 norm approach as not being very sensitive to numerical changes. In your case, if the volume of the quiescent tank was much larger than the area of the free surface. I don't see why the L2 norm of the VOF field would yield exactly the same results with either blending width, but it might be fairly close.

tridelat:
> > > > @zhang-alvin nice to fix the mesh for tests, that's much more robust.
> > > > And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
> > > > Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
> > > > It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.
> > > 
> > > 
> > > I think the use of gauges is not bulletproof either. Gauges are more specific than a global L2-norm, but the value at a given gauge is not uniquely defined by the nodal values of the enclosing element. I think if there are options to only store the nodal values for the .h5 file at a desired time step, then we can store .h5 files. Otherwise, there's too much unnecessary information stored in the .h5 files.
> > 
> > 
> > Gauges are not bulletproof, but definitely more robust.
> > I am not sure what you mean by "[the value is] not uniquely defined by the nodal values of the enclosing element", as value at a gauge is actually interpolated from nodal values of enclosing element.
> > Maybe I am nitpicking, but it is easy to imagine test cases that would pass with L2-norm but that would fail with Gauges.
> > For example, a quiescent `[1.,1.]` rectangular tank with horizontal free surface in the middle: if you want to check the VOF field, and you use a smooth heaviside for VOF around the free surface, if you use a width of say `0.1` to smooth it on one case, and then `0.2` to smooth it on another case, then you would get the same L2-norm for this domain.
> > Gauges around the free surface would have caught that the VOF field changed due to numerics changing (here the heaviside smoothing length).
> 
> I can definitely see the L2 norm approach as not being very sensitive to numerical changes. In your case, if the volume of the quiescent tank was much larger than the area of the free surface. I don't see why the L2 norm of the VOF field would yield exactly the same results with either blending width, but it might be fairly close.

The L2-norm would be the same with the two domains with different blending widths in that case because the total VOF would be the same in the two domains.

tridelat:
> > > > @zhang-alvin nice to fix the mesh for tests, that's much more robust.
> > > > And now that the resulting mesh file is very coarse and small, is it not possible to still compare with the (much smaller in theory) `.h5` file? These .h5 could still be stored with git lfs, as tests are not performed locally on a regular basis by users (but they can git lfs pull if needed).
> > > > Otherwise, if it is still too big, Gauges could be used to store smaller .csv files at specific points to compare.
> > > > It's just that the L2-norm is not fully bullet proof, as tests can technically pass with different fields as long as the norm is the same.
> > > 
> > > 
> > > I think the use of gauges is not bulletproof either. Gauges are more specific than a global L2-norm, but the value at a given gauge is not uniquely defined by the nodal values of the enclosing element. I think if there are options to only store the nodal values for the .h5 file at a desired time step, then we can store .h5 files. Otherwise, there's too much unnecessary information stored in the .h5 files.
> > 
> > 
> > Gauges are not bulletproof, but definitely more robust.
> > I am not sure what you mean by "[the value is] not uniquely defined by the nodal values of the enclosing element", as value at a gauge is actually interpolated from nodal values of enclosing element.
> > Maybe I am nitpicking, but it is easy to imagine test cases that would pass with L2-norm but that would fail with Gauges.
> > For example, a quiescent `[1.,1.]` rectangular tank with horizontal free surface in the middle: if you want to check the VOF field, and you use a smooth heaviside for VOF around the free surface, if you use a width of say `0.1` to smooth it on one case, and then `0.2` to smooth it on another case, then you would get the same L2-norm for this domain.
> > Gauges around the free surface would have caught that the VOF field changed due to numerics changing (here the heaviside smoothing length).
> 
> I can definitely see the L2 norm approach as not being very sensitive to numerical changes. In your case, if the volume of the quiescent tank was much larger than the area of the free surface. I don't see why the L2 norm of the VOF field would yield exactly the same results with either blending width, but it might be fairly close.

You are right, it is very unlikely that the L2-norm would be exactly the same with the smoothing actually. 
In another case where the free surface is a sharp interface (either 1. or 0. for VOF, no smoothing), and separating the two-phases so that they have the same total area, no the L2-norm here would be the same different possible positions of the free surface (e.g. vertical or horizontal if the domain is square).

cekees:
On the gauges, they do build the FEM interpolation consistently within the element so if the mesh doesn't change and the solution doesn't change, the gauge values shouldn't change.  In general, however, the gauges are designed for comparing to real data where it makes sense to have a fairly loose tolerance in the comparison of gauge values to real data (or at least a tolerance that is approved by the expert for the case e.g. for a wave experiment).

zhang-alvin:
@cekees  After some back and forth with @tridelat, I think we've settled on saved mesh as input, no gauges, and I'm going to look into saving the nodal field values at a specific time to make a comparison. If that yields reasonable file sizes, I'll do that for the 9 tests that have been failing for the conda PR #1039 . 

I can tackle migrating the remainder of the .h5 comparison tests over the course of the next few weeks once the conda PR is merged.

cekees:
Sounds like a good plan. I went ahead and approved so you can merge it into erdc:conda when you're ready or create a new PR with @davidbrochart

PR #1047, opened 2019-12-31
Fix MPI for multi-threading

davidbrochart:


davidbrochart:
In conda-forge, `scotch` has been compiled with `SCOTCH_PTHREAD`, so either `MPI_Init()` must be called with `MPI_THREAD_MULTIPLE`, or `MPI_Init_thread()` must be called. At import time, `mpi4py` initializes the MPI execution environment calling `MPI_Init_thread()`. `mpi4py` must be imported before `petsc4py` otherwise `petsc4py` initializes MPI with `MPI_Init()`.

PR #1049, opened 2020-01-07
Travis Build Speedup with build_ext

zhang-alvin:
Attempt at speeding up Travis build by explicitly calling `python setup.py build_ext` to make use of threads as per @cekees suggestion.


zhang-alvin:
Going to close this PR since no speedup was observed in build process. See #1039 for more in-depth discussion if this ever comes up again.

PR #1050, opened 2020-01-09
Modify Doxygen / Breathe / Exhale workflow

tridelat:
- use doxygen without exhale (memory problem with exhale)
- still use breathe to allow to manually include doxygen definitions in sphinx .rst
- link to sphinx for python and doxygen for c++
- build  in ./build instead of ../../proteus-website


tridelat:
@cekees done for the copyright, the version on Doxygen was already updated but did not render on the pages because I did not regenerate the docs

PR #1051, opened 2020-01-13
Migrate Mprans to xtensor

JohanMabille:
This PR is based on #1039 and should be merged after.

I open it early so it is possible to follow the changes. The idea is to make a new "standalone" mprans module based on xtensor/xtensor-python with minimal changes in the implementation.

Instead of directly changing mprans, the new implementation is done in mprans2 so it remains easy to compare both of them. Once the implementation is done, the build will swap mprans and mprans2 to test and fix the issues.

Once it has been merged, I will iterate on cleaning the code.


JohanMabille:
Closing it, will push directly to #1504

PR #1065, opened 2020-01-21
Docs for 1.7.2

tridelat:


tridelat:
@cekees added info for installing through conda + autodoc inheritance is now scaling properly

Issue #1067, opened 2020-01-22
TPF: __init__ functions for models has unnecessary logic


zhang-alvin:
Based on discussions with @tridelat 

In the initialization of some models' coefficients class, like `RANS2P.Coefficients`, there is 
already logical evaluations based on the model attributes:

https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS2P.py#L238-L275

This means that the class cannot be instantiated without appropriate attributes being set. This makes the code harder to read and is less user-friendly than necessary.

The proposal is to move such logical statements under a new function, e.g. `RANS2P.Coefficients.start()` after the user has specified the desired attributes. This has the added bonus of eliminating the need for the `CoefficientsOptions` class.

Such a change can be applied to all models.

tridelat:
Closing as this was addressed with #1070

PR #1070, opened 2020-01-23
TPF: Make Coefficients class directly accessible

tridelat:
Attempt at making `Coefficients` class instances directly accessible from the TwoPhaseFlow parameters.
This aims at removing the redundant `CoefficientsOptions`. This PR is related to #1067.
- All logic moved from `Coefficients` classes (e.g. RANS2P, VOF, etc) moved to function separated from `__init__`
- separated function automatically called (unless explicitly asked otherwise) in order to keep functional the classic way of instantiating `Coefficients`

Changes for TwoPhaseFlow:
- Coefficients are directly accessed with `[mymodel_parameters].p.coefficients` instead of `[mymodel_parameters].p.CoefficientOptions` 
- Attributes in `coefficients` are not "frozen" (could look for a way to freeze it again when using TwoPhaseFlow)


proteus/TwoPhaseFlow/utils/Parameters.py
```diff
@@ -489,7 +433,7 @@ def _initializeNumerics(self):
         # TIME
         if self.timeDiscretization=='vbdf':
             self.n.timeIntegration = TimeIntegration.VBDF
-            self.n.timeOrder = self.p.CoefficientsOptions.timeOrder
+            self.n.timeOrder = 2
```
zhang-alvin: Is this always true? I was under the impression that `vbdf` was a scheme such that different orders can be used?

proteus/TransportCoefficients.py
```diff
@@ -198,6 +198,12 @@ def __init__(self,
             self.elementIntegralKeys.append(('u',ci))
             self.elementBoundaryIntegralKeys.append(('u',ci))
 
+    def __getitem__(self, key):
```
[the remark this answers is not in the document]
tridelat: That is just to make class attributes accessible as if they were dictionaries, and this was done for compatibility with old tests

proteus/TwoPhaseFlow/utils/Parameters.py
tridelat: Not necessarily, let me correct that

tridelat:
> This looks fine to me.
> 
> My understanding is that most of this PR is that there's a new `initialize` function that is (by default) called during model instantiation. So if the user wants to instantiate a model like `RANS2P` without supplying all of the necessary parameters, then he/she would have to pass in the argument `initialize=False`. This makes the changes backward compatible with the previous way of doing things.
> 
> Regarding freezable attributes, we might want to have that functionality accessible to a wide variety of user-exposed functions through an abstract class beyond the TwoPhaseFlow app.

That's right, changes are backward compatible, and all logic that needs predefined attributes are moved in a separate `initialize()` functions, which is called automatically if when `initialize=True` (default), or manually when `initialize=False` like in the TwoPhaseFlow app so that everything is initialized only when all the parameters have been defined

zhang-alvin:
did you want to merge this into `master` or `develop/TPF`?

tridelat:
@zhang-alvin I was thinking about putting it into `master` as this was branched out of `master` and the tests are passing, then we can merge into `develop/TPF` and fix conflicts there, let me know if you prefer otherwise

Issue #1069, opened 2020-01-23
TPF: proposal for refactoring initial conditions


zhang-alvin:
The current method to set initial conditions is through a user-specified dictionary

```
theProblem.modelDict['flow'].p.initialConditions={0: zero(),
                            1: zero(),
                            2: zero()}
theProblem.modelDict['vof'].p.initialConditions={0: VF_IC()}
theProblem.modelDict['ls'].p.initialConditions={0: PHI_IC()}
theProblem.modelDict['rd'].p.initialConditions={0: PHI_IC()}
```
The user has to know that for the `rans2p` or `rans3p` models, 0 corresponds to pressure. The use of numerical keys for the dictionary doubles as indices when setting the `numpy arrays` for the dofs in https://github.com/erdc/proteus/blob/master/proteus/Transport.py#L1627-L1650.

To make this more user friendly, we can ...
1. force the user to use specific keywords as keys
```
theProblem.modelDict['flow'].p.initialConditions={"pressure": zero(),
                            "u": zero(),
                            "v": zero()}
```
We can then keep a separate dictionary to map the keywords to indices to use the current approach to set the `numpy` arrays
```
idxDict = {"pressure": 0, "u": 1, "v": 2} 
```

2. Alternatively, we can define an `InitialCondition` class for each model which forces the user to populate specific attributes:

```
        class InitialCondition(FreezableClass):
            def __init__(self):
                self.pressure = None
                self.u = None
                self.v = None
                self.w= None
       

        self.p.initialConditions = InitialCondition()
        ...
        self.p.initialConditions._freeze()
```
user would set pressure condition as follows:
```
theProblem.modelDict['flow'].p.initialConditions.pressure=zero()
```
We can then use the internal class dictionary (e.g. `self.p.initialConditions.__dict__["pressure"]`) along with an index dictionary as with the first approach to populate the `numpy` arrays. The advantage with the second approach is that the freezing capability prevents the user from adding/defining attributes and throws an error very quickly.

zhang-alvin:
Handled in #1035 which went into #1038

PR #1071, opened 2020-01-27
[WIP] TPF: sketch of initial conditions refactoring

zhang-alvin:
This PR is a draft of how the initial condition refactoring might be done in TPF as per issue #1069.

The idea is to force the user to supply initial conditions for each model based on hard-coded keywords. 

This deviates from the previous approach in that user now cannot use integers to associate an initial condition with a model variable, e.g. `{0: foo()}` would now be `{'p': foo()}` to denote the pressure initial condition.

I've only done the modifications necessary to get the `RANS2P` model compliant to solicit some feedback as it would take some time to implement this across the board.


zhang-alvin:
@cekees and @tridelat would be good to have a brief look before the developer call.

tridelat:
@zhang-alvin @cekees sounds good, and using the variableNames of the Coefficients class would be nice. One thing to think about is that some variableNames need to already know `nd` in order to be set with the right number of variables (e.g. https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS2P.py#L379 ).
In practice this is not known until the TwoPhaseFlowProblem is "initialized" as it depends on its `domain` variable.

tridelat:
@zhang-alvin do you have a backup of this? It is from a deleted repository ("unknown repository") right now

zhang-alvin:
> @zhang-alvin do you have a backup of this? It is from a deleted repository ("unknown repository") right now

Yes I have a backup of this, I'll close this PR and reopen shortly.

zhang-alvin:
See #1194 for continuation

PR #1075, opened 2020-01-29
Update master branch with adaptation trigger logic from other branch

zhang-alvin:
Cherry-picked features from other branch that includes adaptation trigger logic. 

Contingent on update of the `scorec` feedstock on conda-forge: https://github.com/conda-forge/scorec-feedstock/pull/2 

The `hashdist` stack likely needs to be updated as well.


proteus/MeshAdaptPUMI/cMeshAdaptPUMI.cpp
```diff
@@ -305,6 +315,14 @@ int MeshAdaptPUMIDrvr::getSimmetrixBC()
...
   return 0;
 } 
+static int countTotal(apf::Mesh* m, int dim)
+{
+  int total = apf::countOwned(m, dim);
+  PCU_Add_Ints(&total, 1);
+  return total;
+}
+
+#include "PyEmbeddedFunctions.h"
```
cekees: This may just me being old-fashoned, but I like having all the headers at the top of the file to make it immediately clear about the dependencies.

proteus/MeshAdaptPUMI/createAnalyticGeometry.cpp
```diff
@@ -668,7 +668,7 @@ gmi_model* MeshAdaptPUMIDrvr::createSphereInBox(double* boxDim,double*sphereCent
   }
   m->end(it);
 
-  gradeMesh();
+  gradeMesh(1.5);
```
cekees: We should eventually push this up to a parameter definition.

proteus/flcbdfWrappersModule.cpp
```diff
@@ -0,0 +1,7295 @@
+#define FLCBDF_WRAPPERS_MODULE
```
cekees: This file was removed when we completed the transition to Cython wrappers. Is there stuff in here for PUMI that we need to port over to the new setup?

proteus/MeshAdaptPUMI/cMeshAdaptPUMI.cpp
zhang-alvin: I think that perspective is still in vogue, header was updated.

proteus/MeshAdaptPUMI/createAnalyticGeometry.cpp
zhang-alvin: Yes, this will likely happen during the broader refactoring of MeshAdaptPUMI. I'll add an issue for this so it won't be forgotten (see #1081)

proteus/flcbdfWrappersModule.cpp
zhang-alvin: File was removed once again. Not sure why it came back.

zhang-alvin:
Something's strange with the Travis environment and the scorec package. It compiles locally.

Issue #1078, opened 2020-02-01
conda-forge gmsh version updated, gmsh tests failing 


zhang-alvin:
gmsh tests are still too rigid. We can increase the threshold, but for the 2D case, there's a factor of 10 difference in the expected value.

```
=================================== FAILURES ===================================
_______________________ TestGMSH.test_gmsh_generation_2D _______________________
[gw0] linux -- Python 3.7.6 /home/travis/miniconda/envs/proteus-dev/bin/python
self = <proteus.tests.test_gmsh_generation.TestGMSH testMethod=test_gmsh_generation_2D>
    def test_gmsh_generation_2D(self):
        domain = Domain.PlanarStraightLineGraphDomain()
        domain.vertices = [[0., 0., 0.],
                           [5., 0., 0.],
                           [5., 5., 0.],
                           [0., 5., 0.]]
        domain.segments = [[0, 1], [1, 2], [2, 3], [3, 0]]
        domain.facets = [[[0, 1, 2, 3]]]
        domain.writeGeo('gmsh_mesh_test_2D', he_max=0.1)
        gmsh_cmd = "gmsh {0:s} -v 10 -2 -o {1:s} -format msh2".format(domain.geofile+".geo", domain.geofile+".msh")
        check_call(gmsh_cmd, shell=True)
        MeshTools.msh2simplex(domain.geofile, nd=2)
        #with open('gmsh_mesh_test_2D.node', 'r') as nodefile:
        #    npt.assert_equal(nodefile.readline(), '3433 2 0 1\n')
        #with open('gmsh_mesh_test_2D.edge', 'r') as edgefile:
        #    npt.assert_equal(edgefile.readline(), '10096 1\n')
        #with open('gmsh_mesh_test_2D.ele', 'r') as elefile:
        #    npt.assert_equal(elefile.readline(), '6664 3 1\n')
        with open('gmsh_mesh_test_2D.node', 'r') as nodefile:
            numNodes = int(nodefile.readline().split(' ', 1)[0])
>           assert abs(1.0-old_div(numNodes,numNodes_reference_2D)) < THRESHOLD
E           AssertionError: assert 0.12088552286629772 < 0.02
E            +  where 0.12088552286629772 = abs((1.0 - 0.8791144771337023))
E            +    where 0.8791144771337023 = old_div(3018, 3433.0)
proteus/tests/test_gmsh_generation.py:52: AssertionError
----------------------------- Captured stdout call -----------------------------
[...]

PR #1079, opened 2020-02-03
TST: loosen gmsh test comparison

zhang-alvin:
Address issue #1078 

The comparison value for the gmsh test was changed. Tests will pass even if there's up to 15% deviation from the reference value.

### DIFF
--- a/proteus/tests/test_gmsh_generation.py
+++ b/proteus/tests/test_gmsh_generation.py
@@ -16,7 +16,8 @@ logEvent("Testing Gmsh Mesh Conversion")
 
 
 #TEST VARIANCE THRESHOLD, 2%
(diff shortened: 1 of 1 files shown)


zhang-alvin:
> Looks good. Out of curiosity, anywhere below 0.15 resulted in a fail?

The assertion error in the 2D case was because there was a 12% deviation. I simply rounded to 15%. I'm not sure if there's significant value in having the test in its current form as its possible for the test to break further down the line for the same reasons (node count deviation becomes 20% instead of 15%). 

We can move towards either having the test just check whether a mesh is generated or consider only the element count which is at least bounded by the domain volume.

zhang-alvin:
Merging this in so we can move ahead with other PRs.

PR #1083, opened 2020-02-05
Changed flcbdfWrappers.pxd file for OSX fix

ejtovar:
Moved the block 

`cdef extern from "FullDataFile.h" namespace "Daetk":                                                                                                                                          
    cdef cppclass FullDataFile:                                                                                                                                                               
        FullDataFile(const double t0, const char* filename)`

to the top of the flcbdfWrappers.pxd for fix on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,9 @@ jobs:
       - conda env create -f environment-dev.yml
       - conda activate proteus-dev
       - pip install -v -e .
+    script:
+      - export MPLBACKEND="AGG"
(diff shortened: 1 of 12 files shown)


cekees:
@davidbrochart and @ejtovar this builds and runs fine on both osx (conda) and linux (conda and hashdist), but pytest is having some sort of issue on the osx build environment on travis. @ejtovar can you run pytest locally as it's run in  .travis.yml?

cekees:
@tridelat and @zhang-alvin this adds an osx stage to travis and fixes a weird header ordering issue with the pxd. Also, it updates the stack to the latest master, which I've been using for months. I think it can go in.

PR #1084, opened 2020-02-05
Add Travis CI deploy workflow

tridelat:
- [x] Make Travis build docs automatically
- [x] Make Travis push docs to external repo (needs read/write access)
- [x] Make sure it happens only when code is committed to a specific branch (`master`)
- [x] Make sure that the deployment happens only on one job

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ after_success:
 
 jobs:
   include:
-  - stage: compile
-    name: "proteus-hashdist install"
+  - stage: test
+    name: "hashdist"
     python: 2.7
     env: TEST_PROFILE="proteus-hashdist"
     install:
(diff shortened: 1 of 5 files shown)


cekees:
Thanks for adding this. Will we still push the release docs to RTD manually?

tridelat:
@cekees as releases also come from `master` and docs would be rebuilt automatically with every commits to `master`, the idea would be that we would not have to push any new documentation files manually when there is a release, but we would only have to:
- create a new branch (e.g. 1.7.3) on https://github.com/erdc/proteus-docs from the branch where the latest (release) docs were automatically built (e.g. gh-pages)
- let RTD know about that new branch

It might actually be possible to have Travis automatically create new branches on proteus-docs when there is a new tagged release, but we would still have to tell RTD about it

tridelat:
@cekees @zhanga This is ready for review, it will allow for automatic generation of the docs every time there is a new commit in master, and push it to the proteus-docs master. I have added some safety checks to ensure that it adds, commits, and pushes it to the right repo

PR #1089, opened 2020-02-10
Hotfix for MCorr when using TPF

tridelat:
Mini PR to fix linkage between MCorr model and LS model when using TPF to fix bug encountered by @msklia and @BranocRichards

### DIFF
--- a/proteus/TwoPhaseFlow/utils/Parameters.py
+++ b/proteus/TwoPhaseFlow/utils/Parameters.py
@@ -437,7 +437,7 @@ class ParametersModelRANS2P(ParametersModelBase):
             self.n.timeIntegration = TimeIntegration.BackwardEuler_cfl
         else:
             raise ValueError("{scheme} scheme is not valid. Accepted schemes values are 'be' and 'vbdf'".format(scheme=self.timeDiscretization))
- 
+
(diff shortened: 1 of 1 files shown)


proteus/TwoPhaseFlow/utils/Parameters.py
```diff
@@ -1641,6 +1641,7 @@ def _initializePhysics(self):
         coeffs.flowModelIndex = V_model
         coeffs.me_model = ME_model
         coeffs.VOFModelIndex = VOF_model
+        coeffs.levelSetModelIndex = LS_model
```
zhang-alvin: This is the only real change right? I think it's fine, although the naming schemes seem to not all be uniform. flowModel and levelSet is camel case, me_model is underscored, and VOFModel is neither.

tridelat: Yes, this is because that is how they are defined in the Coefficients class, see `proteus/mprans/MCorr.py`. This is a recurrent problem across models, those names for storing indices are rather inconsistent (some are `ME_model`, `me_model`, or `modelIndex` for  designing their own index for example..)

zhang-alvin: I see. That's fine then for now.

PR #1091, opened 2020-02-11
Submodules: relative paths converted to absolute paths

tridelat:
This fixes issues of relative paths when cloning a fork of proteus, as the submodules are not necessarily forked with it (and therefore won't be found at e.g. `../air-water-vv`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "air-water-vv"]
 	path = air-water-vv
-	url = ../air-water-vv.git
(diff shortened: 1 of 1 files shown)


cekees:
I can't get on there and test, unfortunately.  You  don't have to do a build, though, just try to clone and run `git submodule update --init --recursive`. I tend to think we should just go ahead and merge this, however. It can be manually handled for the hpc build.

zhang-alvin:
I gave the modifications a try on an HPC system, and `git submodule update --init --recursive` does work without issue.

PR #1094, opened 2020-02-14
TestPR again to check if tests fail on Travis

zhang-alvin:


zhang-alvin:
Cleanup didn't break tests.

PR #1098, opened 2020-02-19
Add .gitattributes that were not initially included 

zhang-alvin:
Running tests can lead to unintended behavior with some comparison files because the .gitattributes files were not added.

### DIFF
--- a/proteus/tests/SWEs/dam_over_bumps/.gitattributes
+++ b/proteus/tests/SWEs/dam_over_bumps/.gitattributes
@@ -1,0 +1,6 @@
+hump.poly filter=lfs diff=lfs merge=lfs -text
+hump.1.neigh filter=lfs diff=lfs merge=lfs -text
+hump.1.poly filter=lfs diff=lfs merge=lfs -text
+hump.edge filter=lfs diff=lfs merge=lfs -text
+hump.ele filter=lfs diff=lfs merge=lfs -text
+hump.node filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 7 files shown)


proteus/tests/cylinder2D/ibm_method/mesh.edge
```diff
@@ -1,3 +1,335 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:3f9c96cdbab38fd8e3a2ee647c41a63731f836028382b1b692460530757c1680
-size 5840
+333  1
```
cekees: Why put this one under git? Just curious.

zhang-alvin: Actually, this file was inadvertently added. It's not supposed to be here.

zhang-alvin: It was the same error you ran into in which git auto added the file as one that needed to be committed.  

Force pushed same commit but omitting mesh.edge

zhang-alvin:
`git status` with hashdist build gives: 

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   proteus/tests/solver_tests_mprans/comparison_files/comparison_twp_navier_stokes_cavity_2d_p_t1.csv

Untracked files:
```

`comparison_twp_navier_stokes_cavity_2d_p_t1.csv` is not tracked by git-lfs so there are no other files not being accounted for by missing `.gitattributes` files.

PR #1099, opened 2020-02-19
Add TPF + FSI test

tridelat:
Adds a two-phase flow test using FSI so that more models are covered


proteus/mprans/MoveMesh.py
```diff
@@ -6,8 +6,10 @@
 from proteus.mprans.cMoveMesh import *
```
zhang-alvin: A couple of import * statements

tridelat:
> This looks fine. I would check if any of the new mesh files are over 100kB as I don't think they need to be any larger than that.
> 
> Not sure what's happening with the failing tests though.

@zhang-alvin forgot to reply but I checked before merging and the mesh files are all <100kB

PR #1100, opened 2020-02-19
Isolating MeshAdaptPUMI from NumericalSolution

zhang-alvin:
There's a lot of unnecessary code in NumericalSolution from the MeshAdapt feature. This PR is to move that code away from NumericalSolution to make it easier to maintain and debug.

This requires turning on the MeshAdaptPUMI tests again to ensure the post-adapt solution transfer is functional.


cekees:
This is a good idea. There's a lot of accumulated complexity outside of the adaptivity as well, so maybe I can do a pass after you complete this project.

zhang-alvin:
> This is a good idea. There's a lot of accumulated complexity outside of the adaptivity as well, so maybe I can do a pass after you complete this project.

Yes I think one candidate for migrating out of NumericalSolution is the mesh generation logic.

PR #1102, opened 2020-02-25
MBD: chrono 4.0.0 -> 5.0.0

tridelat:
PR to update files in proteus/mbd to chrono 5.0.0

## To do

- [x] update files
- [x] update stack (see https://github.com/erdc/stack/pull/48)
- [x] re-enable / find workaround for AM logic with ChMatrixDynamic
- [x] make sure tests pass

## Issues/Comments (checked=solved):
- [x] Needs eigen3 headers
- [x] Some class member variables became private
- [x] Segmentation fault when using FEA things such as cables (works fine on 4.0.0), happens on `ChSystem.DoStepDynamics(dt)`. It looks like this happens also when setting a simple pure python case using pychrono (!)
- [x] some unrelated tests that do not use any chrono logic or import fail


tridelat:
This is almost ready, the dependencies build fine on hashdist thanks to the new buildbot worker using and caching the latest stack `master`, but Travis CI runs out of memory when trying to compile Proteus itself.
Not sure why this happens, but maybe we somehow reached the limit allocated on Travis CI
@cekees @zhang-alvin

tridelat:
@cekees This is ready to go

@davidbrochart we will switch to 5.0.0 once this is merged in

tridelat:
> Looks fine to me. There is some functionality commented out (e.g. https://github.com/erdc/proteus/pull/1102/files#diff-7661a4ab7bcf212acbebb2de5fc83d33R164), but I'm guessing you know that.

Yes, the commented-out lines are not necessary anymore. Also it might be worth it to phase out the cython declarations here and use pybind11 instead

PR #1103, opened 2020-02-26
Switch back to dev

tridelat:
Update version number after release 1.7.3

cc @zhang-alvin


tridelat:
> What's the reasoning behind using the 'dev' suffix?

@zhang-alvin it is to indicate that it is a development version, and not an officially tagged/release version (which means that it is theoretically less stable)

tridelat:
We have a problem with the tests on conda here since some packages were likely updated in conda-forge. All tests pass on hashdist.

I am not necessarily familiar with the specific tests that are failing, but the difference between "actual" and "expected" seems a little bit too big. For example, one of the tests in `proteus/tests/CLSVOF/disc_ICs/test_CLSVOF_discICs.py` needs to go from a comparison of 10 decimal points (probably overkill) to 2 decimal points (probably too little) in order to pass. Some other tests have very different values from the expected ones (for example in `proteus/tests/post_processing`). This seems like it could be too much to be attributed to slight numerical changes due to updated dependencies.
As @zhang-alvin mentioned, this might be due to the coarsening of some tests that now use a fixed mesh, probably from https://github.com/erdc/proteus/pull/1046

\cc @cekees @zhang-alvin @davidbrochart

tridelat:
Merging as hashdist tests are passing and #1107 will fix conda tests issues

PR #1105, opened 2020-03-02
Residuals

dloney:
Adds residual logging of chrono link lock joints into proteus on the chrono timestep. The assumption is that each joint has all 6 DOF, so uninitialized values are entered for the remaining values where the joint is less constrained.


proteus/mbd/CouplingFSI.pyx
```diff
@@ -1576,7 +1577,14 @@ cdef class ProtChSystem:
             else:
                 self.log_springs_text(self.t + self.proteus_dt, self.log_chrono_springs)
 
-        if (self.log_chrono_bodies or self.log_chrono_springs) and self.log_chrono_format == 'h5':
+        if self.log_chrono_residuals:
```
tridelat: @dloney just to make sure, is the `log_chrono_residuals` list initialised manually from the main file (where the simulation setup is defined)?

dloney: Yep. The user creates the list of link lock objects, and anything contained in the list will have the residuals output.

PR #1106, opened 2020-03-04
Making MeshAdaptPUMI more flexible with models

zhang-alvin:
In the current form of MeshAdaptPUMI, there are assumptions made regarding the models that are used to drive interface-based adaptivity as well as the ordering of the models in `modelList`. This can be less of a problem with the `TwoPhaseFlow` app, but to maintain compatibility and flexibility, the user needs to specify a model dictionary that describes which index corresponds to the flow, the model used for interface tracking, whether there are corrections that need to be made during "restart" post-adapt. 

This PR also introduces the AdaptManager object that helps store this model dictionary but can be used in the future to allow for a more flexible approach to passing inputs to the MeshAdaptPUMI object. 

This PR also introduces the `ibm` size-field configuration which is limited to the rans2p+ibm case to have solid-interface driven adaptation. 

# Example with ibm+error driven adaptation

Initial meshes (adapt vs no-adapt):
adapt_vs_noAdapt_initialMesh

Final mesh and solution (adapt vs no-adapt):
adapt_vs_noAdapt_mesh
adapt_vs_noAdapt_velocity

Coarse Mesh results
coarseMesh_and_velocity

The new setup now involves the following syntax:
```
    from proteus.MeshAdaptPUMI import MeshAdapt
    domain = Domain.PUMIDomain(dim=nd,manager=MeshAdapt.AdaptManager()) #initialize the domain
    he = 0.06
    domain.AdaptManager.PUMIAdapter.loadModelAndMesh(b"Reconstructed.dmg", b"Reconstructed.smb")
    modelDict = {'flow':0,'phase':2,'corrections':[3,4]}
    domain.AdaptManager.modelDict = modelDict
    domain.AdaptManager.sizeInputs = [b'pseudo']
    domain.AdaptManager.adapt = opts.adapt
    domain.AdaptManager.hmax = he
    domain.AdaptManager.hmin= he/4.0
    domain.AdaptManager.hphi= he/2.0
    domain.AdaptManager.numIterations= 5
    domain.AdaptManager.PUMIAdapter.setAdaptProperties(domain.AdaptManager)
```

The manager is an object that is part of the PUMIDomain class and stores the inputs to the adapter (object which carries adaptation functions). Having this separation allows for inputs to be more naturally defined and used from the python-side and then passed over to the C++ side. 

Previously, the user specified inputs to the C++ side which then had to be extracted back out to the Python-side. 

The `AdaptManager.sizeInputs` dictionary specifies how/which size fields are computed. Each of those sizefields are then intersected/combined just prior to adaptation. This flexibility allows for different combinations for isotropic mesh adaptation and makes the adaptive module more extensible.

To Do:
[x] make fixes to sdf in FSI test
[x] fix adaptation trigger for `ibm`
[x] allow error estimator with `ibm`
[x] image of rans2p+ibm case
[x] generalize the configuration of size fields to be additive

New files in test with sizes:
```
16K	cylinder.py
4.0K	cylinder_so.py
0	__init__.py
4.0K	test_cylinder2D_ibm_rans2p.py
4.0K	twp_navier_stokes_n.py
8.0K	twp_navier_stokes_p.py
```


zhang-alvin:
> Nice work. Could you change the names a little bit here:
> 
> ```
> domain.PUMIManager=MeshAdaptPUMI.AdaptManager()
> domain.PUMIManager.PUMIAdapter=MeshAdaptPUMI.MeshAdaptPUMI()
> ```
> 
> It sound a little redundant at first glance and maybe could be more expressive
> 
> domain.PUMIManager=MeshAdaptPUMI.DefaultManager(adapter=MeshAdaptPUMI.DefaultAdapter)
> 
> Or the two line version you had but making it clearer that t;he first module is the Adaption Manager and the second one is specifically the Mesh Adapter.

This is a good suggestion. I should be able to condense the two-lines into a one liner. In fact, I can probably set a default adapter to avoid the user ever knowing/worrying about it).

zhang-alvin:
There are now 4 different relevant classes/objects to adaptation.

1) `Domain.PUMIDomain` (domain object)
2) `MeshAdaptPUMI.MeshAdapt.AdaptManager` (python object to manage inputs)
3) `MeshAdaptPUMI.MeshAdapt.MeshAdapt` (the interface to the C++ routines)
4) `MeshAdaptPUMI.AdaptHelper` (helper function class)

`PUMIDomain`, by default, will use `MeshAdapt.AdaptManager()` for the adapt manager.
And the adapt manager uses the `MeshAdapt()` adapter by default. 
`AdaptHelper` is used by `NumericalSolution` and carries helper functions associated with prepping for adaptation.

The user should only ever need to interact with 1) and 2).
So 3) and 4) are hidden away. 
4), which is the `AdaptHelper` will not be accessible by the user directly.

Example:

```
    from proteus.MeshAdaptPUMI import MeshAdapt
    domain = Domain.PUMIDomain(dim=nd,manager=MeshAdapt.AdaptManager()) #initialize the domain
    he = 0.06
    domain.AdaptManager.PUMIAdapter.loadModelAndMesh(b"Reconstructed.dmg", b"Reconstructed.smb")
    modelDict = {'flow':0,'phase':2,'corrections':[3,4]}
    domain.AdaptManager.modelDict = modelDict
    domain.AdaptManager.sizeInputs = [b'pseudo']
    domain.AdaptManager.adapt = opts.adapt
    domain.AdaptManager.hmax = he
    domain.AdaptManager.hmin= he/4.0
    domain.AdaptManager.hphi= he/2.0
    domain.AdaptManager.numIterations= 5
    domain.AdaptManager.PUMIAdapter.setAdaptProperties(domain.AdaptManager)
```
Using the following declaration for the domain would yield the same result as described above
```
    domain = Domain.PUMIDomain(dim=nd) #initialize the domain
```

zhang-alvin:
@tridelat , I've updated the OP to include the coarse mesh result

tridelat:
@zhang-alvin so you added a test for this IBM case in the tests folder? (I can only see modified files)

zhang-alvin:
@tridelat Ah I see, sorry misread your comment. Haven't added a test yet. I can add one in pretty easily.

zhang-alvin:
> @zhang-alvin Looks good, thanks for adding the test!
> Can MeshAdapt be integrated into the TwoPhaseFlow workflow or does it have to use _n and _p files for now?

@tridelat I've not dealt with TPF app with MeshAdapt. There's already some code to facilitate the workflow, so it should be largely compatible. My next PR will likely deal with MeshAdapt with TPF.

PR #1107, opened 2020-03-05
Pin dependencies to a successful CI state

davidbrochart:


tridelat:
Thanks @davidbrochart 
Should I merge #1103 in master so that you can rebase this PR? Even if conda tests are not passing on #1103, hashdist tests are passing, meaning that it is supposedly safe.
Otherwise I will close #1103 and we can merge this one only

davidbrochart:
This PR is already rebased on #1103, so no need to merge right now.
Tests still fail on Linux. I pinned as much as I could to the package versions we have on OSX, where tests pass. So I don't know what's going on.
Could it be that we have some C code reading data from uninitialized memory? We could check the erroneous results are reproducible (if not, this is likely due to uninitialized memory).

tridelat:
> This PR is already rebased on #1103, so no need to merge right now.
> Tests still fail on Linux. I pinned as much as I could to the package versions we have on OSX, where tests pass. So I don't know what's going on.
> Could it be that we have some C code reading data from uninitialized memory? We could check the erroneous results are reproducible (if not, this is likely due to uninitialized memory).

@davidbrochart I could reproduce the failure locally on Linux with the unpinned packages.
I think you already know, but OSX "test" passes on Travis CI because only the installation is checked, but actual tests are not activated.
I will look at the tests again on my local Linux machine with this branch.

davidbrochart:
No I didn't know, I should have looked at the CI! So what I did was useless.
Where can I look at a successful Linux CI?

tridelat:
> No I didn't know, I should have looked at the CI! So what I did was useless.
> Where can I look at a successful Linux CI?

@davidbrochart here is one that passed (and the last one that affected tests I think):
https://travis-ci.com/erdc/proteus/jobs/290783196

davidbrochart:
@tridelat the CI passes on Linux with this last commit. It's a lot of pinning, somehow I have the feeling that the reason of the failure was related to python3.8 (it's pinned to 3.7 here).

tridelat:
@davidbrochart thanks! Interesting that python 3.8 might be what caused the issue.
pinning everything can look a bit cumbersome, but I think it could be preferable in our case to guarantee that results are reproducible.
The conda install on OSX doesn't pass however, it looks like it does not like the pinned gmsh version

tridelat:
@davidbrochart I'm going to go ahead and merge #1103, and then we can merge this PR once the OSX build is passing

davidbrochart:
BTW @tridelat why don't we activate the tests on OSX?

tridelat:
@davidbrochart good question, I am not sure actually, and I am not a OSX user. It might have been that some of the tests were giving problems there, but it's probably worth checking if they now pass.
OSX build on Travis CI was introduced in #1083, where it just tests that a single simulation case can run, and #1068 added the commented out pytest line.

davidbrochart:
Trying out in #1110

PR #1113, opened 2020-03-10
Fix MoveMesh for 3D cases + add 3D FSI test

tridelat:
This PR fixes a code-breaking bug in MoveMesh (missing import statement for 3D case). 
A new 3D FSI test with a floating cube has also been added to make sure that this is covered for the future


tridelat:
> This looks okay. Although again, it'd be good to know the size of the added mesh files.
> 
> And to clarify, the reason this test is needed is because the existing FSI cylinder test case is only 2D whereas the observed bug was in the 3D code?

@zhang-alvin that's right, the bug only appeared in 3D. This test ensures that the whole workflow (two-phase, FSI, ALE) is working in 3D as well.
Concerning the mesh, it is very coarse and the files are relatively small in kb

PR #1114, opened 2020-03-12
Minor cleanup changes.

ejtovar:
In this PR, I did the following:

* Removed the directory `proteus/SWFlows` as seen in the issue: https://github.com/erdc/proteus/issues/1108

* Removed the directory `proteus/SWFlow/tests` since the tests should go  in `proteus/tests/SWFlow`. 

* Cleaned up some of the original tests by adding better descriptions and re-writing some functions in a cleaner way. 

* Added a new test involving a solitary wave propagating towards a conical island. 

* Minor changes to other doc files.

Things to do next: 

1. Verify that new entropy viscosity rescaling is robust and does not cause any issues for established tests. 
2. Update the numerical method to include the proper limiting (requires some thinking on my end)
3. Change point gauges for SWFlow to output free surface elevation.
4. Update the `FCTSTep`  function in `SW2DCV.h` and `GN_SW2DCV.h` files to be able to handle multiple limiting iterations


setup.py
```diff
@@ -720,7 +720,7 @@ def __str__(self):
 
 def setup_given_extensions(extensions):
     setup(name='proteus',
-          version='1.7.3.dev0',
+          version='1.7.3',
```
tridelat: We should keep the `.dev0` here, only tagged release drop it

PR #1118, opened 2020-03-24
mass conservation with porosity fix along with closed sloshing test

jhcollins:
# Mandatory Checklist


cekees:
This may be a tricky to debug. I'll try to run it locally to see if maybe there's just a missing file or something: https://www.google.com/maps/@31.5654087,-90.2885475,1302m/data=!3m1!1e3

The vortex test errors are probably just the result of the slight change in floating point operations due to correct inclusion of porosity--they have a very tight tolerance in the test.

PR #1119, opened 2020-03-26
Make velocity post-processing into BDM1 the default for Two-Phase App

cekees:
# Mandatory Checklist


cekees:
@zhang-alvin I don't see any reason why this won't also work for h- adaptivity, but I haven't tested it. The post-processing happens after RANS2P runs and simply over writes the ('velocity',0) arrays in q and ebqe. This option used to be the default in most of the problems I've done. I believe we started to disable it because I hadn't updated it to include the porosity scaling, but @jhcollins and I did that. It had already been updated to work with mesh motion. The only test it broke with the comparison of the cube location in a still tank. @jhcollins and I have found that several aspects of simulations are significantly better, especially on coarse grids including numerical diffusion of water mass through the open top and non-physical variation of the VOF around  0 or 1.

zhang-alvin:
@cekees I think the post-processing didn't work with adaptivity previously because post-processing relied on nodal-based partitions of the mesh. I assume that's still true?

cekees:
That's right @zhang-alvin. It's still true that it requires nodal partitioning.  The algorithm naturally works on the support of each nodal DOF, so it's convenient that each "owned" node has it's support on the owning processor. We do assign element ownership in the partitioning, so there is an underlying, non-overlapping element partitioning. I wonder if we could pass that mesh to MeshAdapt. I guess the tricky thing is re-initializing the adapted mesh. The most straight-forward way to do it would be to run the same nodal partitioning scheme on the adapted mesh.

cekees:
@jhcollins is doing some more checks on conservation properties, and I'm running some ALE examples on onyx before we merge this.

cekees:
@tridelat the issue I found in parallel is unrelated to this PR. I'm going to merge this and do a new PR for the parallel solver issue.

PR #1121, opened 2020-03-30
remove hard-wired 'r-true' for petsc and reorder init

cekees:
# Mandatory Checklist

# Description

For some reason, PETSc started removing our custom (true) norm at
or around 3.11.x, and we didn't catch the change. I also noticed
that in the current implementation 'r-true' was hardwired, so
I removed that line. I believe we set it to 'r-true' in most cases.


zhang-alvin:
@cekees, if this is a bug fix, is there an existing test that can be tweaked to cover it?

cekees:
No, but it shouldn't be hard to write one that runs quickly just to test that 'r' and 'r-true' modify the PETSc solver accordingly without actually needing to run anything.  Once I finish testing on the HPC I'll add it.

PR #1123, opened 2020-04-04
fixed pandas install that was breaking docker build

cekees:
# Mandatory Checklist

# Description

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN rm -rf proteus && \
     cd proteus && \
     git checkout master && \
     git submodule update --init --recursive && \
-    make N=4 develop && \
+    make N=1 develop && \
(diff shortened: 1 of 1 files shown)


zhang-alvin:
@cekees so the dockercloud test is failing because dockercloud as a service has ceased is that correct? 

Also, I noticed that the `erdc/stack_base` docker image wasn't building on Dockerhub because the build was taking too long (> 2hrs). Does "importing" that image with this Dockerfile still work? Or are all images that are imported with `FROM` built locally first based on the corresponding Dockerfile?

cekees:
@zhang-alvin I think the problem is that the stack_base container wasn't updated yet, so it tried to build the whole stack. Docker cloud just redirects to hub.docker.com now.

zhang-alvin:
> @zhang-alvin I think the problem is that the stack_base container wasn't updated yet, so it tried to build the whole stack. Docker cloud just redirects to hub.docker.com now.

I tried the `stack_base` container locally and added the 18.04 build-bot build cache to speed up the build. Many of the large packages like `blas` and `scipy` still build though because the build hashes don't match those in the cache for some reason. I can post the details if you're interested.

PR #1128, opened 2020-04-09
Rans3pf

JohanMabille:
# Mandatory Checklist

# Description


cekees:
Maybe try compiling with PROTEUS_OPT="-w -g0 -O0 -DNDEBUG" ?

JohanMabille:
I pushed to try, but are these flags acceptable in production? I mean, they would lead to a significative performance hit, no?

JohanMabille:
The build is now passing, I need to fix the python file.

cekees:
PROTEUS_OPT="-w -g0 -O0 -DNDEBUG" would just be for travis. My thinking is that the main objective of the travis build is to run the tests quickly. If errors arise then a developer can rebuild and test with PROTEUS_OPT="-w -g -O0 -UNDEBUG" and so on. I posted some comparisons of compile times on the proteus slack channel.

setup.py
```diff
@@ -202,11 +202,13 @@ def __str__(self):
         libraries=['m',PROTEUS_LAPACK_LIB,
                    PROTEUS_BLAS_LIB]
     ),
-    Extension("mprans.cRANS3PF",['proteus/mprans/cRANS3PF.pyx'],
-              depends=['proteus/mprans/RANS3PF.h','proteus/mprans/RANS3PF2D.h', 'proteus/ModelFactory.h', 'proteus/CompKernel.h'],
-              language='c++',
-              extra_compile_args=PROTEUS_OPT+["-std=c++14","-mavx"],
-              include_dirs=get_xtensor_include()),
+    Extension(
+        'mprans.RANS3PF',
```
cekees, line 206: mprans.RANS3PF ==> mprans.cRANS3PF

JohanMabille: I'm gonna make a global review of these names, it looks like I missed some naming inconsistency between the setup.py, the XXX.py files and the XXX.cpp iles.

cekees:
I think if you split the 2D and 3D bindings in to separate files, then it will fix our build problems. I'm just talking about RANS3PF/RANS3PF2D, RANS2P/RANS2P2D, and RANS3PSed/RANS3PSed2D:

https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-reduce-the-build-time

JohanMabille:
I'll try that in a dedicated commit once the tests pass.

proteus/MeshAdaptPUMI/Adapter.cpp
```diff
@@ -0,0 +1,14820 @@
+/* Generated by Cython 0.29.15 */
```
cekees, line 1: @zhang-alvin we should delete this, right? This is just the cython generated file, not a pybind11 interface.

JohanMabille:
Indeed, this was added by error with a greedy `git add proteus/`.

PR #1133, opened 2020-04-14
Splitted RANS3PF module

JohanMabille:
# Mandatory Checklist

# Description

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ proteus/equivalent_polynomials.cpp
 proteus/fenton/Fenton.cpp
 proteus/flcbdfWrappers.cpp
 proteus/mbd/CouplingFSI.cpp
+proteus/MeshAdaptPUMI/Adapter.cpp
 proteus/mprans/BoundaryConditions.cpp
 proteus/mprans/MeshSmoothing.cpp
 proteus/mprans/cAddedMass.cpp
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
       - export PATHSAVE=$PATH
       - export PATH=$PWD/linux/bin:$PATH
       - export LD_LIBRARY_PATH=$PWD/linux/lib:$LD_LIBRARY_PATH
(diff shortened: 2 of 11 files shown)


JohanMabille:
The split of RANS3PF fixes the memory issue but now we the file takes too long to compile, resulting in travis time out ...

cekees:
> The split of RANS3PF fixes the memory issue but now the file takes too long to compile, resulting in travis time out ...

To be clear: the memory issue we are seeing is on Docker Hub, not travis, and it's still there on your branch: https://hub.docker.com/repository/registry-1.docker.io/erdc/proteus/builds/cc7c9588-ab4a-4ff2-a56a-bf391be73638

I'm guessing you can fix the compile time issue by modifying the Travis file to use PROTEUS_OPT='-g0 -O0 -DNEBUG'' as I mentioned on the proteus channel.

JohanMabille:
I think we're talking about two different memory issues, sorry for not being clear. Without the split of RANS3PF module, we had a memory issue when trying to build with -O2 or -O3 on travis (only on hdist). Thus the switch to -O0 in #1128. Now, with the split, we don't have this memory issue anymore, but the build takes too long.

cekees:
I see you split RANS3PF first, so you may see more of an improvement when you split RANS3PSed, which seems to be the most resource-intensive compile. I think we should go ahead and split RANS2P as well for consistency (probably has marginal effect on compile times).

JohanMabille:
I think the `RANS2P` module has already been splitted.

JohanMabille:
@cekees @zhang-alvin @tridelat I think is one is ready to be merge.

.gitignore
```diff
@@ -62,6 +62,7 @@ proteus/equivalent_polynomials.cpp
 proteus/fenton/Fenton.cpp
 proteus/flcbdfWrappers.cpp
 proteus/mbd/CouplingFSI.cpp
+proteus/MeshAdaptPUMI/Adapter.cpp
```
zhang-alvin, line 65: @JohanMabille Thanks for adding this.

JohanMabille:
> So to clarify, the "split" referred to in this PR is separating the 2D RANS3PF module from the 3D RANS3PF module at compile time? Likewise for RANS3PFSed.

Yes.

PR #1134, opened 2020-04-14
Added ArgumentDict and plugged it into AddedMass

JohanMabille:
# Mandatory Checklist

# Description


JohanMabille:
@cekees this is an example of how we could replace the long list of arguments with dictionnaries of `pyarray` and scalar objects. Although this shortens the signature of the native functions, I am not sure that this really simplifies the code and I wonder if gathering the different arrays in structures (that can be splitted and nested) would be a better alternative.

JohanMabille:
This new class allows to keep the numpy arrays synchronized with their C++ counterpart.
Next step is to add operator[] overloads to `arguments_dict` that dispatch to the internal map depending on the type of the argument. This way we can replace `argDict.darray['mesh_trial_ref'] = self.u[0].femSpace.elementMaps.psi` with `argDict['mesh_trial_ref'] = self.u[0].femSpace.elementMaps.psi` and the user does not have to wonder about the argument type on the Python side.

JohanMabille:
@cekees I'm quite happy with this API. On the Python side, you can now fill the dictionnary without wondering about the type of the arguments. I also added a '__repr__' function to ArgumentsDict so that you can easily print it. If that looks good to you, I will start to use it everywhere it is required in the mprans module once this is merged.

cekees:
I like this! Could you maybe test just on RANS3PSed to see if it reduces compilation time and memory requirements for the compilation step?

JohanMabille:
I am working on it, more precisely, I worked on a script to automate the migration to the dictionary of arguments. When I started to use it on RANS3PSed, I noticed that the number of arguments expected by [calculateResidual](https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS3PSed.h#L39) (166) does not match that passed by the [python call](https://github.com/erdc/proteus/blob/master/proteus/mprans/RANS3PSed.py#L1683) (178).

cekees:
@zhang-alvin take a look at this approach. I think in the long run we may want to create the new Python/C++ dictionary object at the level of NumericalSolution using some hierarchical naming convention (e.g. model(rans2p,..)_set(q,ebqe,...)_field, but I think it's probably safest to do an incremental implementation where @JohanMabille converts the function signatures first (i.e. one local dict per call to C++) because he can nearly automate that conversion, then the active proteus team works together on a larger refactorization.

JohanMabille:
That looks like a bug in my script actually.

JohanMabille:
I think the module that requires too much memory is RANS3PF, let's make it use ArgumentsDict too!

JohanMabille:
@cekees I think this is ready to go! I could enable the O2 flag again. Once this is merged, I will start refactoring the remaining of mprans, with small PRs so avoid conflicts and time-consuming merges.

Issue #1139, opened 2020-04-17
init_RANS functions


cekees:
@JohanMabille I was doing some local testing of the RANS3PSed class (just merged your PR #1133) and noticed some 2D calls in the 3D pybind11 wrappers:
```bash 
$ grep init_RANS *.cpp
RANS3PF2D.cpp:void init_RANS3PF2D(py::module& m)
RANS3PF.cpp:void init_RANS3PF2D(py::module& m);
RANS3PF.cpp:    init_RANS3PF2D(m);
RANS3PSed2D.cpp:void init_RANS3PFSed2D(py::module& m)
RANS3PSed.cpp:void init_RANS3PFSed2D(py::module& m);
RANS3PSed.cpp:    init_RANS3PFSed2D(m);
```

I think it probably works fine anyway (just a naming issue), but can you explain what is going on here? I noticed that the RANS2P wrappers don't declare or call a module init function.

JohanMabille:
RANS2P and RANS2P2D are actually two separate Python extensions. RANS3PF and RANS3PF2D sill belong to the same Python extension, I just split the bindings in two separate cpp files. This reduces the amount of memory required for building each of these cpp files while not requiring any change on the Python side.

The cpp file defining the bindings still has to bind both classes, thus the `init_XXX` function.

PR #1140, opened 2020-04-20
Decouple compilation of 2D and 3D mprans classes

cekees:
I merged master back into a testing branch I made recently. This is in response to the discussion #1139 .  I don't find that this compiles significantly faster, but the design is simpler in my opinion. In this version the 2D and 3D extension modules don't depend on one another, so when you're developing on say, RANS3P2D, it doesn't force recompilation of RANS3P.

# Mandatory Checklist

# Description


Dockerfile
```diff
@@ -14,37 +14,13 @@ ENV F90 mpif90
 RUN rm -rf proteus && \
     git clone https://github.com/erdc/proteus && \
     cd proteus && \
-    git checkout master && \
+    git checkout docker_no_opt && \
```
zhang-alvin, line 17: I know dockerhub might still not compile, but dockerhub should be tracking when `docker_no_opt` right?

Dockerfile
```diff
@@ -14,37 +14,13 @@ ENV F90 mpif90
...
-RUN chmod a+rx /usr/local/bin/*
-
-RUN ipython kernel install
-
-USER $NB_USER
-
-# Import matplotlib the first time to build the font cache.
-#ENV XDG_CACHE_HOME /home/$NB_USER/.cache/
-#RUN MPLBACKEND=Agg python -c "import matplotlib.pyplot"
+ENV LD_LIBRARY_PATH /home/$NB_USER/proteus/linux/lib:/home/$NB_USER/proteus/linux/lib64:$LD_LIBRARY_PATH
```
zhang-alvin, line 26: I took another look at the singularity container and setting the LD_LIBRARY_PATH actually causes an issue with inheriting the LD_LIBRARY_PATH of the system. Not really an issue for this PR, but something to keep in mind if you happen to know another way of possibly doing it.

cekees:
@JohanMabille can you take a look at this PR. I'm thinking I'd like to go ahead and split the RANS* classes to have separate modules following the same pattern. Mainly just to prevent confusion in the future. The compile times appeared to be slightly faster on my laptop, but the difference was marginal. This is mainly just for consistency/simplicity.

cekees:
Also @JohanMabille can you explain why #define FORCE_IMPORT_ARRAY is necessary?

proteus/mprans/RANS3PF2D.h
```diff
@@ -10,6 +10,9 @@
 #include "ModelFactory.h"
 #include "SedClosure.h"
 #include "equivalent_polynomials.h"
+const  double DM=0.0;//1-mesh conservation and divergence, 0 - weak div(v) only
+const  double DM2=0.0;//1-point-wise mesh volume strong-residual, 0 - div(v) only
+const  double DM3=1.0;//1-point-wise divergence, 0-point-wise rate of volume change
```
JohanMabille, line 15: I think these could be static to avoid multiple symbol definition if this header is included in many cpp in a same library.

proteus/mprans/RANS3PSed2D.h
```diff
@@ -9,13 +9,12 @@
 
 namespace py = pybind11;
 
+const  double DM=0.0;//1-mesh conservation and divergence, 0 - weak div(v) only
+const  double DM2=0.0;//1-point-wise mesh volume strong-residual, 0 - div(v) only
+const  double DM3=1.0;//1-point-wise divergence, 0-point-wise rate of volume change
```
JohanMabille, line 14: Same here

proteus/mprans/cRANS3PSed.cpp
```diff
@@ -0,0 +1,31 @@
...
+PYBIND11_MODULE(cRANS3PSed, m)
+{
+    xt::import_numpy();
+
+    py::class_<cppRANS3PSed_base>(m, "cppRANS3PSed_base")
+        .def(py::init(&proteus::newRANS3PSed))
+        .def("calculateResidual", &cppRANS3PSed_base::calculateResidual)
+        .def("calculateJacobian", &cppRANS3PSed_base::calculateJacobian)
+        .def("calculateVelocityAverage", &cppRANS3PSed_base::calculateVelocityAverage);
+
```
JohanMabille, line 30: This file looks like a duplicata of `RANS3PSed.cpp` or am I missing something? If this is a duplicate, this defines the module twice. Also I would be in favor of keeping the name of the implementation file consistent with that of the header file. The 'c' prefix makes me think of a file generated by cython while it is not.

proteus/mprans/cRANS3PSed2D.cpp
```diff
@@ -0,0 +1,30 @@
...
+PYBIND11_MODULE(cRANS3PSed2D, m)
+{
+    xt::import_numpy();
+
+    py::class_<cppRANS3PSed2D_base>(m, "cppRANS3PSed2D_base")
+        .def(py::init(&proteus::newRANS3PSed2D))
+        .def("calculateResidual", &cppRANS3PSed2D_base::calculateResidual)
+        .def("calculateJacobian", &cppRANS3PSed2D_base::calculateJacobian)
+        .def("calculateVelocityAverage", &cppRANS3PSed2D_base::calculateVelocityAverage);
+}
```
JohanMabille, line 30: Same comment w.r.t. `RANS3PSed2D.cpp`

setup.py
```diff
@@ -171,8 +171,15 @@ def __str__(self):
...
         'mprans.cRANS3PSed',
-        ['proteus/mprans/RANS3PSed.cpp', 'proteus/mprans/RANS3PSed2D.cpp'],
-        depends=['proteus/mprans/RANS3PSed.h','proteus/mprans/RANS3PSed2D.h', 'proteus/ModelFactory.h', 'proteus/CompKernel.h'],
+        ['proteus/mprans/cRANS3PSed.cpp'],
+        depends=['proteus/mprans/RANS3PSed.h','proteus/ModelFactory.h', 'proteus/CompKernel.h'],
+        include_dirs=get_xtensor_include(),
+        extra_compile_args=PROTEUS_OPT+['-std=c++14'],
+        language='c++'),
+    Extension(
+        'mprans.cRANS3PSed2D',
```
JohanMabille, line 180: Ah you replaced `RANS3PSed.cpp` with `cRANS3PSed.cpp` here (same for the 2D file) so the modules are not defined twice. Therefore, the `RANS3PSed.cpp` and `RANS3PSed2D.cpp` are not used anymore and could be deleted. However I would be in favor of doing the contrary to keep naming consistency with other modules.

JohanMabille:
Regarding the `FORCE_IMPORT_ARRAY` token, first you should be aware of [this](https://docs.scipy.org/doc/numpy/reference/c-api.array.html?highlight=import_array#miscellaneous).

To resume, the `import_array` function must called only once in the initialization routine of the extension module. Besides, each compilation unit that does not contain the initialization routine must define `NO_IMPORT_ARRAY` before including numpy headers, otherwise you end up with link errors.

Now, when you include a header from `xtensore-python`, you indirectly include a header form numpy. Requiring the user to define `NO_IMPORT_ARRAY` in every cpp module is a bit cumbersome, so we did the opposite: By default, `NO_IMPORT_ARRAY` is defined except if you define `FORCE_IMPORT_ARRAY` before including a header from `xtensor_python`.

This way, extension authors only have to be careful when they write the module that contains the initialization routine. Besides, when `FORCE_IMPORT_ARRAY` is not defined, the `import_numpy` function does nothing, preventing accidental calls to it in other cpp files.

proteus/mprans/RANS3PF2D.h
cekees: These headers really shouldn't be used in more than one compilation unit, and the point here is to make the symbols local constants. They're just switches that are specific to this file (could also be macros).

JohanMabille:
LGTM!

PR #1152, opened 2020-04-28
Pllugged ArgumentsDict into SW2DCV

JohanMabille:
# Mandatory Checklist

# Description

### DIFF
--- a/proteus/mprans/SW2DCV.h
+++ b/proteus/mprans/SW2DCV.h
@@ -6,6 +6,7 @@
 #include <cmath>
 #include <iostream>
 #include <valarray>
+#include "ArgumentsDict.h"
 #include "xtensor-python/pyarray.hpp"
 
 namespace py = pybind11;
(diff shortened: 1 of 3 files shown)


ejtovar:
I think we can merge this once the SWEs tests pass. It looks like they're still failing.

JohanMabille:
The failure is due to arguments type mismatch: a double is expected on the C++ side while an int is passed from the Python. Before using the dictionaries, the conversion was automatic and everything worked fine. Now the parameter ends up in the wrong scalar dictionary. When asking for this scalar in C++, since it is not found in the map, a default value is returned.

We need to catch these bugs quickly, I will improve the dictionaries for scalar in a dedicated PR and rebase all the failing PRs on it once it is merged (the other failures are probably due to the same issue).

JohanMabille:
See #1157

JohanMabille:
@ejtovar I think this one is ready for a merge.

ejtovar:
Thanks @JohanMabille!

PR #1153, opened 2020-04-28
Plugged ArgumentsDict into SW2D

JohanMabille:
# Mandatory Checklist

# Description

### DIFF
--- a/proteus/mprans/SW2D.h
+++ b/proteus/mprans/SW2D.h
@@ -4,6 +4,7 @@
 #include <iostream>
 #include "CompKernel.h"
 #include "ModelFactory.h"
+#include "ArgumentsDict.h"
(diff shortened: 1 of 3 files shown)


ejtovar:
@cekees The `SW2D.h` and `SW2D.py` are not being used anymore (at least not by me). Should we just remove these all together? It will save @JohanMabille time in the future when he's refactoring all these model files.

cekees:
> @cekees The `SW2D.h` and `SW2D.py` are not being used anymore (at least not by me). All the development has been done on `SW2DCV` and `GN_SW2DCV` and tests use those models anyway. Should we just remove these all together? It will save @JohanMabille time in the future when he's refactoring all these model files.

I think they're still worth keeping around. I'd like to do a comparison of explicit EV vs  implicit SUPG at some point.

PR #1156, opened 2020-04-28
Plugged ArgumentsDict into RANS2P

JohanMabille:
# Mandatory Checklist

# Description

### DIFF
--- a/proteus/LinearSolvers.py
+++ b/proteus/LinearSolvers.py
@@ -24,6 +24,7 @@ from petsc4py import PETSc as p4pyPETSc
 from math import *
 import math
 from .Profiling import logEvent
+from .mprans import cArgumentsDict
 
 class LinearSolver(object):
     """ The base class for linear solvers.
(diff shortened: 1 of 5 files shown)


JohanMabille:
The methods "getTwoPhasesXXX" are never called in the Python file.

cekees:
> The methods "getTwoPhasesXXX" are never called in the Python file.

They're called in the solvers:
https://github.com/erdc/proteus/blob/master/proteus/LinearSolvers.py#L3482

Once you finish this, I may look at creating a separate class for those functions, but for now I'd like to leave them there.

JohanMabille:
Thanks for the pointer, I need to fix the linear solvers then.

JohanMabille:
Theoretically it should, since it is where the python interface of `arguments_dict` is defined. I can make a pass on setup.py to clean it once we're done with the migration.

proteus/LinearSolvers.py
```diff
@@ -3479,76 +3480,94 @@ def updateTPAdvectionOperator(self,
...
+
+        argsDict = cArgumentsDict.ArgumentsDict()
+        argsDict["mesh_trial_ref"] = self.model.u[0].femSpace.elementMaps.psi
+        argsDict["mesh_grad_trial_ref"] = self.model.u[0].femSpace.elementMaps.grad_psi
+        argsDict["mesh_dof"] = self.model.mesh.nodeArray
+        argsDict["mesh_l2g"] = self.model.mesh.elementNodesArray
+        argsDict["dV_ref"] = self.model.elementQuadratureWeights[('u',0)]
+        argsDict["p_trial_ref"] = self.model.u[0].femSpace.psi
+        argsDict["p_grad_trial_ref"] = self.model.u[0].femSpace.grad_psi
+        argsDict["vel_trail_ref"] = self.model.u[1].femSpace.psi
```
cekees, line 3492: trail -> trial (spelling)

proteus/mprans/RANS2P.py
```diff
@@ -1460,216 +1461,226 @@ def getResidual(self, u, r):
...
-                                      self.coefficients.particle_alpha,
-                                      self.coefficients.particle_beta,
-                                      self.coefficients.particle_penalty_constant,
-                                      self.coefficients.phi_s,
-                                      self.coefficients.phisField,
-                                      self.coefficients.use_pseudo_penalty,
-                                      self.coefficients.useExact)
+
+        argsDict = cArgumentsDict.ArgumentsDict()
+        argsDict["NONCONSERVATIVE_FORM"] = self.coefficients.NONCONSERVATIVE_FORM
```
cekees, line 1466: NONCONSERVATIVE_FORM and MOMENTUM_SGE, etc. are real valued flags, but one of the tests sets them to integers. Maybe just wrap them in `float()~ so reasonable dynamic casts work (e.g. bool or int promotion to float will be fine in all cases).

cekees: OK, looks like also `argsDict["shockCapturingDiffusion"] = int(self.shockCapturing.shockCapturingFactor)` should use float instead of int. That's another postive real number.

cekees:
Made these changes and fixed other issues in #1163

JohanMabille:
Thanks! I will go on with the remaining files tonight.

PR #1178, opened 2020-05-11
Pytest changes for macOS.

ejtovar:
# Mandatory Checklist

- [x ] Title of pull request describes the changes/features
- [x  ] Request at least 2 reviewers
- [x  ] If new files are being added, the files are no larger than 100kB. Post the file sizes.

# Description
1. Changed the pytest command in `.travis.yml` file to allow the tests to runon macOS. 
2. Added a `test-macos` command in the makefile that runs the basic tests suite. 
3. In a few tests, I added the command `@pytest.mark.skipif(sys.platform == "darwin", reason="does not run on macOS")` to skip particular tests when using macOS.
4. Fixed a typo in `test_CLSVOF_discICs.py`. Changed class name ` TestSWEs` to `TestCLSVOF`


Makefile
```diff
@@ -316,6 +316,19 @@ test-conda: air-water-vv check
 	@echo "Running air-water-vv test set 2"
 	-source ${PROTEUS_PREFIX}/bin/proteus_env.sh; MPLBACKEND=Agg py.test -n ${N} --dist=loadfile --forked -v air-water-vv/Tests/2nd_set -m ${TEST_MARKER}
 
+test-macos: air-water-vv check
```
zhang-alvin, line 319: while this is functional, is there a way to instead place this under the usual `test` target? For example, is there a way to automatically determine that the os is mac and have it run these commands when `make test` is called?

proteus/mprans/SW2DCV.py
```diff
@@ -356,7 +356,7 @@ def evaluate(self, t, c):
     def preStep(self, t, firstStep=False):
         if firstStep:
             # Init boundaryIndex
-            assert (self.model.boundaryIndex is None and self.model.normalx is not None, "Check boundaryIndex, normalx and normaly")
+            assert self.model.boundaryIndex is None and self.model.normalx is not None, "Check boundaryIndex, normalx and normaly"
```
zhang-alvin, line 359: I was under the impression that python3 syntax required parenthesis around `asserts` and `prints`.

proteus/tests/ci/test_poisson.py
```diff
@@ -103,7 +103,7 @@ def test_c0p2(self):
         del ns
         assert(True)
 
-    def check_c0q1(self,test_hexMesh_3x3=False,use_petsc=False,name="_hexMesh_"):
+    def check_c0q1(self,test_hexMesh_3x3=False,use_petsc=True,name="_hexMesh_"):
```
zhang-alvin, line 106: why doesn't this change/break the existing test for non osx builds?

ejtovar: I might have accidentally changed this.  However I'm not sure why it didn't change the outcome of the tests...

proteus/mprans/SW2DCV.py
ejtovar: Hm, maybe. It kept giving me a warning, so I just took off the parenthesis. Let me double check.

Makefile
ejtovar: I suppose there is. I only added it because the pytest commands are different/and it was easier. But we can talk about this more via slack.

cekees: You could use a conditional in the Makefile: https://www.gnu.org/software/make/manual/html_node/Conditional-Example.html#Conditional-Example

proteus/mprans/SW2DCV.py
ejtovar: It seems like you only need parentheses for a multi-line assert.

cekees:
Thanks @ejtovar . I think this is close enough that you can just relax the 4 remaining tests so that they pass (I guess they are already passing on your macbook?). We're just asking for too much precision on these, I think: ```
FAILED proteus/tests/test_analyticalSolutions.py::test_diffusionSin1D
FAILED proteus/tests/test_analyticalSolutions.py::test_diffusionSin2D
FAILED proteus/tests/test_analyticalSolutions.py::test_STflowSphere_P
FAILED proteus/tests/cylinder2D/ibm_rans2p_3D/test_cylinder3D_ibm_rans2p.py::Test_ibm::test_ex2```

ejtovar:
@cekees No they're not passing on my macbook. They did seem like a "precision" type failures which is why I didn't skip them.

proteus/tests/ci/test_poisson.py
cekees: I think this is not being run currently.

PR #1181, opened 2020-05-13
Re-enable solver tests and update petsc to 3.13.1

cekees:
# Mandatory Checklist

# Description


zhang-alvin:
@cekees, there seems to be a file conflict with `test_quads.py`?

proteus/mprans/SW2DCV.py
```diff
@@ -356,7 +356,7 @@ def evaluate(self, t, c):
     def preStep(self, t, firstStep=False):
         if firstStep:
             # Init boundaryIndex
-            assert (self.model.boundaryIndex is None and self.model.normalx is not None, "Check boundaryIndex, normalx and normaly")
+            assert self.model.boundaryIndex is None and self.model.normalx is not None, "Check boundaryIndex, normalx and normaly"
```
zhang-alvin: didn't @ejtovar observe the opposite behavior being necessary for python3?

proteus/tests/periodic/test_periodic.py
```diff
@@ -90,7 +90,7 @@ def test_periodic_2D(load_periodic_opts_2D,
     expected_path = 'comparison_files/' + 'comparison_' + 'ductq1t12dpghe0.0975609756097561' + '_velocity_t25.csv'
     #write comparison file
     #np.array(actual.root.velocity_t25).tofile(os.path.join(script_dir, expected_path),sep=",")
-    np.testing.assert_almost_equal(np.fromfile(os.path.join(script_dir, expected_path),sep=","),np.array(actual.root.velocity_t25).flatten(),decimal=8)
+    np.testing.assert_almost_equal(np.fromfile(os.path.join(script_dir, expected_path),sep=","),np.array(actual.root.velocity_t25).flatten(),decimal=7)
```
zhang-alvin, line 93: Is it just petsc's internal methods that are driving the slight changes in the results?

cekees:
I see, that came from the os x PR merge a few minutes ago. I fixed it. I imagine we may have some os x test failures, but I'm just going to disable/skip any that arise for now.

proteus/tests/solver_tests/test_nseDC_THQuad_2D.py
```diff
@@ -63,13 +55,12 @@ def _runTest(self):
...
         # be updated accordingly.
         relpath = 'comparison_files/drivenCavityNSE_LSC_expected.h5'
-        expected = tables.open_file(os.path.join(self._scriptdir,relpath))
         actual = tables.open_file('drivenCavityNSETrial.h5','r')
-
-        assert numpy.allclose(expected.root.velocaity_t7.read(),
-                              actual.root.velocity_t7.read(),
-                              atol=1e-2) 
-        expected.close()
+        np.savetxt(os.path.join(self._scriptdir,relpath),actual.root.velocity_t7.read(),delimiter=',')
```
zhang-alvin, line 59: What's the motivation for writing the text file here? It doesn't look like the current test really requires it as the data is in memory already.

proteus/mprans/SW2DCV.py
cekees: I thought he saw the same. The conda build  of python seems to do a better job of explaining why, but basically assert is a keywork, not a function, and no-empty tuples always evaluate to True, so this is a nefarious bug. If you want to do multi-line use \ or be careful to enclose only the test in (): 

https://stackoverflow.com/questions/3112171/python-assert-with-and-without-parenthesis

proteus/tests/solver_tests/test_nseDC_THQuad_2D.py
cekees: I meant to fix that. Thanks for catching it. Will push one more commit shortly.

proteus/mprans/SW2DCV.py
zhang-alvin: Okay I see. I've seen a lot of `raise(NameError, 'string')` statements in setup files so it looked like the parentheses were supposed to be there.

cekees:
> Looks good. Again, just a reminder on file size.

Roger on the file size. I had been aiming at < 200kb before I released our target was actually 100kb. Anyway, I mostly got rid of comparison files here and switched to csv. I merged all these solver_tests into one place for a future effort do a better job on test criteria and consistency of options, use of fixtures, etc.

Issue #1179, opened 2020-05-12
Updating arrays after MoveMesh step


zhang-alvin:
There are a number of data structures that are not properly updated after a MoveMesh step. 

`postStep()` (https://github.com/erdc/proteus/blob/master/proteus/mprans/MoveMesh.py#L123) updates the `mesh.nodeArray` which stores the coordinates of the vertices. 

There is also a separate `updateAfterMeshMotion` function (https://github.com/erdc/proteus/blob/master/proteus/mprans/MoveMesh.py#L801) that is called before the next MoveMesh step, which seems to update the element and boundary quadrature fields. 

Neither update the `nodeDiametersArray` or `elementDiametersArray` used in the computation of the Heaviside function. 

Neither update the coefficient weighting in the point gauges, which would lead to inaccurate point gauge readings when the mesh is moved. Similar issues should arise when identifying segments for the line gauges. 

Updating the node diameters is simple with a call to `cmeshTools.computeGeometricInfo`. Updating the gauges might wind up to be quite expensive (e.g. what if a point gauge now lies on a different element because of the mesh motion). 

I'm putting this issue up to get some ideas, maybe have a discussion. I think that all these updates should occur during postStep of the MoveMesh model with the ability to influence the data structures/arrays in other models if necessary. There may also be a way to smartly update the structures used to obtain the local mesh entities of interest (e.g. k-d tree).

zhang-alvin:
#1182 will effectively be reverted in #1211 as there were errors popping up in parallel MoveMesh runs. 

There may be some additional calls needed to get things smoothed out in parallel such as:
```
        par_nodeDiametersArray = ParVec_petsc4py(self.subdomainMesh.nodeDiametersArray,
                                                 bs=1,
                                                 n=self.subdomainMesh.nNodes_owned,
                                                 N=self.nNodes_global,
                                                 nghosts=self.subdomainMesh.nNodes_global - self.subdomainMesh.nNodes_owned,
                                                 subdomain2global=self.nodeNumbering_subdomain2global)                                          
        par_nodeDiametersArray.scatter_forward_insert()
```
which is seen after partitioning

PR #1182, opened 2020-05-14
Node/element diameters need to be updated with MoveMesh

zhang-alvin:
# Mandatory Checklist

# Description

Referring to one of a few problems described in #1179. This PR is simply to get correctness in place. Later PRs will include tests based on a pseudo-adapted mesh. 

Update for simplex meshes only in postStep()


zhang-alvin:
> Looks good. Not suggesting this idea for this PR, but I think it would be nice to sit down one day and review/discuss our mesh API in light of our current workflows for a) initial mesh generation b) mesh motion (r-adaptivity), and c) h-adaptivity. Seems like we could probably make these stages more general to allow for some variations in the future, like quad/hex meshes, mixed cell types, etc.

Yes. In this scenario, maybe we can have an update mesh function that's an abstract function of the `Mesh` object.

PR #1193, opened 2020-05-20
fix some problems with rans3p with new pybind11/cArgumentsDict

cekees:
This fixes some issues:
 - missing import of cArgumentsDict in RANS3PSed
 - missing definition of keyword arguments in SedimentClosure.cpp
 - several instances where 'arg' was used instead of 'argsDict'

@acatwithacomputer you may also need to update some of the input files for the sediment tests to 1) ensure that floating point numbers are used for floating point parameters,  otherwise you'll get an error that the parameter is not available (e.g. use rho_s = 1.0, not 1, etc.) 2) We got rid of the RDLS3P class because it was redundant, so you may need get errors trying to import RDLS3P--just change it to RDLS.

/cc @JohanMabille 

# Mandatory Checklist

# Description


JohanMabille:
Arf sorry for missing it during the migration.

PR #1197, opened 2020-05-26
SWFlow update for solvers and tests

ejtovar:
# Mandatory Checklist

- [x ] Title of pull request describes the changes/features
- [x ] Request at least 2 reviewers
- [x  ] Contributor has read and agreed with CONTRIBUTING.md and has added themselves to CONTRIBUTORS.md 

# Description
This is a first part update regarding the SWFlow app. In this PR, I have updated several things regarding the SWFlow solvers and tests. Specifically:

1. I modified the solvers for SWEs and Green-Naghdi to incluce multiple FCT/convex limiting iterations
2. Changed the way the entropy viscosity coefficients are defined
3. Cleaned up the solvers by removing unused functions, etc
4. Modified the python model files and the NonlinearSolvers.py file to include fixes for parallel runs 
5. Modified the SWFlow tests to include the option of having unstructured meshes and some minor clean up

The planned future changes are as follows:

## SWFlow update 2

1. Fix parallel issues. Current state of this issue is that the parallel runs are not perfect, ie the solution for np > 1 is not "exactly" the same for np = 1. The deviations at the moment are very tiny, but still there.

2.  I will clean up the gauge capabilities to include the free surface values and velocity (not the momentum) values depending on what the user chooses. Currently, this can only be done in a "hacky" way.  

## SWFlow update 3

1. Add the capability to use wave tools for SWFlows


proteus/NonlinearSolvers.py
```diff
@@ -472,7 +472,7 @@ def setLinearSolverTolerance(self,r):
         self.norm_r_last = self.norm_r
         self.linearSolver.setResTol(rtol=eta,atol=self.linearSolver.atol_r)
     def solve(self,u,r=None,b=None,par_u=None,par_r=None,linear=False):
-        r""" Solves the non-linear system :math:`F(u) = b`.
+        """ Solves the non-linear system :math:`F(u) = b`.
```
zhang-alvin, line 475: was this just a typo or something related to formatting?

proteus/NonlinearSolvers.py
```diff
@@ -755,6 +755,8 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
         self.F.secondCallCalculateResidual = 0
         self.computeResidual(u,r,b)
         u[:] = r
+        if par_u is not None:
```
zhang-alvin, line 758: was this the magic call that got parallelization to work?

proteus/mprans/GN_SW2DCV.h
```diff
@@ -17,6 +17,7 @@ namespace py = pybind11;
 #define REESTIMATE_MAX_EDGE_BASED_CFL 0
 #define LAMBDA_MGN 1
 #define IF_BOTH_GAMMA_BRANCHES 0
+#define LIMITING_ITERATION 2
```
zhang-alvin, line 20: out of curiosity what's the significance of stopping at the second iteration?

proteus/mprans/GN_SW2DCV.h
```diff
@@ -909,45 +372,32 @@ class GN_SW2DCV : public GN_SW2DCV_base {
       bar_deltaSqd_kin[i] =
           bar_deltaSqd_kin[i] /
           (csrRowIndeces_DofLoops[i + 1] - csrRowIndeces_DofLoops[i]) / 2.0;
-    }
+    } // i loops ends here
```
zhang-alvin, line 375: I like the comment at the end of the loop, it's good practice.

proteus/mprans/GN_SW2DCV.h
```diff
@@ -1842,24 +1420,31 @@ class GN_SW2DCV : public GN_SW2DCV_base {
...
-        double new_rescale = fmax(fabs(entropy_flux) + fabs(sum_entprime_flux), 1E-300);
+        sum_entprime_flux =
+            -(ith_flux_term1 * eta_prime1 + ith_flux_term2 * eta_prime2 +
+              ith_flux_term3 * eta_prime3);
+        double new_rescale =
+            fmax(fabs(entropy_flux) + fabs(sum_entprime_flux), 1E-30);
 
         // COMPUTE ENTROPY RESIDUAL //
-        double one_over_entNormFactori = 1.0 / rescale;
+        double one_over_entNormFactori = 1.0 / new_rescale;
```
zhang-alvin, line 1447: will there be a use for old rescale?

proteus/mprans/SW2DCV.h
```diff
@@ -1034,243 +911,191 @@ class SW2DCV : public SW2DCV_base {
       // CALCULATE EDGE BASED CFL //
       //////////////////////////////
       double mi = lumped_mass_matrix[i];
-      edge_based_cfl[i] = 2 * fabs(dLowii) / mi;
+      edge_based_cfl[i] = 1.0 * fabs(dLowii) / mi;
```
zhang-alvin, line 914: was this a bug or was one of the other factors modified?

proteus/tests/SWFlow/comparison_files/comparison_solitary_island_h_t2.csv
```diff
@@ -1 +1 @@
...
...00105107793427,0.32000195410340865,0.3200005341073283,0.31999984235976126,0.31999995038139656,0.3199998456927162,0.31999996902126726,0.3199999329301024,0.31999999186049055,0.31999997802577995,0.3199999987468589,0.31999999420688047,0.3199999999516725,0.3199999987170015,0.32000000003477125,0.31999999975165094,0.3200000000131761,0.3199999999564599,0.32000000000323053,0.3199999999918939,0.32000000000125983,0.3199999889015452,0.32000003847261793,0.32000019206298447,0.3200005334323018,0.3200033299924424,0.3200078400916868,0.3200505631070416,0.32013090322752835,0.3207058631679516,0.32230491606901057,0.3288677307962416,0.3484757300065728,0.37343314876624617,0.36761792111885605,0.34303865071270895,0.32732430169006754,0.3219002770579276,0.3205613888393235,0.3201359684819116,0.32004018263465006,0.32000987868383135,0.3200030196256361,0.3200009169840541,0.32000030873233276,0.32000019656782536,0.32000005562445977,0.3200000719145213,0.3200000131227546,0.3200000255392641,0.32000000264086215,0.3200000076179129,0.3200000003398002,0.3200000019028069,0.32000000000261386,0.3200000004092204,0.3199999999875415,0.32000000007816254,0.3199999999957525,0.3200000000138391,0.3199999999988272,0.3200000000044532
```
zhang-alvin, line 1: maybe lfs this too? @cekees should we just move all .csv to git lfs moving forward?

proteus/NonlinearSolvers.py
ejtovar: I removed the "r" at the beginning of the comment, I assumed it was a typo.

proteus/NonlinearSolvers.py
ejtovar: Yes, this was one of the fixes. Wasn't the main fix however.

proteus/mprans/GN_SW2DCV.h
ejtovar: Well the limiting helps preserve positivity of the water height when we use a higher-order (in space) numerical method. The idea is that you can apply the limiting however many times you'd like, but we do it twice as a safe guard. Besides more computations, there shouldn't be a difference between doing it 2 or 10 times.

proteus/mprans/GN_SW2DCV.h
ejtovar: Yes, it helps me keep track of where things are in the code B)

proteus/mprans/GN_SW2DCV.h
ejtovar: I'm not sure at the moment. I left it there just in case I want to use it again. The "new_rescale" is what is implemented in the TAMU code, but still not 100% sure on the benefits. They seem to perform similarly in the "eye" norm.

proteus/mprans/SW2DCV.h
ejtovar: No, not a bug. This just changes the CFL condition by dividing it by 2. It's not a big deal. I only changed it because the TAMU code has the factor of 1 and I like the codes to be "exactly" the same when I compare them.

proteus/NonlinearSolvers.py
cekees: The "r" before the string signifies that that it's a raw string: https://docs.python.org/3/reference/lexical_analysis.html. That means you get almost exactly what is between the """'s, otherwise Python will interepret escape sequences. Mainly important when you include latex commands in the string.

proteus/mprans/GN_SW2DCV.h
cekees: I would say opinions differ on this practice, but it doesn't really hurt anything and can be useful for the kind of very long loop bodies that we sometimes have in optimized FEM assembly. Any good code editor, however, will easily show you matching brackets or parenthesis, which I trust more than a comment telling me when the loop ends (though the comment nicely records intent for posterity).  In emac, just set your options to highlight matching parenthesis. Of course Python went even further in assuming you have a decent editor by simply using indentation.

proteus/tests/SWFlow/comparison_files/comparison_solitary_island_h_t2.csv
cekees: Yes, my vote is to push them all to git lfs. It hasn't been that much of an issue to use lfs even on the hpc machines, and I think we still frequently have need to store non-source-code items in the repository.

proteus/mprans/GN_SW2DCV.cpp
```diff
@@ -5,30 +5,28 @@
...
-            std::allocator<int> a0;
-            std::allocator<double> a1;
-        }
-    }
+namespace workaround {
+inline void define_allocators() {
+  std::allocator<int> a0;
+  std::allocator<double> a1;
+}
+} // namespace workaround
```
cekees, line 13: We don't have a mandated indentation style for C++ and probably won't adopt one. So if you want to use K&R style instead of Allman style, it's fine by me, but it's better not to touch code that you're not changing unless you have a good reason. Making it all consistent in files that you are going to take ownership of is good enough reason, but you might want to do it in a separate PR so reviewers can focus on the substantive changes.

proteus/mprans/GN_SW2DCV.h
```diff
@@ -83,7 +84,7 @@ inline double ENTROPY_FLUX2(const double &g, const double &h, const double &hu,
           g * h * z) *
          hv * one_over_hReg;
 }
-// FOR ESTIMATING MAX WAVE SPEEDS, GIVE REFERENCE
+// FOR ESTIMATING MAX WAVE SPEEDS
```
cekees, line 87: I'm guessing @manuel-quezada was intending to come back at put a reference to the paper that derives these estimates.

Issue #1184, opened 2020-05-15
Work on consistent/safe way to use PetscOptions and/or file options in tests


cekees:

cekees:
* closes #14 
* closes #15

Issue #14, opened 2013-05-27
p1 and p2 linear solver options used in petsc.options not converging for SIPG


mfarthin:
superlu_dist (or LU in serial) work though

PR #1206, opened 2020-05-31
Address several old issus higher-order FEM, poisson tests, and testing problems

cekees:
# Mandatory Checklist

# Description

Initial work to close out #14, #15, and #16. Mainly fixing the tests themselves and petsc options. In the process I found that the DG archiving had never been properly updated and that a larger DG boundary penalty improves conditioning.  There were some messy test interactions that caused other changes--a project for the future.  On a positive note, @ejtovar @jhcollins @zhang-alvin I used the pytest parameterized testing support to generate more comprehensives tests here: https://github.com/erdc/proteus/pull/1206/files?file-filters%5B%5D=.csv&file-filters%5B%5D=.py#diff-50950c09443ffea929f7eb626c9cb336. It's nothing new but might be a useful example for you when writing tests. It replaces the old "yield tests".


proteus/NumericalFlux.py
```diff
@@ -761,7 +761,7 @@ def __init__(self,vt,getPointwiseBoundaryConditions,
         Advection_DiagonalUpwind_Diffusion_IIPG.__init__(self,vt,getPointwiseBoundaryConditions,
                                                          getAdvectiveFluxBoundaryConditions,
                                                          getDiffusiveFluxBoundaryConditions)
-        self.penalty_constant = 10.0
+        self.penalty_constant = 100.0
```
zhang-alvin, line 764: Might be good to make a global constant again.

proteus/TwoPhaseFlow/utils/Parameters.py
```diff
@@ -224,7 +224,7 @@ def __init__(self,
         # self.n.linearSmoother = None
         # NUMERICAL FLUX
         # self.n.massLumping = False
-        # self.n.conservativeFlux = None
+        self.n.conservativeFlux = None
```
ejtovar, line 227: What does un-commenting this do?

cekees:
> I saw that there were slight changes in how the petsc options were being loaded into the tests, or at least the defaults were reset. Does that resolve #1184 on an individual test level?

No, not really. I think that issue needs a more comprehensive review of how they're being set/loaded and a good across-the-board cleanup of that issue.

proteus/TwoPhaseFlow/utils/Parameters.py
cekees: None is the default, and it should be None here, so it's really just explicitly setting it to the default. There seem to be some strange test interactions with pytest, however, and I had an issue with this being set to something besides None.

proteus/NumericalFlux.py
cekees: On this one I'm going to just leave it as is. The best value of the penalty_constant in the DG-methods varies, unfortunately.

Issue #1208, opened 2020-06-08
FSI issue while running in parallel


jhcollins:
Running setups with moving meshes results in failure in parallel.

To reproduce the error, run the floating_body.py setup in proteus_tutorial/2d in parallel with any commit beyond: e9a061da02d393c336e70854c0d2f7a15956c62d

zhang-alvin:
@jhcollins thanks for catching this. Working on fix.

PR #1211, opened 2020-06-09
Fix MoveMesh cases

zhang-alvin:
# Mandatory Checklist

# Description
Addresses #1208

https://github.com/erdc/proteus/commit/8e7c357baa9ed44d11fb404415bc6f7950dd6abe introduced an error in the postStep for MoveMesh in parallel. Specifically, the global mesh object was being updated instead of the subdomain mesh object, leading to a bookkeeping error. 

A parallel test command is added to Travis in the FSI directory for testing/covering this behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ jobs:
       - PATH=./linux/bin:$PATH MPLBACKEND=Agg py.test -n 1 --dist=loadfile --forked -v proteus/tests --ignore proteus/tests/POD --cov=proteus
       - PATH=./linux/bin:$PATH MPLBACKEND=Agg py.test -n 1 --dist=loadfile --forked -v air-water-vv/Tests/1st_set --cov=proteus
       - PATH=./linux/bin:$PATH MPLBACKEND=Agg py.test -n 1 --dist=loadfile --forked -v air-water-vv/Tests/2nd_set --cov=proteus
+      - PATH=./linux/bin:$PATH MPLBACKEND=Agg mpiexec -np 2 pytest -v proteus/tests/FSI
   - stage: test
     name: "conda linux"
     os: linux
@@ -82,6 +83,7 @@ jobs:
       - py.test -n 1 --dist=loadfile --forked -v proteus/tests --ignore proteus/tests/POD --cov=proteus
       - MPLBACKEND=Agg py.test -n 1 --dist=loadfile --forked -v air-water-vv/Tests/1st_set
       - MPLBACKEND=Agg py.test -n 1 --dist=loadfile --forked -v air-water-vv/Tests/2nd_set
+      - MPLBACKEND=Agg mpiexec -np 2 pytest -v proteus/tests/FSI
     deploy:
       provider: script
       script: bash ./scripts/deploy.sh docs
--- a/proteus/mprans/MoveMesh.py
+++ b/proteus/mprans/MoveMesh.py
@@ -149,10 +149,10 @@ class Coefficients(proteus.TransportCoefficients.TC_base):
         #TODO: unclear if this needs to apply to all mesh types
         if self.nd == 2:  
            cmeshTools.computeGeometricInfo_triangle(self.mesh.subdomainMesh.cmesh)
-           self.mesh.buildFromC(self.mesh.cmesh)
+           self.mesh.subdomainMesh.buildFromC(self.mesh.subdomainMesh.cmesh)
         if self.nd == 3:  
            cmeshTools.computeGeometricInfo_tetrahedron(self.mesh.subdomainMesh.cmesh)
-           self.mesh.buildFromC(self.mesh.cmesh)
+           self.mesh.subdomainMesh.buildFromC(self.mesh.subdomainMesh.cmesh)
 
         copyInstructions = {'clear_uList': True}
         return copyInstructions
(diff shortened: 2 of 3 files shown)


zhang-alvin:
The parallel test does not pass locally for me. 
```
mpirun -np 2 pytest test_FSI.py -k "floatingCylinder"
============================= test session starts ==============================
platform linux -- Python 3.7.4, pytest-3.4.1, py-1.5.2, pluggy-0.6.0
rootdir: /home/zhanga/barn/proteus/temp_proteusRpath, inifile:
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
============================= test session starts ==============================
platform linux -- Python 3.7.4, pytest-3.4.1, py-1.5.2, pluggy-0.6.0
rootdir: /home/zhanga/barn/proteus/temp_proteusRpath, inifile:
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
collected 4 items

test_FSI.py collected 4 items

test_FSI.py F                                                            [100%]
...
[       5]    Newton it 1 == maxIts FAILED convergenceTest = r
[       5]    Newton it 1 norm(r) =  2.89750e-04  		 norm(r)/(rtol*norm(r0)+atol) =  1.15900e+03
[       5] Step failed due to solver failure
[       5] Sequence step failed
[       5] Sequence failed
[       5] System Step Failed
...
=================================== FAILURES ===================================
_______________________ TestIBM.test_floatingCylinderALE _______________________
        ns = NumericalSolution.NS_base(so,pList,nList,so.sList,opts)
        ns.calculateSolution('floatingCylinderALE')
        pos = case.body.getPosition()
    
>       npt.assert_almost_equal(pos, np.array([0.5, 0.5074055958, 0.]), decimal=5)
E       AssertionError: 
E       Arrays are not almost equal to 5 decimals
E       
E       Mismatched elements: 1 / 3 (33.3%)
E       Max absolute difference: 0.
E       Max relative difference: 0.01
E        x: array([0.5 , 0.51, 0.  ])
E        y: array([0.5    , 0.50741, 0.     ])

test_FSI.py:189: AssertionError
```

cekees:
Which model's Newton iteration is failing?

zhang-alvin:
> Which model's Newton iteration is failing?

AddedMass

cekees:
Have you tried it by commenting out those new postStep lines in MoveMesh.py instead of trying to fix them (just in case the fix is not really a fix)? It runs for me with them commented out. There is also another issue that needs fixing, which is that in the TwoPhaseFlow app Parameters.py, we need to force the AddedMass PETSc configuration to always use the iterative solver configuration. The superlu/superlu_dist configuration seems to work for me reliably, but since the system is singular (constant null space), I think we generally need to be using an iterative solver configured to be robust and scalable for pure Neumann Poisson problems. That doesn't explain why your commit fails and my local changes succeed though, because I haven't made that change yet.

zhang-alvin:
> Have you tried it by commenting out those new postStep lines in MoveMesh.py instead of trying to fix them (just in case the fix is not really a fix)? It runs for me with them commented out. There is also another issue that needs fixing, which is that in the TwoPhaseFlow app Parameters.py, we need to force the AddedMass PETSc configuration to always use the iterative solver configuration. The superlu/superlu_dist configuration seems to work for me reliably, but since the system is singular (constant null space), I think we generally need to be using an iterative solver configured to be robust and scalable for pure Neumann Poisson problems. That doesn't explain why your commit fails and my local changes succeed though, because I haven't made that change yet.

Yes commenting out those lines did allow the tests to pass. Maybe for this test the cylinder isn't moving enough for the postStep corrections to matter so much. 

I can achieve a similar effect as modifying the AddedMass PETSc configuration for this test case by directly supplying a petsc configuration file right?

cekees:
You may have to tighten the nonlinear and linear tolerance to get 5 decimal places of accuracy given that the problem is actually nonlinear and there will be some variation in the result of the linear solve.  It looks to me like the test doesn't set the tolerances so they are defaulting to values relative to the characteristic cell diameter, he, which is presumably pretty large for this quick test.

zhang-alvin:
> You may have to tighten the nonlinear and linear tolerance to get 5 decimal places of accuracy given that the problem is actually nonlinear and there will be some variation in the result of the linear solve. It looks to me like the test doesn't set the tolerances so they are defaulting to values relative to the characteristic cell diameter, he, which is presumably pretty large for this quick test.

Is there a setting akin to a direct solve like with use superLU in serial? It looks like it's using the superlu_dist petsc options so I'm a little surprised that there would be any differences from the linear algebra.

zhang-alvin:
I think this might take a little longer to diagnose the issue. I'm not convinced that the values should deviate on the order of 1e-5 in parallel if it's using superlu_dist as the linear solver. I'm also not convinced that the mesh coordinates are being reliably updated in parallel. 

Just for the sake of getting the tutorials test in place and a parallel test in place, I suggest rolling back the node/element diameter correction as it hasn't led to any noticeable errors and then reopening the corresponding issue. I can take a deeper look at this when I'm actively looking at a relevant case (PUMI falling cylinder case).

cekees:
OK, maybe try rolling it back on this PR and check if the new parallel tests pass.

cekees:
Also, to answer your question, use useSuperlu=True automatically does superlu in serial and superlu_dist in parallel.

PR #1223, opened 2020-08-19
SWFlow update: Fixed parallel capabilities and changed the reflecting boundary condition set up

ejtovar:
# Mandatory Checklist

# Description
I have made changes to the SWFlow solvers that have fixed the parallel capabilities. I also generalized the boundary conditions so that the user can use reflecting conditions and Dirichlet conditions instead of being restricted to just one. I added two problem files `mach_flow.py` and `obstacle_flow.py` that highlights this. 

I also made a small change to the Circle object in `SpatialTools.py` so that it can take in a boundaryTag provided by the user.   

New files added:

`proteus/tests/SWFlow/mach_flow.py -- 5KB`
`proteus/tests/SWFlow/obstacle_flow.py -- 7KB`


zhang-alvin:
It looks like the code coverage is dropping by a significant amount. Do you know what's causing that @ejtovar ?

ejtovar:
No not too sure, let me investigate.

proteus/NonlinearSolvers.py
```diff
@@ -755,26 +755,24 @@ def solve(self,u,r=None,b=None,par_u=None,par_r=None):
...
+            par_u.scatter_forward_insert()
 
         #############################################
         # UPDATE SOLUTION THROUGH calculateResidual #
         #############################################
         self.F.secondCallCalculateResidual = 1
         self.computeResidual(u,r,b)
-        self.F.check_positivity_water_height=True
+        if par_u is not None:
+            par_u.scatter_forward_insert()
```
zhang-alvin, line 773: Could you remind me what `scatter_forward_insert` does?

proteus/SpatialTools.py
```diff
@@ -119,6 +119,10 @@ def _checkFlags(self, flagSet):
         checkFlag = min(flagSet)
         assert checkFlag == 1, 'Minimum boundary/region tag/flag must be 1'
         for flag in flagSet:
+            # hack for SWFlow BCs
+            if (flag==99):
```
zhang-alvin, line 123: I think you should open up a corresponding issue to remove this hack, because I think it leads to unnecessarily large arrays. We talked briefly about this in the meeting, but we can probably migrate to using a different data structure for handling flags so that arrays don't wind up being formed based on the maximum flag value.

proteus/SpatialTools.py
```diff
@@ -945,23 +949,47 @@ def __init__(self, domain, radius, coords, barycenter, nPoints):
...
-        self.BC = {'circle': self.BC_class(shape=self, name='circle',
-                                       b_or=self.b_or, b_i=0),
-                  }
-        self.BC_list = [self.BC['circle'],
-                       ]
+        if boundaryTag is None:
+            self.boundaryTags = bt = {'circle':1}
+            vertFlag=[]
+            segmFlag=[]
+            for kkk in range(nVertices):
```
zhang-alvin, line 956: I think we can do better with this choice of index name. I don't have full context, but I'm quite sure anything else is better. For some of these loops, you can get away with using an underscore:

```
for _ in range(foo):
```

proteus/SpatialTools.py
```diff
@@ -1043,6 +1071,9 @@ def __init__(self, domain, barycenter=None, vertices=None,
         self.b_or = b_or
         for tag, flag in boundaryTags.items():
             b_i = flag-1  # start at index 0
+            # hack for SWFlow BCs
+            if (b_i==98):
```
zhang-alvin, line 1075: point this out in the issue that is to be opened.

proteus/mprans/SW2DCV.h
```diff
@@ -1604,9 +1512,9 @@ class SW2DCV : public SW2DCV_base {
                          ith_dHij_minus_muHij_times_hvStarStates -
                          ith_muHij_times_hvStates);
           // clean up potential negative water height due to machine precision
-          if (globalResidual[offset_h + stride_h * i] >= -hEps &&
-              globalResidual[offset_h + stride_h * i] < hEps)
-            globalResidual[offset_h + stride_h * i] = 0;
+          // if (globalResidual[offset_h + stride_h * i] >= -hEps &&
```
zhang-alvin, line 1515: should this just be removed?

proteus/SpatialTools.py
ejtovar: Yes of course. I wasn't aware of the context (I didn't write this function), so I didn't change the variable name.

proteus/mprans/SW2DCV.h
ejtovar: Yes, I need to clean out some of the commented code.

proteus/NonlinearSolvers.py
cekees: You can take a look at the PetscVec GhostUpdate functionality here: https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecGhostUpdateBegin.html

Basially, a parallel vector has a local representation u and a distributed/global representation U. The local representation will have both owned entries and ghost entries owned by other MPI processes.  A forward scatter updates the ghost entries with values from the owned entries and the INSERT mode says to just insert the values from the owned elements. We do this when a process has correctly computed the owned entries of u but not necessarily the ghost entries. So after a scatter_foward_insert, the local representations of U are all correct on all processors.  On the other hand for the residual r (R) with a non-overlapping,  element-based partitioning you usually do a reverse scatter, sending ghost values to the owning process and you do it in ADD mode because you're trying to accumulate all the element contributions to the owning node. After that all the owned entries are correct but not necessarily the ghost entries.

proteus/SpatialTools.py
cekees: +1

cekees: @ejtovar you could just add the issue then mark this as resolved.

cekees:
Looks like just the new test failing to get 7 sig figs on linux. Not sure what's going on--don't have a linux machine at the moment.

proteus/SpatialTools.py
ejtovar: @zhang-alvin When the flag 99 is read, it only adds an extra entry to the boundary tags so that the size is size(normalTags)+1 and not size(normalTags)+99. I'll go ahead open the issue anyway so we can generalize this later.

ejtovar:
@cekees Yeah I'm not sure why the test is failing. I'm comparing both solutions and they look the "same".

PR #1225, opened 2020-09-07
Add full topography effects for dispersive SWEs solver.

ejtovar:
# Mandatory Checklist

# Description
In this PR, we add the full topography effects into the solver for the dispersive Serre--Saint-Venant equations (i.e., dispersive shallow water equations). This consists of adding an extra equation in the PDE system along with other terms in the equations. (See paper: `Hyperbolic relaxation technique for solving the dispersive Saint-Venant shallow water equations with topography.`) 
I also changed a few test file names to better reflect the tests. 

# New files 
santos_step.py -- 6KB


zhang-alvin (commented):
My understanding is that this is an extension or modification to an existing template/method. Why do some of the test results (.csv files) need to be changed? Was there a bug fix or did you simply switch the underlying method used to obtain the results? If it's the latter, it may be better to simply add an additional comparison rather than replacing the existing one.

cekees (approved):
Looks good!

PR #1228, opened 2020-09-17
Release candidate 1.7.5

cekees:
# Mandatory Checklist

# Description


cekees:
@zhang-alvin the name change of MeshAdapt.pyx to Adapter.pyx was breaking the documentation generation because the extension module was still named MeshAdapt.pyx. Maybe when wrapping up one of the other PR's you can make the change to Adapter just making sure you change the setup.py file and whatever python modules depend on MeshAdapt/Adapter. Unless I misunderstood what is going on here.

cekees:
@ejtovar in reviewing the docs I found an extra "SWFlows" file, which I removed because it wasn't being used. The current docs use the "Green-Naghdi" naming scheme so I guess if we want to be consistent width the papers we should do some editing so  the paper and a future release are using the same terminology.

cekees:
> Didn't know that there'd be an issue with documentation. I'll redo the name change at a later PR.

Thanks, I should have been more specific: it just shows up as a warning when generating the webpage page but then also creates a broken link. I believe sphinx load all the modules for the package with python bug then actually parses the pyx files for the docstrings or something along those lines. With the proteus module named MeshAdapt but the pyx file for named Adapter, it seems to get confused.

ejtovar:
> @ejtovar in reviewing the docs I found an extra "SWFlows" file, which I removed because it wasn't being used. The current docs use the "Green-Naghdi" naming scheme so I guess if we want to be consistent width the papers we should do some editing so the paper and a future release are using the same terminology.

Do you want me to make the changes for this release or a different one?

ejtovar (approved):
Looks good to me!

PR #1231, opened 2020-09-22
fix meshing in conda environments and renable tests

cekees:
# Mandatory Checklist

# Description

### DIFF
--- a/proteus/TwoPhaseFlow/utils/Parameters.py
+++ b/proteus/TwoPhaseFlow/utils/Parameters.py
@@ -1762,14 +1762,9 @@ class ParametersModelAddedMass(ParametersModelBase):
 
     def _initializePETScOptions(self):
         prefix = self.n.linear_solver_options_prefix
-        if self._Problem.useSuperlu:
-            self.OptDB.setValue(prefix+'ksp_type', 'preonly')
-            self.OptDB.setValue(prefix+'pc_type', 'lu')
-            self.OptDB.setValue(prefix+'pc_factor_mat_solver_type', 'superlu_dist')
-        else:
-            self.OptDB.setValue(prefix+'ksp_type', 'cg')
-            self.OptDB.setValue(prefix+'pc_type', 'gamg')
-            self.OptDB.setValue(prefix+'ksp_max_it', 2000)
+        self.OptDB.setValue(prefix+'ksp_type', 'cg')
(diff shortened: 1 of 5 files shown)


proteus/TwoPhaseFlow/utils/Parameters.py
```diff
@@ -1762,14 +1762,9 @@ def _initializeNumerics(self):
...
         prefix = self.n.linear_solver_options_prefix
-        if self._Problem.useSuperlu:
-            self.OptDB.setValue(prefix+'ksp_type', 'preonly')
-            self.OptDB.setValue(prefix+'pc_type', 'lu')
-            self.OptDB.setValue(prefix+'pc_factor_mat_solver_type', 'superlu_dist')
-        else:
-            self.OptDB.setValue(prefix+'ksp_type', 'cg')
-            self.OptDB.setValue(prefix+'pc_type', 'gamg')
-            self.OptDB.setValue(prefix+'ksp_max_it', 2000)
+        self.OptDB.setValue(prefix+'ksp_type', 'cg')
```
ejtovar, line 1765: I'm not really familiar with the solver options for the TwoPhaseFlow models, so can you explain what is going on here @cekees? Why were the `useSuperlu` options removed?

ejtovar (commented):
Looks like the `proteus/tests/proteus_tutorial/tests/test_floating_body.py` is still failing on the OSX build with `nan`'s in the Added Mass matrix. Is this a solver issue?

proteus/TwoPhaseFlow/utils/Parameters.py
cekees: The added mass model here is a pure Neumann problem for the Poisson equation so it has the constant vectors in the null space.  I think generally it's safer to use an iterative solver that knows about the null space to solve this type of singlar problem.  It seems that superlu can handle it sometimes, but I'm not familiar with the details on how it's doing it. Unfortunately, this does not seem to be the issue with the travis osx test failure in any case.

cekees: Turns out it was the issue... The way the TwoPhaseFlow app is implemented right now you have to override this LU option in the child class _initializeNumerics funciton. /cc @zhang-alvin

ejtovar:
Seems like everything is good to go now that the tests are passing!

ejtovar (approved):

Issue #1235, opened 2020-10-07
Hot start not advancing the time step


SavantG:
Hot starting using -H isn't progressing the time step on successful convergence. Below is a snippet of the log
"
[      39] ==============================================================
[      39] Solving over interval [ 4.30000e+01, 4.40000e+01]
[      39] ==============================================================
[      39] System time step t= 4.30067e+01, dt= 6.65406e-03
[      39] saving previous velocity dofs 0
[      39] Split operator iteration 0
[      39] Model: rans2p
[      39] Fractional step  4.30067e+01 for model rans2p
[      39] Model step t= 6.65406e-03, dt= 6.65406e-03 for model rans2p
[      39] Model substep t= 6.65406e-03 for model rans2p
[      39] Model substep t= 6.65406e-03 for model rans2p model.timeIntegration.t=  4.30067e+01
[      39] Maximum CFL = 0.75
[      39]    Newton it 0 norm(r) =  2.25572e+01                 norm(r)/(rtol*norm(r0)+atol) = 4.01017e+06 test=r
[      65] before ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= False ksp.its= 0 ksp.norm= 0.0
[      65] after ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= True ksp.its= 1 ksp.norm= 0.0 reason = 4
[      65] Maximum CFL = 0.7500612245648367
[      65]    Newton it 1 norm(r) =  3.62882e-01                 norm(r)/(rtol*norm(r0)+atol) = 64512.3 test=r
[      80] before ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= False ksp.its= 1 ksp.norm= 0.0
[      80] after ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= True ksp.its= 1 ksp.norm= 0.0 reason = 4
[      80] Maximum CFL = 0.7500737283082229
[      80]    Newton it 2 norm(r) =  2.87280e-02                 norm(r)/(rtol*norm(r0)+atol) = 5107.19 test=r
[      95] before ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= False ksp.its= 1 ksp.norm= 0.0
[      95] after ksp.rtol= 0.01 ksp.atol= 5.625e-08 ksp.converged= True ksp.its= 1 ksp.norm= 0.0 reason = 4
[...]

PR #1236, opened 2020-10-09
Fix to allow for hotstarting with systemStepExact=False

zhang-alvin:
# Mandatory Checklist

# Description

Addresses #1235. Not setting the model time prior to proceeding with calculation leads to an infinite loop.

Thanks @cekees for the suggested fix.

Added a test within `tests/TwoPhaseFlow` to test for the fix. Note to keep in mind is that the test is to avoid changes to the infinite loop fix.

I tried setting a test to compare numerical results but I believe something along the lines of #1184 was causing a different behavior between the directory-level `pytest` call and a `make test` call at the root.

### DIFF
--- a/proteus/NumericalSolution.py
+++ b/proteus/NumericalSolution.py
@@ -870,6 +870,7 @@ class NS_base(object):  # (HasTraits):
                 #the number of nodes in an adapted mesh is not necessarily going to be the same as that of the solution field when archived...but it's not important because things should be bookkept correctly later on
                 #if not isinstance(p.domain,Domain.PUMIDomain):
 
+                m.stepController.t_model_last=time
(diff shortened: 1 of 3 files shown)


cekees (approved):

zhang-alvin:
@cekees, I think the line in NumericalSolution is causing a massive code cov decrease. I can try moving the line elsewhere or is there some simpler fix to the drop in coverage.

I think @ejtovar ran into something similar pretty recently. I forget how it was resolved.

zhang-alvin:
For reference, prior to hack/removal of test, codecov indicated -36% coverage.
Removing test yields only a -0.01% drop.
Going to turn the test back on to see if behavior is corrected.

PR #1239, opened 2020-12-01
Partition Meshes 2D

zhang-alvin:
# Mandatory Checklist

# Description
Try to get the partition logic for 2D triangle meshes working

### DIFF
--- a/proteus/MeshTools.py
+++ b/proteus/MeshTools.py
@@ -675,19 +675,37 @@ class Mesh(object):
             logEvent("base {0:d}".format(base))
             logEvent("nLayersOfOverlap {0:d}".format(nLayersOfOverlap))
             logEvent("parallelPartitioningType {0:d}".format(parallelPartitioningType))
-            (self.elementOffsets_subdomain_owned,
-             self.elementNumbering_subdomain2global,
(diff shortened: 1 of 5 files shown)


zhang-alvin:
When running a serial case with the -F option on a mesh that has already been generated, the analysis proceeds with identical results: 

serial_vs_loadSerial

zhang-alvin:
Running the same case with two cores yields the following assertion error:

```
python: proteus/partitioning.cpp:3961: int proteus::partitionNodesFromTriangleFiles(const MPI_Comm&, const char*, int, Mesh&, int): Assertion `edges_overlap.size() + nEdges_subdomain_new[rank] == edgeNodesMap.size()' failed.
```

A print statement for `edges_overlap.size()`, `nEdges_subdomain_new`, and `edgeNodesMap.size` shows the following:
```
rank 0 
edges_overlap=6 
nEdges_subdomain_new[0]=235 
edgeNodesMap.size()=241

rank 1
edges_overlap=18
nEdges_subdomain_new[1]=227
edgeNodesMap.size()=244
``` 
The assertion fails for rank 1 because the sum of the number of overlapping edges and number of owned edges is off by one.

PR #1246, opened 2021-02-19
New update stack2

cekees:
# Mandatory Checklist

# Description

This is a major infrastructure update, but it doesn't change any frequently used functionality in Proteus proper. The coverage is going to show up as reduced, but too much non-proteus package code was getting included in the coverage computation. Moving forward we should be able to  focus on increasing the coverage on the package itself and further removing unused code and dependencies as well as tests that are being skipped on any platforms.


cekees:
@adimako one thing I found that would be a small job to increased portability and remove dependencies would be to remove the curses.h code in the Fenton library. It's not a big deal if the interactive, text-based selection of Fenton theory parameters is useful (though in the long run it might be easier to reimplement that interface code using python and leave the C++ just for the computations.

cekees:
@ejtovar LFS is working fine for me on both mac os x and linux. I did have some headaches getting all the files properly out of git and into LFS with the proper gitattributes set. I decided that we're never going to be free of binary and large data files for testing, so I'd rather just consistently use LFS than keep those file types in the repository. If you need to add csv, npy, bin, etc.  Just do `git lfs track filename` before your do `git lfs add`. Or if you accidently add something, do `git rm --cached filename; git commit;` before adding it back properly.

proteus/WaveTools.h
```diff
@@ -406,7 +406,7 @@ namespace proteus
 
 
 
- inline double* __cpp_uWindow(double* U, double x[nDim], double x0[nDim], double t, double* t0, double* kDir, double* kAbs, double* omega, double* phi, double* amplitude, double mwl, double depth, int N,int Nw, double* waveDir, double* vDir, double* tanhF, double gAbs , bool fast)
+ inline void __cpp_uWindow(double* U, double x[nDim], double x0[nDim], double t, double* t0, double* kDir, double* kAbs, double* omega, double* phi, double* amplitude, double mwl, double depth, int N,int Nw, double* waveDir, double* vDir, double* tanhF, double gAbs , bool fast)
```
cekees, line 409: @adimako I found this pernicious little bug in the optimized WaveTools.h code. My understanding is that you designed the _u* functions to have no return value but this one declared it as double*.  I checked over the module and didn't find any other cases, but you might want to take a quick look. It generates a seg fault when optimization is on for the newer compilers. @ejtovar I suspect this may be what causes random waves test to fail on the conda stacks. Note, I also found an issue like this in the optimized quadrilateral mesh generation.  With these bug fix commits I also reset the hashdist test compile options to -O2 to verify that all the tests run with full optimization on the hashdist stack.

ejtovar (approved):

